PR #3, opened 2019-05-16
Fix typo in documentation

mbuhot:
Hi, just came across a typo when reading though the docs.

strucs -> structs

Looks like a great project, thanks for creating and sharing it :)


sorentwo:
Thanks, EctoJob was definitely an inspiration 👍

PR #8, opened 2019-05-30
Add `Oban.drain_queue/1` for integration testing

### LINKED ISSUE #7: Telemetry and testing questions
Hi @sorentwo,

Great work on the library! I am really glad to see a database-based library with this robust feature set. I have two quick questions (if there is a better place to post them, please let me know):

1. Regarding telemetry, we have been standardizing around the name `:duration` for the time measurement to execute something. So what do you think about replacing the current `%{timing: timing}` even in favor of `%{duration: timing}`? Other than that, thank you for using telemetry and help us push it as a standard in the community! 🎉 I also like how you structured the events around success and failure.

2. Regarding testing, sometimes I may actually want the jobs to run (for example, during integration tests) because they may do important work that is needed for the continuation of the testing. Is it possible to have a testing function that executes all of the jobs in a given queue? This way I can do some action and execute all enqueued jobs in the test before proceeding.

Thanks!

sorentwo:
Draining a queue synchronously executes all available jobs in the queue from within the calling process. This avoids any sandbox based database connection issues and prevents race conditions from asynchronous processing.

@josevalim @LostKobrakai would you mind taking a look?

Closes #7

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -177,7 +177,7 @@ defmodule Oban do
     test "we stay in the business of doing business" do
       :ok = Business.schedule_a_meeting(%{email: "monty@brewster.com"})
 
-      assert 1 == Oban.drain_queue(:mailer)
+      assert %{success: 1, failure: 0} == Oban.drain_queue(:mailer)
 
       # Make an assertion about the email delivery
     end
@@ -393,9 +393,11 @@ defmodule Oban do
   ## Failures & Retries
 
   Draining a queue uses the same execution mechanism as regular job dispatch. That means that any
-  job failures or crashes will be captured and result in a retry. Failures are _not_ reported back
-  to the calling process. If you expect jobs to fail, would like to track failures, or need to
-  check for specific errors you can use one of these mechanisms:
+  job failures or crashes will be captured and result in a retry. Retries are scheduled in the
+  future with backoff and won't be retried immediately.
+
+  Exceptions are _not_ raised in to the calling process. If you expect jobs to fail, would like to
+  track failures, or need to check for specific errors you can use one of these mechanisms:
 
   * Check for side effects from job execution
   * Use telemetry events to track success and failure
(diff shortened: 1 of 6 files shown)


lib/oban.ex
```diff
@@ -353,6 +381,38 @@ defmodule Oban do
...
+
+  Draining a queue from within the current process is especially useful for testing. Jobs that are
+  enqueued by a process when Ecto is in sandbox mode are only visible to that process. Calling
+  `drain_queue/1` allows you to control when the jobs are executed and to wait synchronously for
+  all jobs to complete.
+
+  ## Failures & Retries
+
+  Draining a queue uses the same execution mechanism as regular job dispatch. That means that any
+  job failures or crashes will be captured and result in a retry. Failures are _not_ reported back
```
josevalim: Are they retried immediately? Or will they be requeued and then I need to drain the queue again?

lib/oban.ex
```diff
@@ -353,6 +381,38 @@ defmodule Oban do
...
+  * Check for side effects from job execution
+  * Use telemetry events to track success and failure
+  * Check the database for jobs with errors
+
+  ## Example
+
+  Drain a queue with three available jobs:
+
+      Oban.drain_queue(:default)
+      3
```
josevalim: Is this the number of failures, of successes, or of jobs?

Perhaps you should return a map in case you want to return more information in the future? Something like `%{total: 3}`?

lib/oban.ex
sorentwo: There is a backoff for all jobs which causes them to be scheduled in the future. Draining only executes currently available jobs so they won't be retried immediately. I'll make that clear in the docs. 👍

lib/oban.ex
sorentwo: It is the number of jobs, but it is completely ambiguous. Returning a map with more information (or at least making it explicit) would be great.

lib/oban.ex
LostKobrakai: Explicitly needing to redrain the queue actually feels more explicit and like more control over testing stuff, which includes failures.

lib/oban.ex
sorentwo: Updated to return a map with success and failure counts, i.e. `%{success: 1, failure: 0}` 👍

lib/oban.ex
sorentwo: @LostKobrakai Agreed, preventing an automatic retry seems more explicit to me. I'm having a hard time imagining a situation where, during an integration level test, a job would error and I would want to retry it again.

sorentwo:
Thanks for the review!

PR #10, opened 2019-06-06
Fix typo in scheduling example

lukerollans:
What it say's on the tin!

I think the typo form is preferrable - `schedule_in`, however, updating the changeset field name would be a breaking change.. so perhaps a good candidate for the next major version release

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,11 +269,11 @@ The worker's defaults may be overridden by passing options:
 |> MyApp.Repo.insert()
 ```
 
-Jobs may be scheduled down to the second any time in the future:
+Jobs may be scheduled or delayed down to the second any time in the future:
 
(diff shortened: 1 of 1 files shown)


sorentwo:
Thanks for catching the inconsistency! I agree that `schedule_in` sounds better than `scheduled_in` and have changed the interface to match the README.

The project is pre 1.0 and minor breaking changes are acceptible (from [semver](https://semver.org/#spec-item-4)):

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

Closing in favor of 524fb22 👍

lukerollans:
Honestly I didn't actually notice the library was pre 1.0, so no worries!

PR #11, opened 2019-06-06
Raise on test run so `mix ci` alias chain does not report false pass

lukerollans:
As seen here on CircleCI 👉 https://circleci.com/gh/sorentwo/oban/57

`mix test` is currently exiting with a non zero code, but given how alias chains work, CircleCI still believes the task completes successfully given Dialyzer exits with 0.

This change will raise on failure so we actually know if a test is failing on CI

EDIT: Apologies for not actually fixing the failing tests, I will loop back and do just that if someone else doesn't get to it first :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,5 +365,5 @@ following commands:
 
 * Check formatting (`mix format --check-formatted`)
 * Lint with Credo (`mix credo --strict`)
-* Run all tests (`mix test`)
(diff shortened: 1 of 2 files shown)


sorentwo:
Thank you for this! Don't worry about fixing the failing test(s), I've fixed it. The failures were coming from some async tests leaking jobs.

PR #19, opened 2019-06-19
Added prune_limit option to prevent timeouts

sorenone:
Added prune limit option for stability and to prevent timeouts.


sorentwo:
Thank you! ❤️ 💜 💛

PR #20, opened 2019-06-20
Treat returned error tuples as a job failure

### LINKED ISSUE #17: Semantics of how to `:error` tuples should be treated as job failures
We are digging into integrating oban telemetry events with [AppSignal](https://github.com/appsignal/appsignal-elixir) and we are thinking of how oban could better support idiomatic `ok`/`error` tuples from application code to understand that something failed. We have something like this:

```elixir
def perform(%{"id" => id}) do
  record = MyApp.Repo.get!(MyApp.Client, id)
  MyApp.Context.do_work(record)
end
```

`Context.do_work/1` was originally written to return `{:ok, %MyApp.Client{}}` or `{:error, error}` and it should never raise an exception, but as oban ignores return values the job runs successfully and is never retried/discarded. We could have some code to manually `raise`/`throw` inside our job module, but it feels like something that Oban could handle by itself (I don't know how other elixir libraries deal with it).

I'm proposing that we change the following function: https://github.com/sorentwo/oban/blob/419d5b90eb7e229d2dd01ce777f374a48338906e/lib/oban/queue/executor.ex#L39-L51

If `perform/1` returns `{:error, error}`, I think `safe_call` should return `{:failure, job, :error, error, __STACKTRACE__}` and the job will be treated as failed, and the scenario where exceptions were raised should use `:exception` instead of `:error` for the `kind` prop (this makes easier for the event listeners to know it's an actual exception or some other data structure that represents the error).

What do you think?

sorentwo:
This changes `safe_call` to treat error tuples as a failure. The job will be retried just as if there was an exception or an exit, but the telemetry kind is different. Now telemetry events differentiate between a manual error and an exception.

Closes #17

@lucasmazza Would you mind taking a look at this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   deleted on each prune iteration. This prevents locking the database when there
   are a large number of jobs to delete.
 
+### Changed
+
+- [Oban.Worker] Treat `{:error, reason}` tuples returned from `perform/1` as a
+  failure. The `:kind` value reported in telemetry events is now differentiated,
+  where a rescued exception has the kind `:exception`, and an error tuple has
+  the kind `:error`.
+
 ### Fixed
 
 - [Oban.Testing] Only check `available` and `scheduled` jobs with the
--- a/README.md
+++ b/README.md
@@ -244,9 +244,10 @@ defmodule MyApp.Workers.Business do
 end
 ```
 
-The return value of `perform/1` doesn't matter and is entirely ignored. If the
-job raises an exception or throws an exit then the error will be reported and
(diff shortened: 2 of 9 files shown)


lucasmazza:
@sorentwo the diff here looks awesome, thanks a lot! Tomorrow I can try out this on our app to test it out quickly and see how it does on practice

sorentwo:
@lucasmazza did you get a chance to test this out? I would like to include it in v0.5.0, which I’m hoping to release this week.

lucasmazza:
I haven't deployed this to our staging server yet. I'll try to get back to
you tomorrow with this.

sorentwo:
@lucasmazza This looks good on my end. Getting it in for a 0.5 release.

lucasmazza:
@sorentwo :+1: seems to be working fine here too

PR #31, opened 2019-07-16
Show available jobs when assert_enqueued fails

luizpvas:
Hi there!

I was having some trouble figuring out why a job was not enqueued (a misspelling, it turns out), and I thought it was a good idea to print the available jobs in case it fails to match. Here's a print of the error message in my terminal:

image

Thanks!

### DIFF
--- a/lib/oban/testing.ex
+++ b/lib/oban/testing.ex
@@ -170,10 +170,12 @@ defmodule Oban.Testing do
   end
 
   defp available_jobs(repo, opts) do
+    fields = Keyword.keys(opts)
+
     base_query([])
-    |> select(^Keyword.keys(opts))
+    |> select(^fields)
     |> repo.all()
-    |> Enum.map(&Map.take(&1, Keyword.keys(opts)))
+    |> Enum.map(&Map.take(&1, fields))
   end
 
   defp base_query(opts) do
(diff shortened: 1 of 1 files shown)


lib/oban/testing.ex
```diff
@@ -166,6 +169,13 @@ defmodule Oban.Testing do
     |> repo.one()
   end
 
+  defp available_jobs(repo) do
+    base_query([])
+    |> select([:worker, :args])
```
sorentwo: Technically any field may be used to construct the query and that should be reflected in the inspected value. For example, if they were looking for `worker: MyWorker, attempt: 2, max_attempts: 3` all of those fields would be considered in the search but wouldn't be reflected here.

You can pull the keys out of `opts` to select only the fields that are being compared, similar to how `normalize_opts/1` does it below.

luizpvas: Thanks for the explanation :+1: 

I changed the function to call `Keyword.keys` in both select and Map.take

lib/oban/testing.ex
```diff
@@ -169,11 +169,11 @@ defmodule Oban.Testing do
     |> repo.one()
   end
 
-  defp available_jobs(repo) do
+  defp available_jobs(repo, opts) do
     base_query([])
-    |> select([:worker, :args])
+    |> select(^Keyword.keys(opts))
```
sorentwo: Let's pull these out above the pipe chain and reference them in both places.

luizpvas: sure :+1:

Issue #27, opened 2019-07-08
Support a `discarded` callback


andykent:
I have a requirement to take some action when a job reaches it's max attempts and is discarded from the queue. I was hoping it would be possible to add a callback to `Oban.Worker` to handle this. I imagine something like this...

```elixir
defmodule MyWorker do
  use Oban.Worker, max_attempts: 5

  def discarded(job) do
    Logger.info("job #{job.id} was discarded after #{job.attempts}")
  end
end
```

_note: my use case is more complex than simple logging but hopefully it illustrates the point._

If this is something you would be willing to support in Oban I can have a go at putting together a PR to add the functionality.

sorentwo:
This is currently possible through telemetry handlers, though it is a bit obtuse. These was [recently asked on the forum](https://elixirforum.com/t/oban-reliable-and-observable-job-processing/22449/57?u=sorentwo), so it is clearly a common need.

To recap what was posted on the forum, the handler function would look like this:

```elixir
def handle_discard([:oban, :failure], _, %{attempt: attempt, max_attempts: max}, _) do
  if attempt == max do
    Logger.info("job #{job.id} was discarded after #{job.attempts}")
  end
end
```

_This could probably use some better documentation and/or more fleshed out examples_

The reason I've promoted using telemetry events is because they allow you to handle events from multiple workers (or every worker) without duplicating the same code in each one. For example, you can define a single error handler for all of your workers in a central place rather than sprinkling callbacks everywhere.

tcoopman:
A possible issue with telemetry is that it isn't 100% reliable. If telemetry or your listener or the app crashes, you will have missed the message.
For logging that's not an issue, but for business logic it might be.

andykent:
Yep I did think about the using the telemetry handler but it felt wrong for my use case. If that's the best option then maybe I should just use that though. 

To explain my use case... I have a scheduled job that I want to ensure gets attempted daily. I would like it to look something like the following...

```elixir
defmodule MyWorker do
  use Oban.Worker, max_attempts: 5
  
  def perform(email) do
    send_notifications(email)
    reschedule(email)
  end

  def discarded(%{args: args}) do
        reschedule(args)
  end
end
```

Ideally I was thinking that the `discarded` callback would run inside the same transaction as the DB update that sets the job as discarded, this would mean that if anything died during discarding a message then the message would not get discarded in the DB, giving you a guarantee that a job wouldn't be marked as discarded without the callback running to completion.

[...]

PR #32, opened 2019-07-17
Accept a job struct as the argument to perform/1

sorentwo:
The `perform/1` function now receives an `Oban.Job` struct as the sole argument, calling `perform/1` again with only the `args` map if no clause catches the struct. This allows workers to use any attribute of the job to customize behaviour, e.g. the number of attempts or when a job was inserted into the database.

The implementation is entirely backward compatible, provided workers are defined with the `use` macro. Workers that implement the `Oban.Worker` behaviour manually will need to change the signature of `perform/1` to accept a job struct.

Addresses the first part of #27 

---

I'm opening this as a PR to get some feedback on the implementation and documentation.

/cc @andykent

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   an `ArgumentError` and halt startup. This prevents misconfiguration through
   typos and passing unsupported options.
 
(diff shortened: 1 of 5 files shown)


andykent:
Looks great and I think the documentation is clear. I agree that this is a good first step.

sorentwo:
@andykent Thanks for taking a look 💛

PR #37, opened 2019-07-18
Add reference to scheduled_at in readme

khedaywi:
Make reference that _scheduled_at_ is available in addition to _schedule_in_ as a parameter.


khedaywi:
Done! 👍🏼

sorentwo:
Thanks again! 💛

PR #41, opened 2019-07-25
Fix exception message when validating node opt

mtarnovan:
Just a minor copy-pasta I noticed while browsing through the source code. 👍 for starting this project, I hope it goes a long way, would really like to see many more features (batch jobs and backpressure for example).


sorentwo:
Thanks for the fix! 💛

PR #46, opened 2019-08-04
Remove perform/1 callback in favor of perform/2

### LINKED ISSUE #45: Deprecate perform/1 in favor of perform/2
Thanks for this fantastic library!

Currently, users expect the job args to be passed to perform/1, and when they need access to the `%Job{}`, they define a new `perform/1` clause which matches on the job. This works because Oban injects:

```elixir
      def perform(%Job{args: args}), do: perform(args)
      def perform(args) when is_map(args), do: :ok
```

While this works with very little code injection, it's non obvious to users how to reach the `Job` struct when needed, and non obvious that once they define their own clause that matches the   struct, they may invalidate other previously valid perform/1 clauses depending on what they are doing. Instead of the argument type dance, I propose we introduce `perform/2` which receives the job struct, followed by arguments. For the general use case, this still looks quite nice:

```elixir
def perform(_job, %{some: arg}) do
```

This way it's clear what is called and when, and also makes the job metadata instantly accessible once needed. I think the throwaway arg is a very reasonable tradeoff for the times the job info isn't needed. Thoughts?

sorentwo:
The new `perform/2` function receives the job's args, followed by the complete job struct. This new function signature makes it clear that the args are always available, and the job struct is also there when it is needed.

This is a breaking change that will require all worker modules to be updated.

Closes #45

/cc @chrismccord

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,17 +22,41 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   window of time. For example, to make a job unique by args, queue and worker
   for 2 minutes:
 
-  ```
-  use Oban.Worker, unique: [period: 120, fields: [:args, :queue, :worker]]
-  ```
+  `use Oban.Worker, unique: [period: 120, fields: [:args, :queue, :worker]]`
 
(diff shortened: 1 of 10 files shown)


sorentwo:
Thanks for taking a look 👍

PR #47, opened 2019-08-07
Fix typo in readme

derrickreimer:
Just spotted a tiny typo.

By the way, this library looks fantastic!


sorentwo:
Thanks for the fix. That was a tiny typo!

PR #50, opened 2019-08-16
Remove query for `scheduled_at` now or in the future from `fetch_available_jobs/3`

seanwash:
Not sure how long the Slack logs will be around so I took a screenshot: https://c.seanwash.com/1565977760.png

I'm running my Elixir app/tests on my laptop and my database is running inside of Docker. Somehow the current time between the two differ, causing jobs to not run in tests when using `drain_queue/1`.


sorentwo:
Thanks for the PR. That line is a remnant from before there was a `scheduled` state. I’m really pleased we figured out your issue!

PR #51, opened 2019-08-16
Don't eat ArgumentError from worker backoff function

balexand:
If an `ArgumentError` is raised in the `backoff` function of a worker then it is silently eaten. This can hide bugs and make debugging confusing. This is particularly problematic since `ArgumentError` is very common exception.

A safer alternative is to check if the `backoff` function is defined before calling it. Then there is no need to catch exceptions.

Thanks ❤️

### DIFF
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -71,7 +71,12 @@ defmodule Oban.Queue.Executor do
   # `safe_call/1`. There is a possibility that the worker module can't be found at all and we
   # need to fall back to a default implementation.
   defp worker_backoff(%Job{attempt: attempt, worker: worker}) do
-    module = to_module(worker)
+    module =
+      try do
+        to_module(worker)
+      rescue
+        ArgumentError -> nil
+      end
 
     if function_exported?(module, :backoff, 1) do
       module.backoff(attempt)
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -71,11 +71,13 @@ defmodule Oban.Queue.Executor do
   # `safe_call/1`. There is a possibility that the worker module can't be found at all and we
   # need to fall back to a default implementation.
   defp worker_backoff(%Job{attempt: attempt, worker: worker}) do
-    worker
-    |> to_module()
-    |> apply(:backoff, [attempt])
-  rescue
-    ArgumentError -> Worker.default_backoff(attempt)
+    module = to_module(worker)
+
+    if function_exported?(module, :backoff, 1) do
+      module.backoff(attempt)
+    else
+      Worker.default_backoff(attempt)
+    end
   end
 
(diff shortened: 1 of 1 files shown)


lib/oban/queue/executor.ex
```diff
@@ -71,11 +71,13 @@ defmodule Oban.Queue.Executor do
   # `safe_call/1`. There is a possibility that the worker module can't be found at all and we
   # need to fall back to a default implementation.
   defp worker_backoff(%Job{attempt: attempt, worker: worker}) do
-    worker
-    |> to_module()
-    |> apply(:backoff, [attempt])
-  rescue
-    ArgumentError -> Worker.default_backoff(attempt)
+    module = to_module(worker)
```
sorentwo: If the module isn't defined this will raise an argument error. You can see that behavior in action in iex:

```
iex(1)> "My.Fake.Worker" |> String.split(".") |> Module.safe_concat()
** (ArgumentError) argument error
    :erlang.binary_to_existing_atom("Elixir.My.Fake.Worker", :utf8)
    (elixir) src/elixir_aliases.erl:119: :elixir_aliases.safe_concat/1
```

For this to work safely we need to use `Module.concat/1`, followed by `Code.ensure_loaded?/1` and `function_exported?/3`. It is important that this function call doesn't fail because it won't be caught up the chain, and if it isn't caught then the job won't be retried.

balexand: I see. I thought the `ArgumentError` was coming from `apply`.

I've pushed an update. Now the `ArgumentError` from `to_module` is being handled. When `nil` is passed as the first argument to `function_exported?` then it returns `false`.

Is it necessary to handle the case where the module is available but doesn't export the `backoff` function? If not then this function could be simplified further.

balexand: Here's a simplified implementation if it is not necessary to check if the module exports `backoff`. It seems like this would be sufficient since `use Oban.Worker` defines the `backoff` function. The only way this could fail is if `module` is a valid module but it isn't an `Oban.Worker`.

```elixir
  defp worker_backoff(%Job{attempt: attempt, worker: worker}) do
    try do
      to_module(worker)
    rescue
      ArgumentError -> nil
    end
    |> case do
      nil -> Worker.default_backoff(attempt)
      module -> module.backoff(attempt)
    end
  end
```

sorentwo: The `Oban.Worker` behaviour specifies `backoff/1` as a required callback, so I think this is a safe assumption.

sorentwo:
The final version of this strikes the proper balance between safe execution and exposing bugs:

- Module not found -> Default
- Module found, but doesn't satisfy the worker behaviour -> Default
- Module found, backoff found, backoff raises error -> Error

The user is entirely in control of the `backoff/1` function, if there is an error they should get a crash report.

Thanks! 💛

balexand:
I've been enjoying using Oban. Thanks for building it 😄

PR #53, opened 2019-08-20
Fix prefix usage in V1 migration `down/1` statements

bryanlep:
I ran into an issue when trying to ensure migrations could be rolled back cleanly.

One question I had was about the test migrations. By default a call to `Oban.Migrations.down/1` only rolls back a single version, so the test migrations never call V1 during a rollback. I added the `version: 1` option to both migrations explicitly in their `down/1` calls, but please let me know if this is not an acceptable solution.

Thank you for this library, it's great! 👍

### DIFF
--- a/lib/oban/migrations.ex
+++ b/lib/oban/migrations.ex
@@ -140,10 +140,10 @@ defmodule Oban.Migrations do
     end
 
     def down(prefix) do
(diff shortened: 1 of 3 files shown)


sorentwo:
This is absolutely an acceptable solution! I always dropped everything and recreated the world so I didn't encounter the issue.

Thanks for the fix 💛

PR #56, opened 2019-08-23
Replace Advisory Locks with Pulse Tracking

sorentwo:
This is still lacking documentation, but the implementation is ready to go. A brief summary of the changes:

* Remove all use of advisory locks
* Add an `attempted_by` column to `oban_jobs`, used to track the node, queue and producer nonce that attempted to execute a job
* Add a new `oban_beats` table, used by producers to publish "pulse" information including the node, queue, running jobs and other information previously published by gossip notifications
* Remove gossip notifications entirely
* Prune `oban_beats` older than an hour—this isn't configurable yet, but it may be in the future
* Use beat information to rescue orphaned jobs—the tests explain this fairly well, essentially any job that belongs to a producer without a beat in the last minute will be rescued. Again, this value isn't configurable yet.
* Periodically rescue orphans, by default a rescue is attempted every minute, but it is configurable.


AndrewDryga:
@sorentwo Hey, I just spotted this PR and just wondering why advisory locks did not work for you?

benwilson512:
There were a few reasons. A fundamental challenge with advisory locks is that it's a global namespace, so you have to be careful about conflicting with advisory locks the application may already be using.

More importantly though, advisory locks forced Oban had to choose between serializing all lock access through a single process and using session locks, or doing shared locks by each worker process. The single process approach creates a bottleneck, and the shared lock situation makes releasing the locks tricky because the workers don't own individual database connections, but rather use the ecto repo pool. Consequently attempts to release them often lead to error messages that the lock wasn't owned by that exact connection.

There is probably a design that could have been used that had a pool of dedicated connections and session locks, but that doesn't get around the core issues advisory locks present in general.

sorentwo:
> My general feeling is that binding a connection to a worker process is a way to go. Having 3600 commits/hour per worker to the DB for heartbeats sounds expensive especially if your DB is not overprovisioned, to a point where it's easier to set up a separate DB and "Fewer Dependencies" along with "Transactional Control" goes away with that.

Individual connections have a constant fixed cost on the database. Typically each connection requires 5-10mb of RAM even when it is inactive. If each queue's producer needs a dedicated process to handle jobs then the number of distinct queues is limited by the database's memory. Using the application's pool is much friendlier to smaller databases than having additional dedicated connections—especially if queues spend any time inactive.

Aside from eliminating the use of advisory locks there are a few other benefits to using heartbeats:

1. We can rescue orphaned jobs after a producer crashes, not only after the entire vm reboots. The advisory lock doesn't hold any real information (like when it was taken or what took it), so that wasn't feasible.
2. We have statistics about which nodes are alive, when they went down, what jobs they processed, etc.

> And then when it's overloaded and some DDL blocks the table all jobs could get rescheduled.

I'm not going to claim that a DDL that blocks the `oban_beats` table is impossible, but it is extremely unlikely. Aside from the relatively small size of the table (~ 3.5k rows per node/queue) there won't be an Oban migration where we do a locking operation on the table.

> Otherwise, on a loaded instance, you will often hit pool timeout issues. We work around this by having a separate repo for BG jobs (and a separate pool), but having a connection checkout (eg. checkout_and_link/0) for a process in Ecto would be really nice.

Chances are that if db_connection/ecto provided a `checkout_and_link` I may not have gone this route. However, at this point I'm really pleased with the results and what the `oban_beats` table provides.

Historic introspection is one of the most important aspects of Oban, and persistent heartbeats really enhance introspecting.

PR #81, opened 2019-10-17
update version in table comment also fro V5

albertored:
I this was intentional just close the PR, it was faster to push a PR than to open an issue :-)


sorentwo:
@albertored Nice catch! This was definitely a mistake. Fortunately, due to the fact that each migration is idempotent the lack of a comment won't break anything. I'm adding tests to verify this behavior and prevent this from regressing.

PR #104, opened 2019-11-19
keep permitted fields in Job.to_map / insert_all

### LINKED ISSUE #103: Oban.insert_all does not propagate all Job attributes
### Precheck

- Do a quick search and make sure the bug has not yet been reported ✅ 
- For support, favor using the Elixir Forum, Slack, IRC, etc. ✅ 
- Be friendly and polite! ✅ 🤞

### Environment

- Oban Version: `v0.11.1`
- PostgreSQL Version: `v11.5`
   ```
   $ elixir --version
   Erlang/OTP 22 [erts-10.5.5] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [hipe] [dtrace]
   
   Elixir 1.9.4 (compiled with Erlang/OTP 22)
   ```

### Current Behavior

I have a worker defined like this:

```elixir
defmodule MyApp.Workers.UpdateThing do
  use Oban.Worker, queue: "my_queue", max_attempts: 3
  require Logger

  @impl Oban.Worker
  def perform(args, job) do
    # ...
  end
end
```

Note the `max_attempts: 3`, which AFAICT follows the instructions in the docs.

In the only place in our code where I actually enqueue his job, I run:

```elixir
some_list
|> Enum.map(fn data ->
  # args_map_from_data is just a plain map, i.e. there is only one arg to this function call:
  MyApp.Workers.UpdateThing.new(args_map_from_data)
end)
|> Oban.insert_all()
```

Note that I am _not_ explicitly setting `max_attempts` while enqueueing this job.

### Expected Behavior

Despite my only usage of `max_attempts` declaring that it should be 3 for this job, we have a bunch of failed instances of this job that have failed 15+ times and have a `max_attempts: 20` set in the database (this query was run a bit after I deleted all instances with > 6 attempts):

```ex
myapp::DATABASE=> SELECT worker, state, count(*) FROM oban_jobs WHERE attempt > 5 AND max_attempts > 5 AND worker = 'MyApp.Workers.UpdateThing' GROUP BY worker, state ORDER BY count DESC;
          worker           |   state   | count 
---------------------------+-----------+-------
 MyApp.Workers.UpdateThing | retryable |    49

myapp::DATABASE=> SELECT worker, state, attempt, max_attempts FROM oban_jobs WHERE attempt > 5 LIMIT 10;
           worker           |   state   | attempt | max_attempts 
----------------------------+-----------+---------+--------------
 MyApp.Workers.UpdateThing  | retryable |      10 |           20
```

I'm not sure how this could have happened 🤔 Even when testing job creation on `iex`, I can see it appears to be setting `max_attempts` correclty:

```elixir
MyApp.Workers.UpdateThing.new(%{foo: "bar"})
#Ecto.Changeset<
  action: nil,
  changes: %{
    args: %{foo: "bar"},
    max_attempts: 3,
    queue: "metrc",
    worker: "MyApp.Workers.UpdateThing"
  },
  errors: [],
  data: #Oban.Job<>,
  valid?: true
>
```

And yet, they sure seem to be resetting the `max_attempts` value. The best guess I have is that this is somehow being thrown away when the job is rescued due to it raising an exception (rather than cleanly erroring with `{:error, err}`).

bgentry:
`Oban.insert_all` uses `Job.to_map/1` before actually inserting `Job` records into the database. Prior to this commit, that function was only returning a limited set of fields from the `Job` struct, meaning certain fields like `max_attempts` were lost (#103).

To fix this, we include all permitted fields when converting a `Job` to a map, preserving any optional values that were either specified by the user or came via Worker defaults.

Fixes #103.

### DIFF
--- a/test/oban/job_test.exs
+++ b/test/oban/job_test.exs
@@ -48,24 +48,44 @@ defmodule Oban.JobTest do
   end
 
   describe "to_map/1" do
+    @keys_with_defaults ~w(args attempt errors max_attempts queue state)a
+    defp to_keys(opts) do
+      %{}
+      |> Job.new(opts)
+      |> Job.to_map()
+      |> Map.keys()
+    end
+
+    defp default_keys_with(extra_keys) do
+      (@keys_with_defaults ++ extra_keys) |> Enum.sort() |> Enum.uniq()
+    end
+
     test "nil values are not retained" do
-      to_keys = fn opts ->
-        %{}
-        |> Job.new(opts)
-        |> Job.to_map()
-        |> Map.keys()
-      end
+      assert to_keys([]) == @keys_with_defaults
+      assert to_keys(worker: MyWorker) == default_keys_with([:worker])
 
-      assert to_keys.([]) == [:args, :queue, :state]
-      assert to_keys.(worker: MyWorker) == [:args, :queue, :state, :worker]
+      assert to_keys(schedule_in: 1, worker: MyWorker) ==
+               default_keys_with([:scheduled_at, :worker])
+    end
 
-      assert to_keys.(schedule_in: 1, worker: MyWorker) == [
-               :args,
-               :queue,
-               :scheduled_at,
-               :state,
-               :worker
-             ]
+    test "retains all non-nil permitted attributes" do
+      attrs = [
+        attempt: 1,
+        attempted_by: ["foo"],
+        attempted_at: DateTime.utc_now(),
+        completed_at: DateTime.utc_now(),
+        errors: [%{err: "baz"}],
+        inserted_at: DateTime.utc_now(),
+        max_attempts: 5,
+        queue: "default",
+        scheduled_at: DateTime.utc_now(),
+        state: "scheduled",
+        worker: "Fake"
+      ]
+
+      map = %{foo: "bar"} |> Job.new(attrs) |> Job.to_map()
+      assert map[:args] == %{foo: "bar"}
+      Enum.each(attrs, fn {key, val} -> assert map[key] == val end)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


test/oban/job_test.exs
```diff
@@ -56,16 +56,37 @@ defmodule Oban.JobTest do
...
+      keys_with_defaults = ~w(args attempt errors max_attempts queue state)a
 
-      assert to_keys.(schedule_in: 1, worker: MyWorker) == [
-               :args,
-               :queue,
-               :scheduled_at,
-               :state,
-               :worker
-             ]
+      default_keys_with = fn extra_keys ->
```
sorentwo: Nice solution. Can we pull this (and the existing `to_keys` function) out into private functions?

test/oban/job_test.exs
```diff
@@ -56,16 +56,37 @@ defmodule Oban.JobTest do
...
+        max_attempts: 5,
+        queue: "default",
+        scheduled_at: DateTime.utc_now(),
+        state: "scheduled",
+        worker: "Fake"
+      ]
+
+      map = %{foo: "bar"} |> Job.new(attrs) |> Job.to_map()
+      assert map[:args] == %{foo: "bar"}
+      attrs |> Enum.each(fn {k, v} -> assert map[k] == v end)
```
sorentwo: Extremely thorough, love it. A couple minor nitpicks:

1. Please avoid piping a single argument
2. Please avoid single character variables

```suggestion
      Enum.each(attrs, fn {key, val} -> assert map[key] == val end)
```

bgentry:
Changes made @sorentwo, thanks :v:

sorentwo:
Thank you!

bgentry:
Thanks for merging @sorentwo :v: I realized I forgot to make a changelog entry for this, do you tend to do that yourself or do you want me to do so?

sorentwo:
Feel free to add a changelog entry. Typically I'll add one the next time I'm in there or before I make a release, but I'll gladly accept a PR.

PR #107, opened 2019-11-20
[#101] Allow to run scheduled jobs with drain_queue/1

### LINKED ISSUE #101: Allow to run scheduled jobs with drain_queue/1
Hey @sorentwo,

here's another odd feature request 🍰 

### Problem

In integration tests, we would like to be able to run scheduled jobs by calling `drain_queue`.

### Solution

Not sure exactly, haven't read to much of the `Producer` yet

### Current Workaround

We currently have a test helper that looks somewhat like this:

```elixir
  @doc """
  Allows to flush scheduled workers from Oban queues.
  """
  def drain_queue_including_scheduled(queue) do
    queue |> to_string() |> clear_scheduled()

    Oban.drain_queue(queue)
  end

  defp clear_scheduled(queue) do
    from(j in Oban.Job, where: j.queue == ^queue and j.state == "scheduled")
    |> MyRepo.update_all(
      set: [
        scheduled_at: DateTime.utc_now(),
        state: "available"
      ]
    )
  end
end
```

What do you think?

cheers,
malte

maltoe:
This PR will add a way to use `Oban.drain_queue` to execute scheduled jobs.

Fixes #101

## TODO

- [x] Solve interface issue
- [x] Add more documentation, enrich commit msg
- [x] Add changelog entry

## Interface Issue [Resolved]

@sorentwo Pushed this in a WIP state for now to ask you for advice. I wanted to give `drain_queue/2` a `with_scheduled` option, but realized that it won't work with the current way of how the `Oban` module is handed as the first (default) argument. 

```elixir
def drain_queue(name \\ __MODULE__, queue, opts \\ [])
end

# This will put :default into name and the keyword list into queue.
drain_queue(:default, with_scheduled: true)
```

Also see the failing test case.

I currently see a few ways to go about this:

1) Mission abort. Don't add a flag to `drain_queue/2`, but instead add a `drain_queue_with_scheduled/2`. Or possible just let the user use `Query.stage_all_scheduled_jobs/2` (exposed through `Oban`).
2) Solve this with heavy guarding of the function, i.e. basically "shift" the arguments when the second one is a list. Least amount of work, but IMO pretty ugly and fragile.
3) I'm assuming the idea behind this whole `name` argument and defaulting to `Oban` is to have multiple different configurations per app? In this case one could also go through with it fully and move functionality from `Oban` to something like `Oban.API` which `Oban` uses. `Oban.API` defines simple redirect functions like 

```
defmacro __using__(_opts) do
  quote do
    def drain_queue(name, opts) do
      Oban.API.drain_queue(__MODULE__, name, opts)
    end
  end
end
```

This is basically the `Ecto.Repo` approach. One would still have the "default" `Oban` module, but the user could also define their own API modules, e.g. `MyApp.Oban`.

Not sure if this is exactly what you were heading for ^^, but it's currently the most appealing solution to me :). Option 1 would also be fine, I guess.

[UPDATE] Resolved with `Oban.drain/2`.

---

cheers,
malte

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,12 @@ create index(
   Circuit breaker activity is logged by the default telemetry logger (both
   `:trip_circuit` and `:open_circuit` events).
 
+- [Oban] Expose `circuit_backoff` as a "twiddly" option that controls how long
+  tripped circuit breakers wait until re-opening.
+
+- [Oban] Add `drain_queue/3` to accept drain options. `with_scheduled: true`
+  allows draining scheduled jobs.
+
 ### Fixed
 
 - [Oban.Query] Avoid using prepared statements for all unique queries. This
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -309,7 +309,7 @@ defmodule Oban do
 
   During integration testing it may be necessary to run jobs because they do work essential for
   the test to complete, i.e. sending an email, processing media, etc. You can execute all
-  available jobs in a particular queue by calling `Oban.drain_queue/1` directly from your tests.
+  available jobs in a particular queue by calling `Oban.drain_queue/3` directly from your tests.
 
   For example, to process all pending jobs in the "mailer" queue while testing some business
   logic:
@@ -330,7 +330,7 @@ defmodule Oban do
   end
   ```
 
-  See `Oban.drain_queue/1` for additional details.
+  See `Oban.drain_queue/3` for additional details.
 
   ## Error Handling
 
@@ -588,7 +588,7 @@ defmodule Oban do
   end
 
   @doc """
-  Retreive the config struct for a named Oban supervision tree.
+  Retrieve the config struct for a named Oban supervision tree.
   """
   @doc since: "0.2.0"
   @spec config(name :: atom()) :: Config.t()
@@ -725,14 +725,49 @@ defmodule Oban do
     |> Query.insert_all_jobs(multi, multi_name, changesets)
   end
 
+  @type drain_option :: {:with_scheduled, boolean()}
+  @type drain_result :: %{success: non_neg_integer(), failure: non_neg_integer()}
+
+  @doc """
+  Synchronously execute all available jobs in a queue.
+
+  See `drain_queue/3`.
+  """
+  @doc since: "0.4.0"
+  @spec drain_queue(queue :: atom() | binary()) :: drain_result()
+  def drain_queue(queue) when is_queue(queue) do
+    drain_queue(__MODULE__, queue, [])
+  end
+
+  @doc """
+  Synchronously execute all available jobs in a queue.
+
+  See `drain_queue/3`.
+  """
+  @doc since: "0.4.0"
+  def drain_queue(queue, opts) when is_queue(queue) and is_list(opts) do
+    drain_queue(__MODULE__, queue, opts)
+  end
+
+  def drain_queue(name, queue) when is_atom(name) and is_queue(queue) do
+    drain_queue(name, queue, [])
+  end
+
   @doc """
   Synchronously execute all available jobs in a queue. All execution happens within the current
   process and it is guaranteed not to raise an error or exit.
 
   Draining a queue from within the current process is especially useful for testing. Jobs that are
   enqueued by a process when Ecto is in sandbox mode are only visible to that process. Calling
-  `drain_queue/1` allows you to control when the jobs are executed and to wait synchronously for
+  `drain_queue/3` allows you to control when the jobs are executed and to wait synchronously for
   all jobs to complete.
+
+  ## Example
+
+  Drain a queue with three available jobs, two of which succeed and one of which fails:
+
+      Oban.drain_queue(:default)
+      %{success: 2, failure: 1}
 
   ## Failures & Retries
 
@@ -747,20 +782,21 @@ defmodule Oban do
   * Use telemetry events to track success and failure
   * Check the database for jobs with errors
 
-  ## Example
+  ## Scheduled Jobs
 
-  Drain a queue with three available jobs, two of which succeed and one of which fails:
+  By default, `drain_queue/3` will execute all currently available jobs. In order to execute scheduled
+  jobs, you may pass the `:with_scheduled` flag which will cause scheduled jobs to be marked as
+  `available` beforehand.
 
-      Oban.drain_queue(:default)
-      %{success: 2, failure: 1}
+      # This will execute all scheduled jobs.
+      Oban.drain_queue(:default, with_scheduled: true)
+      %{success: 1, failure: 0}
   """
   @doc since: "0.4.0"
-  @spec drain_queue(name :: atom(), queue :: queue_name()) :: %{
-          success: non_neg_integer(),
-          failure: non_neg_integer()
-        }
-  def drain_queue(name \\ __MODULE__, queue) when is_atom(name) and is_queue(queue) do
-    Producer.drain(to_string(queue), config(name))
+  @spec drain_queue(name :: atom(), queue :: queue_name(), [drain_option()]) ::
+          drain_result()
+  def drain_queue(name, queue, opts) when is_atom(name) and is_queue(queue) and is_list(opts) do
+    Producer.drain(to_string(queue), config(name), opts)
   end
 
   @doc """
--- a/lib/oban/query.ex
+++ b/lib/oban/query.ex
@@ -69,12 +69,16 @@ defmodule Oban.Query do
     end)
   end
 
-  @spec stage_scheduled_jobs(Config.t(), binary()) :: {integer(), nil}
-  def stage_scheduled_jobs(%Config{prefix: prefix, repo: repo, verbose: verbose}, queue) do
+  @spec stage_scheduled_jobs(Config.t(), binary(), opts :: keyword()) :: {integer(), nil}
+  def stage_scheduled_jobs(%Config{} = config, queue, opts \\ []) do
+    %Config{prefix: prefix, repo: repo, verbose: verbose} = config
+
+    max_scheduled_at = Keyword.get(opts, :max_scheduled_at, DateTime.utc_now())
+
     Job
     |> where([j], j.state in ["scheduled", "retryable"])
     |> where([j], j.queue == ^queue)
-    |> where([j], j.scheduled_at <= ^utc_now())
+    |> where([j], j.scheduled_at <= ^max_scheduled_at)
     |> repo.update_all([set: [state: "available"]], log: verbose, prefix: prefix)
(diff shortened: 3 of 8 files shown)


lib/oban.ex
```diff
@@ -745,15 +745,33 @@ defmodule Oban do
...
   @doc since: "0.4.0"
-  @spec drain_queue(name :: atom(), queue :: atom() | binary()) :: %{
+  @spec drain_queue(name :: atom(), queue :: atom() | binary(), opts :: keyword()) :: %{
           success: non_neg_integer(),
           failure: non_neg_integer()
         }
-  def drain_queue(name \\ __MODULE__, queue)
-      when is_atom(name) and (is_atom(queue) or is_binary(queue)) do
-    Producer.drain(to_string(queue), config(name))
+  def drain_queue(name \\ __MODULE__, queue, opts \\ [])
```
sorentwo: This can be done seamlessly with manual clauses:

```elixir
  def drain_queue(queue) when is_atom(queue) or is_binary(queue) do
    drain_queue(__MODULE__, queue, [])
  end

  def drain_queue(name, queue) when is_atom(name) and (is_atom(queue) or is_binary(queue)) do
    drain_queue(name, queue, [])
  end

  def drain_queue(queue, opts) when (is_atom(queue) or is_binary(queue)) and is_list(opts) do
    drain_queue(__MODULE__, queue, opts)
  end

  def drain_queue(name, queue, opts)
      when is_atom(name) and (is_atom(queue) or is_binary(queue)) and is_list(opts) do
    Producer.drain(to_string(queue), config(name))
  end
```

It may also require some additional `drain_queue` specs.

maltoe: Yes, this is more or less what I meant to say in 2. Reads nice, but don't you find it a bit irritating as an interface? Specifically that there is a `drain_queue/2` that has the `queue` as the first or as the second argument. If I were to `@spec` and document this function, the console help text would look like

```elixir
iex(12)> h Oban.drain_queue/2

                          def drain_queue(name, queue)

  @spec drain_queue(atom() | binary(), atom() | binary() | list()) :: map()

Drain a queue. See drain_queue/3.
```

The headline shows the first implementation only, and the signature looks interesting, but very non-descriptive (i.e. it doesn't convey which combinations of types work).

sorentwo: Ah, now that I've re-read your second option I see that is what you were saying. I don't find it the most appealing interface, and it is a little bit fragile.

While we are technically pre 1.0, I'd prefer not to break things too drastically. Option three is interesting in general (and exactly what I did with Kiq). That seems like overkill for what we're trying to achieve here.

What about another option that isn't so ugly and is only slightly breaking—rename `drain_queue` as `drain`. That would change the `queue` argument to `opts` and enforce `queue` as a required argument.

```elixir
Oban.drain(queue: "default", with_scheduled: true)
```

maltoe: This sounds nice - would you want to move the documentation though?

maltoe: Pushed changes with a midway solution for documentation.

test/integration/draining_test.exs
```diff
@@ -6,18 +6,47 @@ defmodule Oban.Integration.DrainingTest do
   @oban_opts repo: Repo, queues: false
 
   test "all jobs in a queue can be drained and executed synchronously" do
-    start_supervised!({Oban, @oban_opts})
+    with_supervised_oban(fn ->
```
sorentwo: I like this pattern, but I'd rather keep it out of this PR and do it all in one go. Most of the integration tests would benefit from a `with_supervised_oban` helper defined in the `Oban.Case` module.

test/test_helper.exs
```diff
@@ -118,4 +120,12 @@ defmodule Oban.Case do
 
     {:ok, %{}}
   end
+
+  def seconds_from_now(seconds) do
```
sorentwo: Nice centralization here 👍

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+- [Oban] Add `drain/2` as a replacement of `drain_queue/2` for draining queues
```
sorentwo: This should go under the `### Added` header.

lib/oban.ex
```diff
@@ -739,21 +769,31 @@ defmodule Oban do
...
         }
-  def drain_queue(name \\ __MODULE__, queue)
-      when is_atom(name) and (is_atom(queue) or is_binary(queue)) do
-    Producer.drain(to_string(queue), config(name))
+  def drain(name \\ __MODULE__, opts)
+      when is_atom(name) and is_list(opts) do
+    queue = opts |> Keyword.fetch!(:queue) |> to_string()
+    config = config(name)
+
+    if Keyword.get(opts, :with_scheduled, false) do
```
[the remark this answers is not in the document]
maltoe: 👍 Implemented as `Producer.drain/3` with `drain(queue, config, opts)` -> Though if you considered `Producer.drain/2` internal and allowed changes to it, I think I'd prefer to hand off the `opts` from `Oban.drain/2` as they are and unpack there?

lib/oban/query.ex
```diff
@@ -70,11 +70,27 @@ defmodule Oban.Query do
   end
 
   @spec stage_scheduled_jobs(Config.t(), binary()) :: {integer(), nil}
-  def stage_scheduled_jobs(%Config{prefix: prefix, repo: repo, verbose: verbose}, queue) do
+  def stage_scheduled_jobs(%Config{} = config, queue) do
+    stage_scheduled_jobs(config, queue, fn queryable ->
+      where(queryable, [j], j.scheduled_at <= ^utc_now())
```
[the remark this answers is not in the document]
maltoe: Not sure if you had seen it, but I had a `@infinity_of_time DateTime.from_unix!(9_999_999_999)` in there for a while...

Not sure exactly, it didn't feel right. It would create a year 2248 problem.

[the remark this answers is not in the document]
maltoe: Maybe because it felt wasteful, to be precise. Even using the index, checking these timestamps for now reason.

maltoe:
@sorentwo PR updated.

> I'm not sure what to think about the the drain/2 and drain_queue/2 approach. In provides a single upgrade path without any breaking changes, but I don't think we need two ways to accomplish the same thing with such a slight difference in behavior.

^^ Agree. I wasn't sure how much to change the documentation to point to `drain/2` due to this. Would you want to `@deprecate` `drain_queue/2`?

maltoe:
Alternatively, if you feel `drain/2` is the wrong approach, I'd go for exposing `stage_all_scheduled_jobs` through `Oban`..

sorentwo:
> Alternatively, if you feel `drain/2` is the wrong approach, I'd go for exposing `stage_all_scheduled_jobs` through `Oban`..

I've thought about this a bit over the past couple of days:

- I think providing a built-in mechanism for draining scheduled jobs is the right approach. Users shouldn't need dig into the database to schedule things, and I really don't want to expose something as specific as `stage_scheduled_jobs` through the top level module.
- I still think that `Oban.drain_queue/3` is the cleanest and most obvious API, even if the function clauses are a bit tricky. The other queue level functions in the `Oban` module explicitly include `_queue` in the name and I'd like to keep it consistent.

Sorry to waffle on this. There are a lot of nuances in this issue which are difficult to judge without seeing them or experimenting.

maltoe:
Updated the PR again. I left out the `@spec` for `drain_queue/2`, as IMO it doesn't add much value. Having two `@spec` lines looks good in documentation but messes up dialyzer. Also turned around the clauses of `drain_queue/2`, so the `queue, opts` version is the headline in the docs.

You're probably right with your conclusion :) `drain/2` was a lot of interface change only to allow for this simple option.

CHANGELOG.md
```diff
@@ -42,6 +42,9 @@ create index(
 - [Oban] Expose `circuit_backoff` as a "twiddly" option that controls how long
   tripped circuit breakers wait until re-opening.
 
+- [Oban] Add `drain_queue/3` to accept drain options. `with_scheduled: true`
+  allows to drain scheduled jobs.
```
sorentwo:
```suggestion
  allows draining scheduled jobs.
```

lib/oban/query.ex
```diff
@@ -70,11 +70,25 @@ defmodule Oban.Query do
...
+  end
+
+  @spec stage_all_scheduled_jobs(Config.t(), binary()) :: {integer(), nil}
+  def stage_all_scheduled_jobs(%Config{} = config, queue) do
+    stage_scheduled_jobs(config, queue, & &1)
+  end
+
+  @spec stage_scheduled_jobs(Config.t(), binary(), fun()) :: {integer(), nil}
+  def stage_scheduled_jobs(%Config{} = config, queue, filter) do
+    %{prefix: prefix, repo: repo, verbose: verbose} = config
```
sorentwo: It seems redundant, but this needs to match on `%Config` for compile time key checking.

lib/oban/query.ex
```diff
@@ -70,11 +70,25 @@ defmodule Oban.Query do
...
   @spec stage_scheduled_jobs(Config.t(), binary()) :: {integer(), nil}
-  def stage_scheduled_jobs(%Config{prefix: prefix, repo: repo, verbose: verbose}, queue) do
+  def stage_scheduled_jobs(%Config{} = config, queue) do
+    stage_scheduled_jobs(config, queue, fn queryable ->
+      where(queryable, [j], j.scheduled_at <= ^utc_now())
+    end)
+  end
+
+  @spec stage_all_scheduled_jobs(Config.t(), binary()) :: {integer(), nil}
+  def stage_all_scheduled_jobs(%Config{} = config, queue) do
```
sorentwo: It doesn't seem right to have this extra function between `stage_scheduled_jobs/2` and `stage_scheduled_jobs/3`. The standard behavior is to check the `scheduled_at` value (which every queue does once a second), it shouldn't be a special case. The same compound index will always be used to build the query regardless of whether `scheduled_at` is checked or not.

What if the "far future" value was dynamically calculated at compile time rather than a fixed number? It makes no sense to schedule a job 10 years in the future (or one year for that matter).

maltoe: > The same compound index will always be used to build the query regardless of whether scheduled_at is checked or not.

🤔 Yeah, but it would still be more expensive to compare the timestamps while traversing the index rather than just take all after the `state` check?

Nonetheless, I changed it back to a far future timestamp (compile time, as before) and removed `stage_all_scheduled_jobs`.

👍 Thanks again for your reviews! Very insightful, I learned a bunch!

sorentwo:
Thanks for all the work on this one. The result looks great 💛

Issue #92, opened 2019-11-06
Pruner is losing the db connection from timeout


jc00ke:
### Environment

- Oban Version: 0.10.1
- PostgreSQL Version: 11.4
- Elixir & Erlang/OTP Versions (`elixir --version`): 1.8.1/21

### Current Behavior

![image](https://user-images.githubusercontent.com/18191/68343793-77a18f00-00a2-11ea-87b9-79633d878e08.png)

This is from the Heroku console.

### Expected Behavior

To live forever and ever.

### Notes

I'm going to be upgrading to `0.11` soon, but I don't think that'll have any bearing on this issue.

sorentwo:
Well that shouldn’t happen! The prune limit is configurable, so as a quick fix you can lower the limit to something like `1000` to prevent timeouts.

The real fix is to set the prune timeout to `:infinity`.

sorentwo:
My previous comment was wrong (not technically, just the wrong approach). The `prune_limit` setting was meant to prevent this exact issue, but it wasn't used for the beats table. Now it is!

ethangunderson:
We were seeing this too, but on jobs not beats.

We have ~60MM rows in our jobs table. The prune query wasn't being satisfied by any existing index and taking 20+ minutes.

To speed things up I added the following index:
``` sql
CREATE INDEX CONCURRENTLY oban_jobs_id_state_index ON oban_jobs (id desc) WHERE state = 'completed' OR state = 'discarded';
```

Let me know if you think this should be in Oban itself and I can make a PR.

sorentwo:
> We have ~60MM rows in our jobs table. The prune query wasn't being satisfied by any existing index and taking 20+ minutes.

That's a lot of jobs to have sticking around. I can see how the query wasn't working for that without the proper indexes. The limit doesn't help if the query itself isn't fast enough.

My initial instinct is that removing the `order by` clause allows a much faster index only scan of the table:

```
=> explain analyze select id from oban_jobs where state in ('completed', 'discarded') offset 1000 limit 10000;
                                                                                  QUERY PLAN
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
 Limit  (cost=52.80..580.00 rows=10000 width=8) (actual time=0.547..5.642 rows=10000 loops=1)
[...]

PR #108, opened 2019-11-20
Improve prune query with dedicated index

sorentwo:
The prune queries were unable to utilize any of the existing indexes and could be tremendously slow for large tables. This rewrites the queries to depend on the same columns and adds a dedicated partial index purely for pruning.

Addresses #92

/cc @ethangunderson

### DIFF
--- a/lib/oban/migrations.ex
+++ b/lib/oban/migrations.ex
@@ -4,7 +4,7 @@ defmodule Oban.Migrations do
   use Ecto.Migration
 
   @initial_version 1
-  @current_version 6
+  @current_version 7
   @default_prefix "public"
 
   def up(opts \\ []) when is_list(opts) do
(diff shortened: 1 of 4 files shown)


lib/oban/migrations.ex
```diff
@@ -333,4 +333,30 @@ defmodule Oban.Migrations do
...
+    @moduledoc false
+
+    use Ecto.Migration
+
+    import Oban.Migrations.Helper
+
+    def up(prefix) do
+      create_if_not_exists index(
+                             :oban_jobs,
+                             ["attempted_at desc", :id],
```
maltoe: Why do you need to index the `:id` column?

ethangunderson:
This is fantastic. Thank you for adding this in!

lib/oban/migrations.ex
sorentwo: Including `id` allows an index only scan because the only field we are selecting is the `id`.

maltoe: Thank you. TIL.

PR #113, opened 2019-11-26
Don't require migration modules be preloaded

stwf:
When called from Ecto.Migrator.run you can (apparently) encounter times when the app is running but the migration code isn't loaded, raising an ArgumentError in `safe_concat`.

 In the event of erroneous data there the code would bomb on the next call anyway so hopefully you don't feel like it's too risky.


sorentwo:
There was more discussion about this on Slack, which explained in some detail how this situation came up. I admit that I don't understand _how_ this actually fixes the issue, I recognize that it is the simplest way to fix it.

Thanks!

PR #116, opened 2019-11-29
Create a better error message for assert_enqueued

### LINKED ISSUE #115: Oban.Testing.assert_enqueued failing error message
### Environment

- Oban Version: 0.12.0
- PostgreSQL Version: 9.6.10
- Elixir & Erlang/OTP Versions (`elixir --version`): Elixir 1.8.1 (compiled with Erlang/OTP 21)

### Current Behavior

```
      assert_enqueued(
        worker: SystemMessageDispatchWorker,
        args: %{
          system_message_id: system_message.id,
          after_timestamp: NaiveDateTime.to_iso8601(expected_timestamp)
        }
      )
```
Gives the error:
```
Expected a job matching [worker: HobnobMessages.Workers.SystemMessageDispatchWorker, args: %{"after_timestamp" => "2019-01-03T00:00:00", "system_message_id" => "55d037c9-5be5-4b0b-94ca-8109cb1835fb"}] to be enqueued. Found: [%{args: %{"after_timestamp" => "2019-01-03T00:00:00.000000", "system_message_id" => "55d037c9-5be5-4b0b-94ca-8109cb1835fb"}, worker: "HobnobMessages.Workers.SystemMessageDispatchWorker"}]
```
When can be written like (ignore the autogenerated id's):

      # Expected a job matching
      [
        worker: HobnobMessages.Workers.SystemMessageDispatchWorker,
        args: %{
          after_timestamp: "2019-01-03T00:00:00",
          system_message_id: "629d1249-8d6f-4ce8-9161-93a937434360"
        }
      ]

      # found
      [
        %{
          args: %{
            "after_timestamp" => "2019-01-03T00:00:00.000000",
            "system_message_id" => "629d1249-8d6f-4ce8-9161-93a937434360"
          },
          worker: "HobnobMessages.Workers.SystemMessageDispatchWorker"
        }
      ]

### Expected Behavior

I'd expect the "found" value to be in the same shape as the "expected" value. Right now the expected value has the shape: `[worker: MyWorker, args: myargs]`, but the found value has the shape: `[%{worker: MyWorker, args: myargs}]` and that nested map threw me off of the real problem in my test (which is that the timestamps don't match)

axelson:
Now the error message will look like (and it includes pretty printing):
```
Expected a job matching:

%{args: %{dest: "other_node"}, worker: Ping}

to be enqueued. Instead found:

[%{args: %{"dest" => "some_node"}, worker: "Ping"}]
```

This should make it more easy to compare changes. Additionally the options are rendered as a map instead of keyword list to better match what the jobs look like.

Fixes #115

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
-- Better error message for `Oban.Testing.assert_enqueued/2`
+- [Oban.Testing] Better error message for `Oban.Testing.assert_enqueued/2`
 
 ## [0.12.0] – 2019-11-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
+- Better error message for `Oban.Testing.assert_enqueued/2`
+
 ## [0.12.0] – 2019-11-26
 
 **Migration Optional (V7)**
--- a/lib/oban/testing.ex
+++ b/lib/oban/testing.ex
@@ -158,10 +158,19 @@ defmodule Oban.Testing do
   @doc since: "0.3.0"
   @spec assert_enqueued(repo :: module(), opts :: Keyword.t()) :: true
   def assert_enqueued(repo, [_ | _] = opts) do
-    assert get_job(repo, opts),
-           "Expected a job matching #{inspect(opts)} to be enqueued. Found: #{
-             inspect(available_jobs(repo, opts))
-           }"
+    job = Enum.into(opts, %{})
+
+    error_message = """
+    Expected a job matching:
+
+    #{inspect(job, pretty: true)}
+
+    to be enqueued. Instead found:
+
+    #{inspect(available_jobs(repo, opts), pretty: true)}
+    """
+
+    assert get_job(repo, opts), error_message
   end
 
   @doc """
--- a/lib/oban/testing.ex
+++ b/lib/oban/testing.ex
@@ -158,12 +158,10 @@ defmodule Oban.Testing do
   @doc since: "0.3.0"
   @spec assert_enqueued(repo :: module(), opts :: Keyword.t()) :: true
   def assert_enqueued(repo, [_ | _] = opts) do
-    job = Enum.into(opts, %{})
-
     error_message = """
     Expected a job matching:
 
-    #{inspect(job, pretty: true)}
+    #{inspect(Map.new(opts), pretty: true)}
 
     to be enqueued. Instead found:
 
--- a/test/oban/testing_test.exs
+++ b/test/oban/testing_test.exs
@@ -43,6 +43,27 @@ defmodule Oban.TestingTest do
 
       assert_enqueued worker: Ping, scheduled_at: {seconds_from_now(69), delta: 10}
     end
+
+    test "prints a helpful error message" do
+      insert_job!(%{dest: "some_node"}, worker: Ping)
+
+      try do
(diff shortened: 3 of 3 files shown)


lib/oban/testing.ex
```diff
@@ -158,10 +158,19 @@ defmodule Oban.Testing do
   @doc since: "0.3.0"
   @spec assert_enqueued(repo :: module(), opts :: Keyword.t()) :: true
   def assert_enqueued(repo, [_ | _] = opts) do
-    assert get_job(repo, opts),
-           "Expected a job matching #{inspect(opts)} to be enqueued. Found: #{
-             inspect(available_jobs(repo, opts))
-           }"
+    job = Enum.into(opts, %{})
```
axelson: This probably isn't the best variable name

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
+- Better error message for `Oban.Testing.assert_enqueued/2`
```
sorentwo: The CHANGELOG entries usually include the module:

```suggestion
- [Oban.Testing] Better error message for `Oban.Testing.assert_enqueued/2`
```

lib/oban/testing.ex
sorentwo: Maybe not, since it is a map and not a job struct. It could be done inline instead, to avoid naming it entirely:

```elixir
    #{inspect(Map.new(opts), pretty: true)}
```

test/oban/testing_test.exs
```diff
@@ -43,6 +43,27 @@ defmodule Oban.TestingTest do
...
       assert_enqueued worker: Ping, scheduled_at: {seconds_from_now(69), delta: 10}
     end
+
+    test "prints a helpful error message" do
+      insert_job!(%{dest: "some_node"}, worker: Ping)
+
+      try do
+        assert_enqueued worker: Ping, args: %{dest: "other_node"}
+      rescue
+        error in [ExUnit.AssertionError] ->
```
sorentwo: Ah, so _this_ is what you were talking about on Slack. Looks great 👍

test/oban/testing_test.exs
```diff
@@ -43,6 +43,27 @@ defmodule Oban.TestingTest do
 
       assert_enqueued worker: Ping, scheduled_at: {seconds_from_now(69), delta: 10}
     end
+
+    test "prints a helpful error message" do
```
sorentwo:
```suggestion
    test "printing a helpful error message" do
```

test/oban/testing_test.exs
```diff
@@ -43,6 +43,27 @@ defmodule Oban.TestingTest do
...
+      insert_job!(%{dest: "some_node"}, worker: Ping)
+
+      try do
+        assert_enqueued worker: Ping, args: %{dest: "other_node"}
+      rescue
+        error in [ExUnit.AssertionError] ->
+          expected = """
+          Expected a job matching:
+
+          %{args: %{dest: "other_node"}, worker: Ping}
```
sorentwo: I wonder if the difference between the worker names will trip people up. The expected version will typically be a proper module name, but the value from the database will always be a string.

lib/oban/testing.ex
axelson: Okay, I like that :+1:

test/oban/testing_test.exs
axelson: Haha, yeah it was. Was able to find a decent example from the ex_unit tests

test/oban/testing_test.exs
axelson: Yeah it might, would probably be nice to convert the module if it exists already as an atom.

sorentwo:
We can address the module differences if it ever comes up. This is a great improvement! 💛

axelson:
Great, thanks!

PR #117, opened 2019-11-29
Add mix test.setup alias to make it easier to get started with Oban development

axelson:
This PR makes it easier for a developer to get started with oban development since there is less that needs to be typed in to create the test database. Alternatively we could make it even easier by creating the alias as `test` instead of `test.setup` (for my own applications I prefer to have them separate so that the migrations aren't checked every time tests are run).

Adding `config :oban, ecto_repos: [Oban.Test.Repo]` should be fine since any application depending on oban doesn't pull in its configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,11 +464,7 @@ database named `oban_test`. Follow these steps to create the database, create
 the database and run all migrations:
 
 ```bash
-# Create the database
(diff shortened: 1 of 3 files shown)


sorentwo:
Great ergonomic improvement! Thanks 💛

axelson:
Thanks for merging!

PR #136, opened 2019-12-20
Debounce producer dispatch for throughput

sorentwo:
Producers may become overloaded under high insert/execute load when processes complete quickly. This is due to a flood of `DOWN` messages from completed jobs, which triggers an immediate dispatch. Under high load the "demand" (concurrency limit - running jobs) hovers around 1, so every dispatch only retrieves one job. This is wasteful, unnecessarily hard on the database, and floods the producer's message queue.

This introduces dispatch debouncing for each producer. Rather than fetching jobs every time dispatch is called it waits a small period of time (5ms by default) to allow demand to accumulate. According to benchmarks this has a huge impact on overall throughput: without debouncing a producer maxes out around ~1000 jobs/sec, with a 5ms debounce producers can handle ~5000 jobs/sec and have an empty message queue.

---

Still in progress:

- [x] Document processing throughput and hard limits
- [x] Document configuration
- [x] Eliminate flakiness from test runs (leading edge debounce issues)

---

Here are some numbers/charts from continuous insertion benchmarks. The benchmark inserts 20 jobs every 10ms for 10s and checks the number of available vs completed jobs in the database afterwards. The `x` value is the queue limit. Without debouncing there isn't a predictable increase in throughput as the queue size increases:

Screen Shot 2019-12-20 at 4 40 22 PM

With debouncing the overall throughput increases along with the queue size:

Screen Shot 2019-12-20 at 4 40 32 PM

/cc @christianjgreen @TheFirstAvenger

### DIFF
--- a/lib/oban/queue/producer.ex
+++ b/lib/oban/queue/producer.ex
@@ -22,6 +22,8 @@ defmodule Oban.Queue.Producer do
     @enforce_keys [:conf, :foreman, :limit, :nonce, :queue]
     defstruct [
       :conf,
+      :cooldown_ref,
+      :dispatched_at,
       :foreman,
       :limit,
       :name,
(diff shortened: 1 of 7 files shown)


sorentwo:
Changes are in and documented. I'm rather pleased with the results! In my benchmarks I was able to achieve 8,975 jobs/sec for a single queue.

lib/oban/queue/producer.ex
```diff
@@ -256,29 +264,58 @@ defmodule Oban.Queue.Producer do
...
+  defp start_jobs(jobs, %State{conf: conf, foreman: foreman}) do
+    for job <- jobs, into: %{} do
+      {:ok, pid} = DynamicSupervisor.start_child(foreman, Executor.child_spec(job, conf))
+
+      {Process.monitor(pid), {job, pid}}
+    end
+  end
+
+  defp unix_now do
+    DateTime.to_unix(DateTime.utc_now(), :millisecond)
```
christianjgreen: It would probably be safer to use monotonic time here: https://hexdocs.pm/elixir/System.html#monotonic_time/1

sorentwo: Great point. Changed it up 👍

PR #147, opened 2020-01-11
Add compile time verification of string key maps

sorentwo:
This adds a compile time hook to `Oban.Worker` to enforce string based keys when pattern matching args in `perform/2` function heads.


sorentwo:
Lack of compile time expansion makes this is too fragile an approach to be used in all workers.

PR #148, opened 2020-01-15
Fix doc typo

benwilson512:
In telemetry events, the metadata key is `:max_attempts` not `:max_attempt`.

Thanks for Oban, it's great!


sorentwo:
Great catch, thanks!

PR #150, opened 2020-01-20
Fix worker docs

sasa1977:
According to [the code](https://github.com/sorentwo/oban/blob/afa30971b8e58c44fd19ab1acf0fc0e3cb983a5a/lib/oban/queue/executor.ex#L39-L58) and [this readme section](https://github.com/sorentwo/oban/blob/afa30971b8e58c44fd19ab1acf0fc0e3cb983a5a/README.md#Defining-Workers), only `{:error, reason}` is considered as an error, while `:error` is considered as success.

Related, I think it would be a bit nicer if `perform` contract was tighter, requiring the client to explicitly indicate success by returning `:ok`. So basically something like:

```elixir
@callback perform(args :: Job.args(), job :: Job.t()) :: :ok | {:error, reason :: term}
```

What do you think?

### DIFF
--- a/lib/oban/worker.ex
+++ b/lib/oban/worker.ex
@@ -42,7 +42,6 @@ defmodule Oban.Worker do
   Any of these return values or error events will fail the job:
 
   * return `{:error, error}`
-  * return `:error`
   * an unhandled exception
   * an unhandled exit or throw
 


sorentwo:
> Related, I think it would be a bit nicer if `perform` contract was tighter, requiring the client to explicitly indicate success by returning `:ok`. So basically something like:
> 
> ```elixir
> @callback perform(args :: Job.args(), job :: Job.t()) :: :ok | {:error, reason :: term}
> ```
> 
> What do you think?

That's something I've considered for a long time. Originally _any_ result was considered a success, and that was eventually loosened to consider error tuples a failure. Tightening the contract is appealing because it would make dialyzer success typing useful, and it would force us to be conscious of `perform/2` return values.

There are a few downsides that I see with an explicit contract:

1. It is a breaking change, albeit an easy one to remedy. We're still pre-1.0, so that isn't too big an issue yet.
2. I often unit test `perform/2` and make assertions on the return value. That works better with a standard success tuple than with just an `:ok` response.

Allowing `:ok` or `{:ok, result :: term}` is preferable:

```elixir
@callback perform(args :: Job.args(), job :: Job.t()) :: :ok | {:ok, result :: term()} | {:error, reason :: term}
```

How does that look to you?

sorentwo:
(Merged this regardless of how things proceed)

sasa1977:
> it would force us to be conscious of perform/2 return values

Yes, that's exactly the idea! It leaves much less room for accidental errors, in particular unintentionally succeeding when the intention was to fail.

>     1. It is a breaking change, albeit an easy one to remedy. We're still pre-1.0, so that isn't too big an issue yet.

I'm not even sure it wouldn't be a breaking change from the runtime behaviour standpoint. If the return value is neither ok or success, we should still treat it as success, because that' a safer (and IMO a saner) choice than retrying the job. A warning should probably emitted in this case, and of course dialyzer would complain, but programs depending on older versions should behave the same.

> Allowing `:ok` or `{:ok, result :: term}` is preferable:
> 
> ```elixir
> @callback perform(args :: Job.args(), job :: Job.t()) :: :ok | {:ok, result :: term()} | {:error, reason :: term}
> ```
> 
> How does that look to you?

This is definitely an improvement over the current state, but a nitpicker in me is a bit disturbed :-) As a user, I'm wondering why do I need to return the result on success, i.e. what will oban do with it? From what I understand, it won't do anything, so this value in the interface is redundant, and you're only supporting it for test convenience, which doesn't seem like the optimal choice to me.

sorentwo:
> I'm not even sure it wouldn't be a breaking change from the runtime behaviour standpoint. If the return value is neither ok or success, we should still treat it as success, because that' a safer (and IMO a saner) choice than retrying the job. A warning should probably emitted in this case, and of course dialyzer would complain, but programs depending on older versions should behave the same.

That's a great point! I'm sold.

> This is definitely an improvement over the current state, but a nitpicker in me is a bit disturbed :-) As a user, I'm wondering why do I need to return the result on success, i.e. what will oban do with it? From what I understand, it won't do anything, so this value in the interface is redundant, and you're only supporting it for test convenience, which doesn't seem like the optimal choice to me.

Hmm. That's also a fair point. The entire purpose of a background job is to generate a side effect, which means that `perform/2` must _do_ something beyond returning a value. It makes testing slightly more expensive to check a database again, but that can be worked around by composing functions.

👍

sasa1977:
> It makes testing slightly more expensive to check a database again, but that can be worked around by composing functions.

If you want to test that perform callback did its job, you can use a side-effect like e.g. sending a message to the test process, or something like that?

sorentwo:
> If you want to test that perform callback did its job, you can use a side-effect like e.g. sending a message to the test process, or something like that?

Right, exactly. Sending messages, capturing stdout, reading from the db, checking an in-memory mailbox, etc.

Issue #149, opened 2020-01-20
Testing: allow custom namespaces when making assertions about enqueued jobs


milmazz:
### Is your feature request related to a problem?

Apparently you can't use `assert_enqueued/1`, `refute_enqueued/1`, and I guess this also applies for `all_enqueued/1`, with custom namespaces (a.k.a. prefixes).

Some unit tests were failing using the following configuration:

```elixir
# config/config.exs
config :my_app, Oban,
  repo: MyApp.Repo,
  prune: {:maxlen, 100_000},
  queues: [cache: 50],
  prefix: "metrics"
# config/test.exs
config :my_app, Oban,
  queues: false,
  prune: :disabled
```

Then I checked the current implementation for `assert_enqueued/1` and I noticed that it does not handle namespaces at the moment, so, as a workaround I changed the configuration for the `test` environment as follows:

```elixir
# config/test.exs
config :my_app, Oban,
  queues: false,
  prune: :disabled,
  prefix: "public"
```

and everything worked smoothly.

### Describe the Solution You'd Like

I think that `assert_enqueued/1` should handle the `prefix` passed to `Oban` or `Repo`, in the following example the prefix is `private`:

```elixir
  # test/integration/isolation_test.exs
  test "inserting and executing jobs with a custom prefix" do
    start_supervised!({Oban, prefix: "private", repo: Repo, queues: [alpha: 5]})
    job = insert_job!(ref: 1, action: "OK")

    assert_enqueued(%{ref: 1, action: "OK"})
    assert_enqueued(worker: Worker, queue: :alpha)

    assert Ecto.get_meta(job, :prefix) == "private"
    assert_receive {:ok, 1}

    :ok = stop_supervised(Oban)
  end
```

But the unit test will fail after the `insert_job!/1`.

### Describe Alternatives You've Considered

As I mentioned before, an easy workaround for me was to change the `test` environment configuration, but I think that's not the ideal solution, because I think that `assert_enqueued/1` should consider the given `prefix` to `Oban` or `Repo`. I would like to know your opinion about this feature request.

sorentwo:
> I can create a PR if you agree, but first I would like to know your opinion about this feature request.

You're absolutely right, passing a prefix should work. Please do open a PR!

PR #152, opened 2020-01-21
allow custom prefix when making assertions about enqueued jobs

milmazz:
Fixes: #149


lib/oban/testing.ex
```diff
@@ -3,7 +3,9 @@ defmodule Oban.Testing do
   This module simplifies making assertions about enqueued jobs during testing.
 
   Assertions may be made on any property of a job, but you'll typically want to check by `args`,
-  `queue` or `worker`.
+  `queue` or `worker`. If you're using namespacing through PostgreSQL schemas, also called
+  "prefixes" in Ecto, you should use the `prefix` option when doing assertions about enqueued
+  jobs during testing, by default the `prefix` option is `public`.
```
sorentwo:
```suggestion
  jobs during testing. By default the `prefix` option is `public`.
```

lib/oban/testing.ex
```diff
@@ -214,6 +222,11 @@ defmodule Oban.Testing do
     |> apply_where_clauses(fields_with_opts)
   end
 
+  defp extract_prefix(opts) do
+    {prefix, opts} = Keyword.pop(opts, :prefix)
```
sorentwo: No need for the `||` below, you can provide a default to `pop/3`:

```suggestion
    {prefix, opts} = Keyword.pop(opts, :prefix, "public")
```

test/integration/isolation_test.exs
```diff
@@ -20,6 +22,8 @@ defmodule Oban.Integration.IsolationTest do
 
     job = insert_job!(ref: 1, action: "OK")
 
+    assert [%Oban.Job{}] = all_enqueued(worker: Worker, prefix: "private")
```
sorentwo: Great place for this!

`Oban.Job` is aliased by `Oban.Case`:

```suggestion
    assert [%Job{}] = all_enqueued(worker: Worker, prefix: "private")
```

lib/oban/testing.ex
```diff
@@ -214,6 +222,11 @@ defmodule Oban.Testing do
...
+    {prefix || "public", opts}
```
sorentwo:
```suggestion
```

sorentwo:
Ok, never mind! I can't apply the changes and merge at the same time so I don't have permissions to make the changes 😞

milmazz:
@sorentwo no worries, great suggestions, already applied the changes.

PR #153, opened 2020-01-21
Tighten worker's perform/2 contract

sorentwo:
This lacks documentation updates and the warning message is a bit rough (as well as a possible performance issue). Opening this up for discussion before proceeding.

Inspired by discussion in #150 

/cc @sasa1977

### DIFF
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -1,6 +1,8 @@
 defmodule Oban.Queue.Executor do
   @moduledoc false
 
+  require Logger
+
   alias Oban.{Config, Job, Query, Worker}
 
   @spec child_spec(Job.t(), Config.t()) :: Supervisor.child_spec()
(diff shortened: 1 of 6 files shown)


test/oban/queue/executor_test.exs
```diff
@@ -0,0 +1,39 @@
...
+    test "raising, catching and error tuples are failures" do
+      assert {:failure, %Job{}, :exception, _, [_ | _]} = call_with_mode("raise")
+      assert {:failure, %Job{}, :throw, :no_reason, [_ | _]} = call_with_mode("catch")
+      assert {:failure, %Job{}, :error, "no reason", [_ | _]} = call_with_mode("error")
+    end
+
+    test "warning on unexpected return values" do
+      message = capture_log(fn -> {:success, %Job{}} = call_with_mode("warn") end)
+
+      assert message =~ "Expected Elixir.Oban.Queue.ExecutorTest.Worker.perform/2"
```
milmazz:
```suggestion
      assert message =~ "Expected #{__MODULE__}.Worker.perform/2"
```

lib/oban/queue/executor.ex
```diff
@@ -41,12 +43,25 @@ defmodule Oban.Queue.Executor do
...
 
-      _ ->
+      returned ->
+        Logger.warn(fn ->
+          """
+          Expected #{worker}.perform/2 to return :ok or {:error, reason}. Instead received:
+
+          #{inspect(returned, pretty: true)}
+
+          The job completed successfully.
```
sasa1977: Something like `The job will be considered as successful.`

lib/oban/queue/executor.ex
```diff
@@ -41,12 +43,25 @@ defmodule Oban.Queue.Executor do
     |> to_module()
     |> apply(:perform, [args, job])
     |> case do
+      :ok ->
```
TylerWitt: I would love to be able to also return a success tuple in the case that workers are returning data for ease of tests. Anything unexpected would still get caught and would log a warning (which I think is the spirit of the issue), but it allows an escape hatch for the current functionality in a clear way.

One thing off the top of my head that is easier to capture with the output from a worker is what contents were put into a zip file, as it saves us from having to open it to check, or otherwise persist that information.

sorentwo:
Thanks for all of the feedback and reviews 💛 .

I ended up compromising slightly to allow `perform/2` to return `:ok` or `{:ok, term}`; partially for ease of testing but primarily for smoother backward compatibility.

PR #156, opened 2020-01-25
[WIP] Job Timeout

benwilson512:
TODO:
- [ ] integration level test
- [ ] docs

### DIFF
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -36,9 +36,35 @@ defmodule Oban.Queue.Executor do
   end
 
   @doc false
-  def safe_call(%Job{args: args, worker: worker} = job) do
+  def safe_call(%Job{} = job) do
+    with {:ok, worker} <- worker_module(job) do
+      case worker.timeout(job) do
+        :infinity ->
+          perform_inline(worker, job)
+
+        timeout when is_integer(timeout) ->
+          perform_task(worker, job, timeout)
+      end
+    end
+  end
+
+  defp perform_task(worker, job, timeout) do
+    task = Task.async(fn -> perform_inline(worker, job) end)
+
+    case Task.yield(task, timeout) || Task.shutdown(task) do
+      # reply here could still be a job failure, or even an exception tuple
+      # it just indicates that the job didn't timeout
+      {:ok, reply} ->
+        reply
+
+      nil ->
+        {:current_stacktrace, stacktrace} = Process.info(self(), :current_stacktrace)
+        {:failure, job, :timeout, timeout, stacktrace}
+    end
+  end
+
+  defp perform_inline(worker, %{args: args} = job) do
     worker
-    |> to_module()
     |> apply(:perform, [args, job])
     |> case do
       {:error, error} ->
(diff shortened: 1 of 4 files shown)


lib/oban/queue/executor.ex
```diff
@@ -59,29 +85,28 @@ defmodule Oban.Queue.Executor do
 
   # Helpers
 
-  defp to_module(worker) when is_binary(worker) do
-    worker
-    |> String.split(".")
-    |> Module.safe_concat()
+  defp worker_module(%Job{worker: worker} = job) when is_binary(worker) do
```
benwilson512: I refactored `to_module` into this function because it was called in several places and each had to rescue the possible exception from `Module.safe_concat()`. This way we can format that error in one place, using it when necessary and ignoring it when we want (as in backoff)

lib/oban/queue/executor.ex
```diff
@@ -36,9 +36,35 @@ defmodule Oban.Queue.Executor do
...
+    task = Task.async(fn -> perform_inline(worker, job) end)
+
+    case Task.yield(task, timeout) || Task.shutdown(task) do
+      # reply here could still be a job failure, or even an exception tuple
+      # it just indicates that the job didn't timeout
+      {:ok, reply} ->
+        reply
+
+      nil ->
+        {:current_stacktrace, stacktrace} = Process.info(self(), :current_stacktrace)
```
benwilson512: The current stacktrace isn't super helpful here, maybe I should try to take it off the task pid? Not sure about the implications of that.

lib/oban/queue/executor.ex
```diff
@@ -36,9 +36,35 @@ defmodule Oban.Queue.Executor do
   end
 
   @doc false
-  def safe_call(%Job{args: args, worker: worker} = job) do
+  def safe_call(%Job{} = job) do
+    with {:ok, worker} <- worker_module(job) do
+      case worker.timeout(job) do
+        :infinity ->
+          perform_inline(worker, job)
```
benwilson512: If the timeout is `:infinity`, I preserve running the work in the current process which eliminates pointless copying to the task process.

lib/oban/worker.ex
```diff
@@ -138,6 +138,13 @@ defmodule Oban.Worker do
   """
   @callback backoff(attempt :: pos_integer()) :: pos_integer()
 
+  @doc """
+  Maximimum job execution time limit
+
+  Defaults to `:infinity`
+  """
+  @callback timeout(job :: Job.t()) :: pos_integer() | :infinity
```
sorentwo: The built in `timeout` type uses `non_neg_integer()` instead of `pos_integer()`. I doubt there is any practical difference between 0ms and 1ms, but I think the type should match.

Using the actual `timeout` type seems a little odd for a function named `timeout`.

lib/oban/queue/executor.ex
```diff
@@ -59,29 +85,28 @@ defmodule Oban.Queue.Executor do
...
-    module =
-      try do
-        to_module(worker)
-      rescue
-        ArgumentError -> nil
-      end
+  defp worker_backoff(%Job{attempt: attempt} = job) do
+    case worker_module(job) do
+      {:ok, module} ->
+        module.backoff(attempt)
```
sorentwo: This is a little riskier than the previous code because we aren't checking that the module exports `backoff/1`. We know the module is real, but that doesn't mean it is actually a worker. We have the same issue with `worker.timeout` now, so perhaps we can verify that it is a worker in `worker_module/1` instead?

lib/oban/queue/executor.ex
sorentwo: Nicely refactored 👍

lib/oban/queue/executor.ex
sorentwo: The stacktrace is only important for consistency, I don't think there is any meaningful information that we'd get out of the task either.

sorentwo:
Looking great so far!

lib/oban/queue/executor.ex
benwilson512: How about this then:

```
defp worker_backoff(%Job{attempt: attempt} = job) do
    {:ok, module} = worker_module(job)
    module.backoff(attempt)
  rescue
    _ ->
      Worker.default_backoff(attempt)
  end
```

benwilson512: Blah, the whole thing is a little fiddly. Here is the issue I have with the version I just posted, which also applies to the version in master too: If I don't include backoff, or I code it badly and it returns an invalid value, it's silently ignored. Let me do a version where we log some useful stuff for users maybe.

sorentwo: Let's skip that for now. I can do some refactoring and work on error messaging after this is merged.

sorentwo:
@benwilson512 Let me know if you're blocked on this or if there is anything I can do to help get this over the line.

lib/oban/queue/executor.ex
benwilson512: 👍 I'll focus on the core timing functionality then and get a test case in, and then yeah I think a general pass on safely calling callback functions may be worthwhile.

sorentwo:
@benwilson512 Recent refactoring in the `Executor` module caused a ton of conflicts with your work. I took bits of it and based d71f79c on it (with attribution to you in the commit).

benwilson512:
Hey sorry, ended up with a bunch of deadlines in the last 2 weeks and couldn't finish it. Thanks for handling!

Issue #144, opened 2020-01-07
Environment-Specific Cron config Reco?


coladarci:
Let's say I want to use your cron-like setup but have it behave differently in staging versus production (maybe run a different frequencies or turn off certain checks for various reasons)? 

We use releases + env vars to do this w/ other features, but given the complexity of the cron-syntax, it presents some interesting challenges.
 
Do you have a reco on how to pull this off?

Thanks!

sorentwo:
To run different workers/frequencies in staging I recommend dynamically selecting a different crontab when the application boots. I'd use env vars to determine which crontab to use, but I wouldn't try and parse anything from the environment.

Something like this in `application.ex`:

```elixir
  defp oban_opts do
    environment = System.get_env("ENVIRONMENT")

    :my_app
    |> Application.get_env(Oban)
    |> Keyword.put(:crontab, crontab(environment))
  end

  defp crontab("staging") do
    [
      {"* * * * *", MyApp.WorkerA},
      {"* * * * *", MyApp.WorkerB},
      {"* * * * *", MyApp.WorkerC}
    ]
  end

  defp crontab(_environment) do
    [
      {"0 * * * *", MyApp.WorkerA},
      {"* 6-9 * * *", MyApp.WorkerB},
      {"*/15 * * * *", MyApp.WorkerC}
    ]
  end
```

Let me know if that works or what you come up with. This could be a nice addition to the guides.

coladarci:
Thank you! This is interesting - the only thing that is less than ideal is that, ideally, this could be controlled externally so each env can configure the crons however they want by providing some sort of config file. I am looking into https://hexdocs.pm/mix/Mix.Tasks.Release.html#module-config-providers as a potentially way of maybe solving this...

sorentwo:
> I am looking into https://hexdocs.pm/mix/Mix.Tasks.Release.html#module-config-providers as a potentially way of maybe solving this...

Interesting. Keep me posted 👍

sorentwo:
There aren't any actionable items from this issue so I'm going to close it. Please post an update if you come up with a config provider solution. 💛

PR #172, opened 2020-02-15
Adding an example of release configuration

coladarci:
As per https://github.com/sorentwo/oban/issues/144 - while I didn't end up needing to change my `crontab` I did want to change concurrency of queues. 

If you think this is worth including, here it is!

Thanks again for your all your help recently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@
 - [Testing](#Testing)
 - [Error Handling](#Error-Handling)
 - [Instrumentation & Logging](#Instrumentation-and-Logging)
+- [Release Configuration](#Release-Configuration)
 - [Isolation](#Isolation)
 - [Pulse Tracking](#Pulse-Tracking)
 - [Troubleshooting](#Troubleshooting)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -790,6 +791,79 @@ of Oban's telemetry events. As in the example above, attach it within your
 For more details on the default structured logger and information on event
 metadata see docs for the `Oban.Telemetry` module.
 
+## Release Configuration
+
+While having the same `Oban` configuration for every environment might be fine, there are certainly times you might
+want to make changes for a specific environment. Maybe you want to increase or decrease a queue's concurrency, for example.
+
+If you are using Elixir Releases, this is straight forward to do using [Moduel Config Providers](https://hexdocs.pm/mix/Mix.Tasks.Release.html#module-config-providers ).
```
maartenvanvliet:
```suggestion
If you are using Elixir Releases, this is straight forward to do using [Module Config Providers](https://hexdocs.pm/mix/Mix.Tasks.Release.html#module-config-providers ).
```

milmazz:
> The reason I haven't merged it in is that I'm unsure about adding it to the README, it feels more like a guide.

I do agree with this, this is a great addition, but it describes a specific scenario for deployments. Given that Oban already depends on `ex_doc`, adding a "Guides" section it's easy. Let me know if you need help with this.

sorentwo:
> I do agree with this, this is a great addition, but it describes a specific scenario for deployments. Given that Oban already depends on `ex_doc`, adding a "Guides" section it's easy. Let me know if you need help with this.

Yes, absolutely! I'm planning on a few initial guides (Installation, First Worker, Testing, Deployment) and would love help with any and all of them.

milmazz:
@coladarci Now we have a `guides` section, maybe you should consider moving the _Release Configuration_ section into its own guide. More guides/recipes are in the works with #247 HTH

sorentwo:
I've converted this into a guide and merged it into master. Thanks @coladarci 💛

PR #178, opened 2020-02-21
Use Multi.name() typespec

### LINKED ISSUE #177: Typespec mismatch within Oban.insert_all/4
### Environment

- Oban Version: v1.1.0
- PostgreSQL Version: n/a
- Elixir & Erlang/OTP Versions (`elixir --version`): 

### Current Behavior

Using a value other than an atom with `Oban.insert_all/4` causes dialyzer to emit the following warning:

```elixir
The function call will not succeed.

Oban.insert_all(_multi :: any(), <<_::8, _::size(1)>>, _jobs :: any())

will never return since it differs in arguments with
positions 2nd from the success typing arguments:

(
  %Ecto.Multi{
    :names => MapSet.t(_),
    :operations => [{_, {_, _} | {_, _, _} | {_, _, _, _}}]
  },
  atom(),
  [
    %Ecto.Changeset{
      :action => atom(),
      :changes => %{atom() => _},
      :constraints => [map()],
      :data => %Oban.Job{...},
      ...
    }
  ]
}
```

The [typespec for `Oban.Query.insert_all_jobs/4`](https://github.com/sorentwo/oban/blob/v1.1.0/lib/oban/query.ex#L62) is too strict. It expect `name` to be an atom, even though `Oban.insert_all/4` allows for any `term()`.

### Expected Behavior

No dialyzer warning is emitted when using `Oban.insert_all/4` with a multi name other than an `atom()`.

mcrumm:


sorentwo:
Thanks for the fix!

PR #184, opened 2020-02-26
Add note about database connections when on Heroku

jordelver:
Explain that Oban is going to use one of your database connections in order to function, and that you should plan for this when running on Heroku in particular as the Postgres Hobby tier plans have a maximum of 20 connections available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -971,6 +971,8 @@ MyApp.Repo.all(
 
 ### Heroku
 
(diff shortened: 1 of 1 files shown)


sorentwo:
Looks great! Thanks 💛

jordelver:
Thanks for merging so quickly @sorentwo and for your help on the Elixir Forum 👍

PR #188, opened 2020-03-03
Fixed Oban.Queue.Executor failure type

anthonator:


anthonator:
I'm not sure why CI isn't running but you can check the results in my repo.

https://github.com/anthonator/oban/actions/runs/48846637

sorentwo:
Thanks for the fix! I was trying to bundle it with a fix to dialyzer caching (which I believe you also had a thread about on the forum).

CI isn't running here because it is only set to `push`, not `[push, pull_request]`. I'll push a change after this is in 👍

sorentwo:
Found it! Here is the comment I was referring to: https://elixirforum.com/t/caching-dialyzer-output-with-github-actions/27648/4?u=sorentwo

anthonator:
Ah, nice!

Although there's a typo on that thread 😬

It should be `:plt_core_path` and not `:plt_cor_path`.

PR #194, opened 2020-03-15
WIP: Add native retry functionality

v0idpwn:
Retrying a job is a very frequent procedure on production systems. I think this fits well as a library responsibility so a user doesn't have to update the jobs table himself.

P.S.: I didn't knew on which file I'd create the tests for this functionality. If you tell me so I'd be happy to write them.

### DIFF
--- a/lib/oban/query.ex
+++ b/lib/oban/query.ex
@@ -219,6 +219,36 @@ defmodule Oban.Query do
     :ok
   end
 
+  @spec force_retry(Config.t(), Job.t()) :: :ok
+  def force_retry(%Config{} = conf, %Job{} = job) do
+    %Config{prefix: prefix, repo: repo, verbose: verbose} = conf
+    %Job{attempt: attempt, id: id, max_attempts: max_attempts} = job
+
+    set =
+      if attempt >= max_attempts do
+        [state: "available", max_attempts: attempt + 1]
+      else
+        [state: "available"]
+      end
+
+    updates = [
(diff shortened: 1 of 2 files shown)


lib/oban/query.ex
```diff
@@ -219,6 +219,32 @@ defmodule Oban.Query do
...
   end
 
+  @spec force_retry(Config.t(), Job.t()) :: :ok
+  def force_retry(%Config{} = conf, %Job{} = job) do
+    %Config{prefix: prefix, repo: repo, verbose: verbose} = conf
+    %Job{attempt: attempt, id: id, max_attempts: max_attempts} = job
+
+    updates =
+      if attempt >= max_attempts do
+        [set: [state: "available", max_attempts: attempt + 1]]
```
sorentwo: You can use `GREATEST` in a fragment to handle this update, rather than relying on data from the job.

Also of note, the state is changed to `available`, but the `completed_at` and `discarded_at` timestamps aren't cleared. That could allow this to run successfully yet still have a `discarded_at` timestamp.

v0idpwn:
Thanks for your review. I'll work on this soon :)

sorentwo:
@v0idpwn Bump. Are you still interested in finishing up this PR? If not I'll close this and defer until a later time.

sorentwo:
This has languished for quite a long time. I'm going to close this out and we can revisit it in the future if desired.

Issue #138, opened 2020-01-04
Additional telemetry events


sorentwo:
The job execution and circuit breaker telemetry events are great and there are a number of other places where measurements would be helpful.

Add new telemetry events around functionality that interacts with the database:

- [ ] Inserting (`Oban`)
- [ ] Batch Inserting (`Oban`)
- [x] Pruning Jobs (`Oban.Pruner`)
- [x] Descheduling (`Oban.Queue.Producer`)
- [x] Dispatching (`Oban.Queue.Producer`)

Each of the new events needs documentation in `Oban.Telemetry` as well.

devstopfix:
I may be able to assist this - we are currently investigating if we can use oban in our production application that already uses telemetry. Looking good so far! I will keep in touch.

sorentwo:
@devstopfix That would be great. Let me know if you have any questions.

devstopfix:
@sorentwo Thank you - I will! The library looks great - I think it will fit our needs very well. I will take a look at this issue on Sunday or next week

sorentwo:
Closed with 6cdf6e7, skipping `insert` and `insert_all` instrumentation because they happen in client applications and there hasn't been any other movement on this for 6 months.

PR #195, opened 2020-03-15
Add some telemetry events

v0idpwn:
Related to #138.

This is a work in progress. Still missing integration tests and documentation.

### DIFF
--- a/lib/oban/pruner.ex
+++ b/lib/oban/pruner.ex
@@ -56,32 +56,29 @@ defmodule Oban.Pruner do
 
   defp prune(%State{conf: conf} = state) do
     %Config{repo: repo, verbose: verbose} = conf
-    start_mono = System.monotonic_time(:microsecond)
 
-    resp =
-      repo.transaction(
-        fn ->
-          if Query.acquire_lock?(conf, @lock_key) do
-            prune_beats(conf)
-            prune_jobs(conf)
-          end
-        end,
-        log: verbose
-      )
-
-    :telemetry.execute([:oban, :prune], %{duration: duration(start_mono)})
-
-    resp
+    repo.transaction(
+      fn ->
+        if Query.acquire_lock?(conf, @lock_key) do
+          prune_beats(conf)
+          prune_jobs(conf)
+        end
+      end,
+      log: verbose
+    )
   rescue
     exception in trip_errors() -> trip_circuit(exception, __STACKTRACE__, state)
   end
 
   defp prune_beats(conf) do
+    start_mono = System.monotonic_time(:microsecond)
     Query.delete_outdated_beats(conf, conf.beats_maxage, conf.prune_limit)
+    :telemetry.execute([:oban, :prune_beats], %{duration: duration(start_mono)})
   end
 
   defp prune_jobs(conf) do
     %Config{prune: prune, prune_limit: prune_limit} = conf
+    start_mono = System.monotonic_time(:microsecond)
 
     case prune do
       {:maxlen, length} ->
@@ -90,6 +87,8 @@ defmodule Oban.Pruner do
       {:maxage, seconds} ->
         Query.delete_outdated_jobs(conf, seconds, prune_limit)
     end
+
+    :telemetry.execute([:oban, :prune_jobs], %{duration: duration(start_mono)})
   end
 
   defp send_after(interval) do
--- a/lib/oban/pruner.ex
+++ b/lib/oban/pruner.ex
@@ -56,16 +56,22 @@ defmodule Oban.Pruner do
 
   defp prune(%State{conf: conf} = state) do
     %Config{repo: repo, verbose: verbose} = conf
+    start_mono = System.monotonic_time(:microsecond)
 
-    repo.transaction(
-      fn ->
-        if Query.acquire_lock?(conf, @lock_key) do
-          prune_beats(conf)
-          prune_jobs(conf)
-        end
-      end,
-      log: verbose
-    )
+    resp =
+      repo.transaction(
+        fn ->
+          if Query.acquire_lock?(conf, @lock_key) do
+            prune_beats(conf)
+            prune_jobs(conf)
+          end
+        end,
+        log: verbose
+      )
+
+    :telemetry.execute([:oban, :prune], %{duration: duration(start_mono)})
+
+    resp
   rescue
     exception in trip_errors() -> trip_circuit(exception, __STACKTRACE__, state)
   end
@@ -89,4 +95,6 @@ defmodule Oban.Pruner do
   defp send_after(interval) do
     Process.send_after(self(), :prune, interval)
   end
+
+  defp duration(start_mono), do: System.monotonic_time() - start_mono
 end
--- a/lib/oban/pruner.ex
+++ b/lib/oban/pruner.ex
@@ -5,7 +5,7 @@ defmodule Oban.Pruner do
 
   import Oban.Breaker, only: [open_circuit: 1, trip_errors: 0, trip_circuit: 3]
 
-  alias Oban.{Config, Query}
+  alias Oban.{Config, Query, Telemetry}
 
   @type option :: {:name, module()} | {:conf, Config.t()}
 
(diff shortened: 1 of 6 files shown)


lib/oban/pruner.ex
```diff
@@ -71,11 +71,14 @@ defmodule Oban.Pruner do
   end
 
   defp prune_beats(conf) do
+    start_mono = System.monotonic_time(:microsecond)
```
sorentwo: What do you think about putting this in the `Oban.Telemetry` module as a helper function? It could have a signature like this:

```elixir
@spec span(name :: atom(), fun :: (-> result :: term())) :: result
```

Something to standardize the measurement and naming would be nice.

v0idpwn: Agreed

test/integration/telemetry_test.exs
```diff
@@ -131,6 +122,19 @@ defmodule Oban.Integration.TelemetryTest do
     :telemetry.detach("oban-default-logger")
   end
 
+  test "span publishes given metric and metadata" do
+    :telemetry.attach("span-handler", [:oban, :test_metric], &Handler.handle/4, self())
+
+    assert :example_return = Telemetry.span(:test_metric, fn -> :example_return end)
+    assert :ok = Telemetry.span(:test_metric, fn -> :timer.sleep(5) end, %{example_metadata: 1})
```
sorentwo, line 129: Functionally they are identical, but `Process.sleep/1` is used everywhere else in the codebase:

```suggestion
    assert :ok = Telemetry.span(:test_metric, fn -> Process.sleep(5) end, %{example_metadata: 1})
```

test/integration/telemetry_test.exs
```diff
@@ -131,6 +122,19 @@ defmodule Oban.Integration.TelemetryTest do
...
+
+    assert_receive {:executed, :test_metric, _duration, %{}}
+    assert_receive {:executed, :test_metric, duration, %{example_metadata: 1}}
+    assert duration > 5
```
sorentwo, line 133: The `sleep` is for 5 milliseconds and the duration is reported in microseconds. I think asserting that this is greater than 0 is sufficient.

test/integration/pruning_test.exs
```diff
@@ -63,6 +64,43 @@ defmodule Oban.Integration.PruningTest do
     :ok = stop_supervised(Oban)
   end
 
+  test "pruning jobs publishes a metric" do
```
sorentwo, line 67: We don't need separate tests for the telemetry events. The tests effectively duplicate the other pruning tests, and I'd rather avoid all the setup. You can drop these entirely, IMO.

lib/oban/telemetry.ex
```diff
@@ -158,4 +158,19 @@ defmodule Oban.Telemetry do
       |> Jason.encode!()
     end)
   end
+
+  @doc """
+  Measure the time it took to run a function and publishes a telemetry
+  """
+  @spec span(name :: atom(), fun :: (() -> term()), metadata :: map()) :: term()
+  def span(name, fun, metadata \\ %{}) do
```
sorentwo, line 166: Reading through this now and seeing how it is used in practice I don't think `span` is the correct name. According to the telemetry.span proposal there would be `start` and `stop` events for each span call. That doesn't fit well with wrapping the execution of a single function so I believe this needs a different name. Something like `wrap` or even `execute`?

mcrumm: Hi @sorentwo! The way this is implemented could actually work well for a span, but the behaviour within this `span` function needs to be extended.  The proposal in beam-telemetry/telemetry#57 defines three events that make up a span: `:start`, `:stop`, and `:exception`.

lib/oban/telemetry.ex
```diff
@@ -158,4 +158,19 @@ defmodule Oban.Telemetry do
...
+    start_mono = System.monotonic_time(:microsecond)
+    result = fun.()
+
+    :telemetry.execute([:oban, name], %{duration: duration(start_mono)}, metadata)
+
+    result
```
mcrumm, line 172: To match more closely to the telemetry "spec", we could emit multiple events:

```elixir
    start_mono = System.monotonic_time(:microsecond)
    :telemetry.execute([:oban, name, :start], %{system_time: start_mono}, metadata)

    try do
      result = fun.()
      :telemetry.execute([:oban, name, :stop], %{duration: duration(start_mono)}, metadata)
      result
    catch
      kind, reason ->
        stacktrace = System.stacktrace()

        :telemetry.execute([:oban, name, :exception], %{duration: duration(start_mono)}, Map.merge(metadata, %{kind: kind, reason: reason, stacktrace: stacktrace}))

        :erlang.raise(kind, reason, stacktrace)
    end
```

v0idpwn: What do you think, @sorentwo?

mcrumm: FYI, I culled this example primarily from teamon/tesla#324, but there are multiple implementations linked to that beam-telemetry issue above that you can reference, too.

lib/oban/telemetry.ex
sorentwo: Thanks for linking to the proposal @mcrumm. I've followed the span discussion a bit, and that's why I suggested using `span` inside of Oban. My thought here was that I'm not sure these are "spans" because only a single function is being executed. However, seeing your support for it being a `span` and thinking about it as a way to maintain consistency, I like it.

lib/oban/telemetry.ex
sorentwo: The implementation looks good. I've modified it with a few changes which I'll call out below:

```elixir
  start_time = System.system_time()
  start_mono = System.monotonic_time(:microsecond)

  :telemetry.execute([:oban, name, :start], %{start_time: start_time}, metadata)

  try do
    result = fun.()

    :telemetry.execute([:oban, name, :stop], %{duration: duration(start_mono)}, metadata)

    result
  catch
    kind, reason ->
      :telemetry.execute(
        [:oban, name, :exception],
        %{duration: duration(start_mono)},
        Map.merge(metadata, %{kind: :error, error: error, stack: __STACKTRACE__})
      )

      :erlang.raise(kind, reason, __STACKTRACE__)
  end
```

1. Use `start_time` instead of `system_time` for consistency with other Oban telemetry
2. Use `__STACKTRACE__` instead because we don't support any versions prior to OTP 21
3. Use kind, error and stack for consistency with the other Oban telemetry

The inconsistency in 1 and 3 will need to be addressed, along with using native time instead of microseconds, in a future PR.

bryannaegele: @sorentwo on the exception, I'd highly encourage using `:stacktrace` for the key as we'll be looking for that key in the tracing handlers. If you have other places using the `:stack` convention, I would suggest duplicating to both keys for now and deprecate it. The standard convention we've outlined is `[:kind, :error, :stacktrace]`.

And yes, monotonic times are expected to be in native.

lib/oban/telemetry.ex
```diff
@@ -158,4 +158,19 @@ defmodule Oban.Telemetry do
...
+
+  @doc """
+  Measure the time it takes to run a function and publish corresponding telemetry events.
+  """
+  @spec span(name :: atom(), fun :: (() -> term()), metadata :: map()) :: term()
+  def span(name, fun, metadata \\ %{}) do
+    start_mono = System.monotonic_time(:microsecond)
+    result = fun.()
+
+    :telemetry.execute([:oban, name], %{duration: duration(start_mono)}, metadata)
```
bryannaegele, line 170: If y'all want to add an identifier, we highly suggest putting it into metadata and not as part of the event name. We've come to realize this pattern is causing us a lot of issues. My suggestion for the jobs is to just use `[:oban, :job, :start | :stop | :exception]` for your events.

v0idpwn:
I'll do the changes and push them here soon

lib/oban/telemetry.ex
sorentwo: Noted. The plan was to introduce span events just like that after everything was nailed down.

This PR is specifically around emitting other non-job related events, namely dispatching and pruning. Are you suggesting that those events shouldn’t be namespaced?

bryannaegele: Correct. Event names should be static and anything dynamic moved to a property in the metadata.

sorentwo: Ok. To try and clarify a few things.

Currently, _before_ this PR there are five telemetry events:

- `[:oban, :started]`
- `[:oban, :success]`
- `[:oban, :failure]`
- `[:oban, :trip_circuit]`
- `[:oban, :open_circuit]`

That isn't ideal and for those existing events and we want to transition to these events _after_ this PR:

- `[:oban, :job, :start | :stop | :exception]` (with existing metadata)
- `[:oban, :circuit]` (with open/closed as metadata)

For _this_ PR we are looking to add some new events:

- `[:oban, :dispatch, :start | :stop | :exception]` (for producer dispatch)
- `[:oban, :deschedule, :start | :stop | :exception]` (for producer descheduling)
- `[:oban, :prune, :start | :stop | :exception]` (for job pruning)

You're proposing that _instead_ of those events, we introduce a single new event like:

- `[:oban, :query, :start | :stop | :exception]` (with dispatch, deschedule, prune as part of the metadata)

Did I summarize things correctly?

bryannaegele: Apologies. I thought the name was the name of the job. The _this_ PR ones are 👍 The `:circuit` ones look like they might fit the pattern better with `[:oban, :circuit, :trip | :open]`. Thoughts?

sorentwo: Using `[:oban, :circuit, :trip | :open]` looks very appropriate to me. 👍

sorentwo:
Note that I've pushed an change for job events so that they match the standard telemetry span format 82acf2b. A change for circuit breakers will come soon.

sorentwo:
Recently, as part of some shuffling for the next big release, I removed the `Pruner` module entirely and broke this PR. I've salvaged part of the `span/3` function and updated all of the telemetry events based on the feedback from @mcrumm and @bryannaegele.

Thanks for all your work pushing this forward @v0idpwn 💛

PR #211, opened 2020-04-19
Only create schema if prefix not "public"

pthompson:
- Patrick Thompson: Don't unnecessarily try to create a new schema if prefix is "public", as public is the default schema and doing so may cause an issue on databases that do not allow new schemas to be created.

### DIFF
--- a/lib/oban/migrations.ex
+++ b/lib/oban/migrations.ex
@@ -138,7 +138,8 @@ defmodule Oban.Migrations do
     import Oban.Migrations.Helper
 
     def up(prefix) do
-      execute "CREATE SCHEMA IF NOT EXISTS #{prefix}"
+      if prefix != "public", do:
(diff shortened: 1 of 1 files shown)


lib/oban/migrations.ex
```diff
@@ -138,7 +138,8 @@ defmodule Oban.Migrations do
     import Oban.Migrations.Helper
 
     def up(prefix) do
-      execute "CREATE SCHEMA IF NOT EXISTS #{prefix}"
+      if prefix != "public", do:
```
sorentwo, line 141: This upsets the formatter. Moving it onto a single line should do it.

pthompson: Sorry about that, just updated.

sorentwo:
Thanks @pthompson!

PR #214, opened 2020-05-01
Update example code to match breaking changes from changelog

bglusman:
The changelog makes it clear what happened here https://github.com/sorentwo/oban/blob/5e3bccf1d94008d88b53c6f60ae1e550529aea53/CHANGELOG.md#breaking-changes

but there were still numerous lingering references to the old telemetry events that confused me for a few minutes, so thought we should update!  Thanks for the great work, loving the library so far!

### DIFF
--- a/lib/oban/telemetry.ex
+++ b/lib/oban/telemetry.ex
@@ -20,7 +20,7 @@ defmodule Oban.Telemetry do
   | `:stop`      | `:duration`   | `:id, :args, :queue, :worker, :attempt, :max_attempts`                             |
   | `:exception` | `:duration`   | `:id, :args, :queue, :worker, :attempt, :max_attempts, :kind, :error, :stacktrace` |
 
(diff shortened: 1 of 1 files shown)


sorentwo:
Thanks for the update! 💛

bglusman:
Oh, you're welcome!  Was actually just about to come back and apologize as I guess these changes aren't in the latest hex release yet and wasn't sure how you were handling docs updates between releases, maybe I'll just switch off hex and use master for now so I have the new version though...  also I'd love to find opportunities to contribute to project beyond documentation if anything comes up that seems good for someone new to the codebase etc.  I'm hoping I'll convince work to subscribe to the oban.dev beta/UI subscription, but either way want to support your awesome work!

sorentwo:
It would be great to have add a comment to the README stating that it applies to master and not to the current 1.2 release. A PR is welcome for that!

PR #220, opened 2020-05-04
Add telemetry change API note to README

bglusman:


sorentwo:
I _think_ this is covered now. After our discussions Friday I added a note to the README https://github.com/sorentwo/oban/commit/cb8cc0b0c63d1e2b9c3fc94d3fbcafc147f2c2f7

bglusman:
Oh!  OK, didn't see...  was about to add a comment wondering if we follow semantic versioning/whether a signature change like this breaks that and would make next version have to be 2.0.0 or something, but, dunno, not sure what motivated the change so just thougth discussing backwards compatibility etc was relevant.  I'm sure it'll be in the changelog/maybe just a prominent note about existing integrations to telemetry events... I started off looking at these from this article https://wwwtech.de/articles/2020/jan/attaching-appsignal-to-oban and anyone else who followed that will need to update their integrations....

bglusman:
Oh I suppose we could also go the way (I think) Tesla did and emit the old events and the new events both? https://github.com/teamon/tesla/pull/324

sorentwo:
I hadn’t seen that article, interesting!

The Tesla approach is great. My plan is to release 2.0 with breaking changes all in one go.  I’m trying to be quick about it!

PR #222, opened 2020-05-06
Allow :infinity as a unique period

### LINKED ISSUE #210: Allow :infinity as a unique period

It is often desirable to treat a job like it is unique "forever", where "forever" means as long as there is historical data in the database.

Allow specifying `period: :infinity` for worker and job uniqueness. Behind the scenes that would translate to a ludicrously long period of time, e.g. 999 years.

The alternative is to specify a very long number of seconds, `period: 60 * 60 * 24 * 365 * 999`, which is rather verbose.

rowofpixels:
Resolves #210 

This isn't the exact proposed solution, if you want me to change it to translate the `:infinity` option to a large period I can change that. Instead I just alter the query to not limit to any date range if `:infinity` was passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [Oban.Telemetry] Add `span/3` for reporting normalized `:start`, `:stop` and
   `:exception` events with timing information.
 
+- [Oban.Query] Add `:infinity` option for unique period.
+
 ### Changed
 
 - [Oban.Notifier] Make the module public and clean up the primary function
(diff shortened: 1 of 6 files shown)


rowofpixels:
@sorentwo I believe I addressed your points, let me know if there's any wording / implementation changes you can think of. Thanks!

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [Oban.Telemetry] Add `span/3` for reporting normalized `:start`, `:stop` and
   `:exception` events with timing information.
 
+- [Oban.Query] Add `:infinity` option for unique period.
```
sorentwo, line 50: From a user's perspective this is exposed at the worker/job level:

```suggestion
- [Oban.Worker] Add `:infinity` option for unique period.
```

PR #223, opened 2020-05-06
Transition worker callback from perform/2 to perform/1

sorentwo:
As the project approaches a major version bump it is time to get in a few breaking changes. The goal of this PR is to unify the worker callbacks so that they all accept a job struct and to make testing cleaner.

### Rationale

Having spent a _lot_ more time writing and testing workers I kept running into the same situation. Most workers never use the `job` argument and I (or my team) would work around it to try and make the test clearer. We would repeatedly do one of the following:

* Give `job` a default of `nil`, e.g. `perform(args, job \\ nil)`
* Call `perform/2` with `nil` or an empty map explicitly

After doing that in dozens of places it seems wrong!

### Solution

Go *back* to a `perform/1` callback, but expect a job struct rather than an args map. That means a worker that looks like this:

```elixir
def perform(%{"id" => id}, _job)
```

Will now look like this:

```elixir
def perform(%{args: %{"id" => id}})
```

It is slightly noisier, but much clearer when invoked manually in tests. Compare this:

```elixir
perform(%{"id" => 1}, nil)
perform(%{"id" => 1}, %{})
perform(%{"id" => 1}, %Job{})
```

with this:

```elixir
perform(%{args: %{"id" => 1}})
```

This also enables a fix for a related issue, which is attempting to match on atom keys when they'll always be strings. A helper function like `Oban.Job.new/1,2` can ensure all keys are strings and be used to craft a real-world payload when testing. With `new/1,2` test calls could look like this:

```elixir
perform(Job.new(id: 1))
perform(Job.new(%{id: 1}, attempt: 42))
```

---

This flips the decision made in https://github.com/sorentwo/oban/pull/46

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,7 +318,7 @@ defmodule MyApp.Business do
   use Oban.Worker, queue: :events
 
   @impl Oban.Worker
-  def perform(%{"id" => id} = args, _job) do
+  def perform(%_{args: %{"id" => id}}) do
     model = MyApp.Repo.get(MyApp.Business.Man, id)
 
     case args do
@@ -350,7 +350,7 @@ defmodule MyApp.LazyBusiness do
     unique: [period: 30]
 
   @impl Oban.Worker
-  def perform(_args, _job) do
+  def perform(_job) do
     # do business slowly
 
     :ok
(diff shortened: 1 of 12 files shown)


sorentwo:
The test helper and corresponding documentation will be in a separate PR 💛

Issue #236, opened 2020-05-21
Support ranges + step in crontab


jc00ke:
### Is your feature request related to a problem? Please describe.

I'm porting over some functions executed by Quantum to Oban and I found out a range with a step isn't supported. Mine in particular is `0 15 2-30/2 * *`. 

> At 15:00 on every 2nd day-of-month from 2 through 30.

from [crontab.guru](https://crontab.guru/#0_15_2-30/2_*_*)

### Describe the Solution You'd Like

To have support for this syntax.

### Describe Alternatives You've Considered

For now I'll change the worker to skip the days it previously wouldn't have run on.

PR #238, opened 2020-05-23
feat: Step values can be used in conjunction with ranges

milmazz:
Fixes: #236


sorentwo:
@milmazz That was quick! Nicely done 💛

PR #239, opened 2020-05-23
chore: minor fixes

milmazz:
This PR introduces some minor changes:

* fix two typos in the README file
* use `is_reference/1` before `Process.cancel_timer/1`
* make `Oban.Job.coerce_field/1` private, it's only used internally
* fix combined concurrency level value in `Oban` docs
* introduce a hard-deprecation for `Oban.kill_job/2`, that way, the compiler will emit a warning during compilation.

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -64,7 +64,7 @@ defmodule Oban do
     documentation for more details.
   * `:queues` — a keyword list where the keys are queue names and the values are the concurrency
     setting. For example, setting queues to `[default: 10, exports: 5]` would start the queues
-    `default` and `exports` with a combined concurrency level of 20. The concurrency setting
+    `default` and `exports` with a combined concurrency level of 15. The concurrency setting
     specifies how many jobs _each queue_ will run concurrently.
 
     For testing purposes `:queues` may be set to `false` or `nil`, which effectively disables all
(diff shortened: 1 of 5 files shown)


lib/oban.ex
```diff
@@ -471,22 +471,9 @@ defmodule Oban do
...
-  Kill a long running job with an id of `1`:
-
-      Oban.kill_job(1)
-      :ok
-  """
-  @doc since: "0.2.0"
-  @doc deprecated: "Use Oban.cancel_job/1 instead"
-  @spec kill_job(name :: atom(), job_id :: pos_integer()) :: :ok
+  @doc false
+  @deprecated "Use Oban.cancel_job/1 instead"
```
milmazz, line 475: I just noticed [here](https://github.com/sorentwo/oban/commit/c0b84efd4bc32c36038f8653f9687dde888f5001) that you _soft-deprecated_ this function on Apr 9, so, let me know if it's too soon for a _hard-deprecation_.

sorentwo: There hasn't been a release since this was soft-deprecated. However, the next release will be a major version bump that will require people to make some changes in their app.

milmazz:
@sorentwo Replaced `kill_job` with `cancel_job` to avoid the compilation warning while testing.

sorentwo:
Thanks for the fixes!

PR #240, opened 2020-05-24
Initial structure for guides

milmazz:
Given that the current `Oban` module takes its `@moduledoc` from the `README`, I think that the "Troubleshooting" section does not belong there. So, that's why I took that out from the `README` and started a "Guides" section, which I think is more appropriate in this case, and it could provide the space for more guides, recipes, or How-tos in the future.

Part of the sidebar main page:

Screen Shot 2020-05-24 at 2 28 48 PM

Sidebar from Troubleshooting guide:

Screen Shot 2020-05-24 at 2 36 11 PM

Please let me know what do you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -563,6 +563,7 @@ wildcards, step values or ranges:
 * `0` — Literal, matches only itself (only 0)
 * `*/15` — Step, matches any value that is a multiple (0, 15, 30, 45)
 * `0-5` — Range, matches any value within the range (0, 1, 2, 3, 4, 5)
+* `0-23/2` - Step values can be used in conjunction with ranges (0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -563,6 +563,7 @@ wildcards, step values or ranges:
 * `0` — Literal, matches only itself (only 0)
 * `*/15` — Step, matches any value that is a multiple (0, 15, 30, 45)
 * `0-5` — Range, matches any value within the range (0, 1, 2, 3, 4, 5)
+* `0-23/2` - Step values can be used in conjunction with ranges (0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 22)
```
milmazz, line 566: This is just to offer an example of the feature added with https://github.com/sorentwo/oban/pull/238

sorentwo:
Troubleshooting was a great candidate for guides. Thanks for getting the ball rolling 💛

Issue #230, opened 2020-05-17
Include prefix in telemetry metadata for job events


sorentwo:
### Is your feature request related to a problem? Please describe.

There isn't any way to differentiate between telemetry events from different prefixes. For example, the official oban site runs oban using both a `public` and `private` prefix. The public version powers the UI demo and the private version handles "real" jobs. I'd like to be notified of `private` failures, but never be notified about `public` failures because they are synthetic.

### Describe the Solution You'd Like

Include the configured `prefix` in `[:oban, :job, _]` metadata.

### Describe Alternatives You've Considered

Match on the worker itself using a whitelist.

PR #242, opened 2020-05-25
feat: include prefix in telemetry metadata for job events

milmazz:
Fixes: #230

### DIFF
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -70,7 +70,7 @@ defmodule Oban.Queue.Executor do
     Breaker.with_retry(fn -> report_finished(exec).state end)
   end
 
-  defp record_started(%__MODULE__{} = exec) do
+  def record_started(%__MODULE__{} = exec) do
     :telemetry.execute([:oban, :job, :start], %{system_time: exec.start_time}, exec.meta)
 
     exec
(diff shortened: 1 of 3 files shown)


lib/oban/queue/executor.ex
```diff
@@ -70,8 +70,7 @@ defmodule Oban.Queue.Executor do
     Breaker.with_retry(fn -> report_finished(exec).state end)
   end
 
-  @spec record_started(t()) :: t()
-  def record_started(%__MODULE__{} = exec) do
+  defp record_started(%__MODULE__{} = exec) do
```
sorentwo, line 73: These are public purposefully so that they can be composed arbitrarily by other modules. Let's keep this `def` instead.

```suggestion
  def record_started(%__MODULE__{} = exec) do
```

lib/oban/telemetry.ex
```diff
@@ -38,7 +38,7 @@ defmodule Oban.Telemetry do
   | event                      | metadata                               |
   | -------------------------- | -------------------------------------- |
   | `[:oban, :circuit, :trip]` | `:error, :message, :name, :stacktrace` |
-  | `[:open, :circuit, :open]` | `:name`                                |
+  | `[:oban, :circuit, :open]` | `:name`                                |
```
sorentwo, line 41: Nice catch 👍

PR #243, opened 2020-05-25
chore: remove dialyzer CLI warning while running `mix ci`

milmazz:
The option `--halt-exit-status` is no longer valid for the dialyzer CLI, that's the default behavior since `1.0.0-rc.7`.

Screen Shot 2020-05-25 at 6 33 25 AM


sorentwo:
@milmazz Nicely spotted. I've updated `nimble_parsec` so there won't be any more of the unused variable warnings muddling things. 💛

Issue #233, opened 2020-05-19
Include queuing time in job telemetry


sorentwo:
### Is your feature request related to a problem? Please describe.

Queue time and total runtime info can be derived through various timestamps. The total runtime is reported with the `[:oban, :job, :stop]` event, but queuing time isn't.

### Describe the Solution You'd Like

Include queuing time with `[:oban, :job, :stop]` timing data. The queuing time is the difference between `scheduled_at` and `attempted_at`.

### Describe Alternatives You've Considered

Extracting details from the database directly.

PR #244, opened 2020-05-25
feat: Include queuing time in job telemetry

milmazz:
This PR adds `enqueue_time` as part of the event measurements for the telemetry event: `[:oban, :job, :stop]`

Fixes: #233


sorentwo:
Looks perfect! Nicely done 💛

PR #246, opened 2020-05-27
Remove unknown plugins option for test env

kianmeng:
Update the README to use the correct config option for test env


sorentwo:
The README is for the master branch where `plugins: false` is the correct option.

The details on HexDocs are correct for the latest published version 1.2.0: https://hexdocs.pm/oban/Oban.html

PR #249, opened 2020-05-27
fix dialyzer error

milmazz:
Current `master` is failing because of this small "issue"


sorentwo:
Weird, I thought I pushed that after I made the initial change. Oh well, thanks for the fix!

PR #255, opened 2020-05-30
chore: make assert_valid_worker private

milmazz:
The function `Oban.Testing.assert_valid_worker/1` seems like a helper for the
newly added `perform_job/3` and I think it should be private.


sorentwo:
Nice catch 👍

Issue #248, opened 2020-05-27
Support starting and stopping queues on a single node


sorentwo:
### Is your feature request related to a problem? Please describe.

Calling `Oban.start_queue/3` will start the queue on all connected nodes. That isn't always desirable because it prevents global queue limits.

### Describe the Solution You'd Like

Introduce `start_queue/4`, which takes a set of options. The options will include `local_only`, which will prevent broadcasting to other nodes.

Alternatively we could deprecate `start_queue/3` and introduce `start_queue/2` which accepts required `queue` and `limit` options.

Whichever interface we arrive at we need to support it for `stop_queue/2` as well.

### Describe Alternatives You've Considered

There isn't an alternative to this for dynamic queues.

### Additional Context

See #114 for more context.

taylordowns2000:
This is great. Opens up all sorts of doors for advanced queue configuration, including dialing up and down concurrency on individual nodes to match compute resources—depending on which boxes they're deployed on.

sorentwo:
Closing in favor of the PR (#256)

PR #256, opened 2020-05-31
Support starting & stopping queues on single node

milmazz:
Fixes: #248

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -378,37 +378,39 @@ defmodule Oban do
   and prefix. You can pass the option `local_only: true` if you prefer to start the queue only on
   the local node.
 
+  ## Options
+
+  * `queue` - specifies the queue name
+  * `limit` - set the concurrency limit
+  * `local_only` - specifies if the queue will be started only on the local node, default: `false`
+
   ## Examples
 
-  Start the `:priority` queue with a concurrency limit of 10 across the
-  connected nodes.
+  Start the `:priority` queue with a concurrency limit of 10 across the connected nodes.
 
-      Oban.start_queue(:priority, 10)
+      Oban.start_queue(queue: :priority, limit: 10)
       :ok
 
   Start the `:media` queue with a concurrency limit of 5 only on the local node.
 
-      Oban.start_queue(:media, 5, local_only: true)
+      Oban.start_queue(queue: :media, limit: 5, local_only: true)
       :ok
   """
-  @doc since: "0.12.0"
-  @spec start_queue(
-          name :: atom(),
-          queue :: queue_name(),
-          limit :: pos_integer(),
-          opts :: Keyword.t()
-        ) :: :ok
-  def start_queue(name \\ __MODULE__, queue, limit, opts \\ [])
-      when is_queue(queue) and is_limit(limit) do
-    local_only? = Keyword.get(opts, :local_only, false)
-    payload = %{action: :start, queue: queue, limit: limit}
+  @doc since: "2.0.0"
+  @spec start_queue(name :: atom(), opts :: Keyword.t()) :: :ok
+  def start_queue(name \\ __MODULE__, opts)
+      when is_atom(name) and is_list(opts) do
+    Enum.each(opts, &validate_queue_opt!/1)
 
-    if local_only? do
-      send_signal(name, payload)
+    queue = Keyword.fetch!(opts, :queue)
+    limit = Keyword.fetch!(opts, :limit)
+
+    if Keyword.get(opts, :local_only) do
+      Midwife.start_queue(name, queue, limit)
     else
       name
       |> config()
-      |> Notifier.notify(:signal, payload)
+      |> Notifier.notify(:signal, %{action: :start, queue: queue, limit: limit})
     end
   end
 
@@ -477,33 +479,39 @@ defmodule Oban do
   @doc """
   Shutdown a queue's supervision tree and stop running jobs for that queue
 
-  By default this action will occur across all the running nodes. Still, if you
-  prefer to stop the queue's supervision tree and stop running jobs for that
-  queue only on the local node, you can pass the option: `local_only: true`
+  By default this action will occur across all the running nodes. Still, if you prefer to stop the
+  queue's supervision tree and stop running jobs for that queue only on the local node, you can
+  pass the option: `local_only: true`
 
-  The shutdown process pauses the queue first and allows current jobs to exit
-  gracefully, provided they finish within the shutdown limit.
+  The shutdown process pauses the queue first and allows current jobs to exit gracefully, provided
+  they finish within the shutdown limit.
+
+  ## Options
+
+  * `queue` - specifies the queue name
+  * `local_only` - specifies if the queue will be stopped only on the local node, default: `false`
 
   ## Examples
 
       Oban.stop_queue(:default)
       :ok
 
-      Oban.stop_queue(:media, local_only: true)
+      Oban.stop_queue(queue: :media, local_only: true)
       :ok
   """
-  @doc since: "0.12.0"
-  @spec stop_queue(name :: atom(), queue :: queue_name(), opts :: Keyword.t()) :: :ok
-  def stop_queue(name \\ __MODULE__, queue, opts \\ []) when is_atom(name) and is_queue(queue) do
-    local_only? = Keyword.get(opts, :local_only, false)
-    payload = %{action: :stop, queue: queue}
+  @doc since: "2.0.0"
+  @spec stop_queue(name :: atom(), opts :: Keyword.t()) :: :ok
+  def stop_queue(name \\ __MODULE__, opts) when is_atom(name) and is_list(opts) do
+    Enum.each(opts, &validate_queue_opt!/1)
 
-    if local_only? do
-      send_signal(name, payload)
+    queue = Keyword.fetch!(opts, :queue)
+
+    if Keyword.get(opts, :local_only) do
+      Midwife.stop_queue(name, queue)
     else
       name
       |> config()
-      |> Notifier.notify(:signal, payload)
+      |> Notifier.notify(:signal, %{action: :stop, queue: queue})
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/oban.ex
```diff
@@ -505,4 +541,18 @@ defmodule Oban do
...
+
+  defp send_signal(name, payload) do
+    name
+    |> child_name("Midwife")
+    |> send({:notification, :signal, json_encode_decode(payload)})
+
+    :ok
+  end
+
+  defp json_encode_decode(map) do
```
milmazz, line 553: Introduced this helper function to share the same payload (map with atom keys) between the default mode (broadcasting the start/stop queue event to the connected nodes) and the local mode, the latter expects a map with string keys.

lib/oban.ex
```diff
@@ -372,21 +372,44 @@ defmodule Oban do
...
+          limit :: pos_integer(),
+          opts :: Keyword.t()
+        ) :: :ok
+  def start_queue(name \\ __MODULE__, queue, limit, opts \\ [])
+      when is_queue(queue) and is_limit(limit) do
+    local_only? = Keyword.get(opts, :local_only, false)
+    payload = %{action: :start, queue: queue, limit: limit}
+
+    if local_only? do
+      send_signal(name, payload)
```
milmazz, line 407: * [ ] Add some unit tests for this case
* [ ] Check if the default path is already tested

lib/oban.ex
```diff
@@ -452,23 +475,36 @@ defmodule Oban do
...
-    name
-    |> config()
-    |> Notifier.notify(:signal, %{action: :stop, queue: queue})
+  @spec stop_queue(name :: atom(), queue :: queue_name(), opts :: Keyword.t()) :: :ok
+  def stop_queue(name \\ __MODULE__, queue, opts \\ []) when is_atom(name) and is_queue(queue) do
+    local_only? = Keyword.get(opts, :local_only, false)
+    payload = %{action: :stop, queue: queue}
+
+    if local_only? do
+      send_signal(name, payload)
```
milmazz, line 502: * [ ] Add some unit tests for this case
* [ ] Check if the default path is already tested

lib/oban.ex
```diff
@@ -372,21 +372,44 @@ defmodule Oban do
   end
 
   @doc """
-  Start a new supervised queue across all connected nodes.
+  Start a new supervised queue
 
-  ## Example
+  By default the new supervised queue will be across the connected nodes. Still,
+  if you prefer to start the queue only on the local node, you can pass the
+  option: `local_only: true`.
```
sorentwo, line 379:
```suggestion
  By default this starts a new supervised queue across all nodes running Oban on the same database
  and prefix. You can pass the option `local_only: true` if you prefer to start the queue only on
  the local node.
```

lib/oban.ex
sorentwo: We can pass a map with string keys to both and avoid this entirely, i.e.:

```elixir
%{"action" => "stop", "queue" => to_string(queue)}
```

lib/oban.ex
```diff
@@ -505,4 +541,18 @@ defmodule Oban do
   end
 
   defp child_name(name, child), do: Module.concat(name, child)
+
+  defp send_signal(name, payload) do
+    name
+    |> child_name("Midwife")
+    |> send({:notification, :signal, json_encode_decode(payload)})
```
sorentwo, line 548: This will technically work, but I think adding an api function directly to the `Midwife` module would be clearer. Something like `start_queue/2` and `stop_queue/2`, which then can handle sending a message within the module.

lib/oban.ex
```diff
@@ -452,23 +475,36 @@ defmodule Oban do
...
-  def stop_queue(name \\ __MODULE__, queue) when is_atom(name) and is_queue(queue) do
-    name
-    |> config()
-    |> Notifier.notify(:signal, %{action: :stop, queue: queue})
+  @spec stop_queue(name :: atom(), queue :: queue_name(), opts :: Keyword.t()) :: :ok
+  def stop_queue(name \\ __MODULE__, queue, opts \\ []) when is_atom(name) and is_queue(queue) do
+    local_only? = Keyword.get(opts, :local_only, false)
+    payload = %{action: :stop, queue: queue}
+
+    if local_only? do
```
sorentwo, line 501: We can inline the keyword check to streamline these a little bit:

```suggestion
    if Keyword.get(opts, :local_only) do
```

lib/oban.ex
milmazz: That’s a great suggestion and I agree that it will more clear after putting the Client API functions in `Oban.Midwife`.

test/integration/controlling_test.exs
```diff
@@ -12,12 +12,30 @@ defmodule Oban.Integration.ControllingTest do
...
-    assert :ok = Oban.start_queue(:delta, 6)
-    assert :ok = Oban.start_queue(:alpha, 5)
+    assert :ok = Oban.start_queue(queue: :gamma, limit: 5)
+    assert :ok = Oban.start_queue(queue: :delta, limit: 6, local_only: true)
+    assert :ok = Oban.start_queue(queue: :alpha, limit: 5)
 
     assert_receive {:ok, 1}
     assert_receive {:ok, 2}
+
+    assert_raise ArgumentError,
```
sorentwo, line 22: Taken together these seem rather verbose. Perhaps a helper function and less complete matching could clean it up?

```elixir
assert_invalid_start([queue: nil], "exepcted :queue to be a binary")
assert_invalid_start([queue: :invalid, limit: -1], "expected :limit to be a positive integer")
```

They also seem like good candidates for a separate testing block.

milmazz: > Taken together these seem rather verbose

You're absolutely right, good catch.

> Perhaps a helper function and less complete matching could clean it up?

Yes, I agree, based on your suggestion I introduced `assert_invalid_stop/2` and `assert_invalid_start/2`.

milmazz: > They also seem like good candidates for a separate testing block.

Good idea!

Those helpers should be specific to `Oban.start_queue/2` and `Oban.stop_queue/2`? I mean, something like: `assert_invalid_start_queue_option/1,2` and `assert_invalid_stop_queue_option/1,2` or do you prefer something general to check invalid options? I ask because I'm seeing a lot of the same checks on `worker_test.exs` like:


```elixir
test/oban/worker_test.exs
96:    assert_raise ArgumentError, ~r/unknown option/, fn ->
106:    assert_raise ArgumentError, ~r/expected :queue to be/, fn ->
116:    assert_raise ArgumentError, ~r/expected :max_attempts to be/, fn ->
126:    assert_raise ArgumentError, ~r/expected :priority to be/, fn ->
136:    assert_raise ArgumentError, ~r/expected :tags to be a list/, fn ->
144:    assert_raise ArgumentError, ~r/expected :tags to be a list/, fn ->
154:    assert_raise ArgumentError, ~r/unexpected unique options/, fn ->
162:    assert_raise ArgumentError, ~r/unexpected unique options/, fn ->
170:    assert_raise ArgumentError, ~r/unexpected unique options/, fn ->
178:    assert_raise ArgumentError, ~r/unexpected unique options/, fn ->
186:    assert_raise ArgumentError, ~r/unexpected unique options/, fn ->
```

Or should we also introduce a helper to check the invalid options given to the worker?something like `assert_invalid_worker_option/1,2`, where the message is optional. If this is the case, I think I can introduce those helpers in another PR, to keep things focused. wdyt?

lib/oban.ex
```diff
@@ -372,21 +372,37 @@ defmodule Oban do
   end
 
   @doc """
-  Start a new supervised queue across all connected nodes.
+  Start a new supervised queue
```
sorentwo, line 375:
```suggestion
  Start a new supervised queue.
```

lib/oban.ex
```diff
@@ -372,21 +372,37 @@ defmodule Oban do
...
+      Oban.start_queue(queue: :media, limit: 5, local_only: true)
       :ok
   """
-  @doc since: "0.12.0"
-  @spec start_queue(name :: atom(), queue :: queue_name(), limit :: pos_integer()) :: :ok
-  def start_queue(name \\ __MODULE__, queue, limit) when is_queue(queue) and is_limit(limit) do
+  @doc since: "2.0.0"
+  @spec start_queue(name :: atom(), opts :: Keyword.t()) :: :ok
+  def start_queue(name \\ __MODULE__, opts)
+      when is_atom(name) and is_list(opts) do
```
sorentwo, line 402:
```suggestion
  def start_queue(name \\ __MODULE__, [_ | _] = opts) when is_atom(name) do
```

lib/oban.ex
```diff
@@ -452,23 +468,34 @@ defmodule Oban do
   end
 
   @doc """
-  Shutdown a queue's supervision tree and stop running jobs for that queue across all running
-  nodes.
+  Shutdown a queue's supervision tree and stop running jobs for that queue
```
sorentwo, line 471:
```suggestion
  Shutdown a queue's supervision tree and stop running jobs for that queue.
```

lib/oban/midwife.ex
```diff
@@ -21,6 +21,96 @@ defmodule Oban.Midwife do
     GenServer.start_link(__MODULE__, opts, name: name)
   end
 
+  @doc """
```
sorentwo, line 24: This shouldn't be documented publicly. The public interface for this is _always_ through `Oban.start_queue`.

lib/oban/midwife.ex
```diff
@@ -21,6 +21,96 @@ defmodule Oban.Midwife do
...
+
+  Start the `:media` queue with a concurrency limit of 5 only on the local node.
+
+      Oban.start_queue(config, queue: :media, limit: 5, local_only: true)
+      :ok
+  """
+  @doc since: "2.0.0"
+  @spec start_queue(Config.t(), opts :: Keyword.t()) :: :ok
+  def start_queue(%Config{name: name} = conf, opts)
+      when is_list(opts) do
```
sorentwo, line 52:
```suggestion
  def start_queue(%Config{name: name} = conf, [_ | _] = opts) do
```

lib/oban/midwife.ex
```diff
@@ -21,6 +21,96 @@ defmodule Oban.Midwife do
...
+      Oban.stop_queue(config, queue: :default)
+      :ok
+
+      Oban.stop_queue(config, queue: :media, local_only: true)
+      :ok
+  """
+
+  @doc since: "2.0.0"
+  @spec stop_queue(Config.t(), opts :: Keyword.t()) :: :ok
+  def stop_queue(%Config{name: name} = conf, opts) when is_list(opts) do
```
sorentwo, line 98:
```suggestion
  def stop_queue(%Config{name: name} = conf, [_ | _] = opts) do
```

sorentwo:
I cleaned up the midwife calls a little, added tests for the `local_only` flag and added a changelog entry locally. This is in master now. Thanks again @milmazz 💛

lib/oban/midwife.ex
milmazz: This `@doc` will not be processed by ex_doc given that the `@moduledoc` is `false`, I added this mainly as a reference for other developers.

sorentwo: Understood. I don't want to maintain the same docs and options in both places though. We could put in a `@doc see Oban.start_queue/2` or something (I think there is a syntax for doing that cleanly, but I don't recall it).

PR #257, opened 2020-06-01
Update examples to give correct type for backoff/1

al2o3cr:
`DateTime.add/4` only accepts `integer()`, and the contract for `backoff/1` requires it to return `pos_integer()`. 

Adds `trunc` (as used in `default_backoff`) to the examples to match this requirement.

Original discussion here: https://elixirforum.com/t/datetime-issue-with-elixir-oban/32015


sorentwo:
Thanks for covering the issue on the forum and for the docs love! 💛

pankaj-ag:
Thank you @al2o3cr for the quick response on this one.

Issue #258, opened 2020-06-02
Support Multi.insert with function instead of changeset


jakesgordon:
### Is your feature request related to a problem? Please describe.

I need to insert a job as part of an `Ecto.Multi` transaction whose parameters depend on preceding parts of that transaction (e.g. I need to pass the ID of my inserted rows to the background job). `Ecto.Multi.insert` supports this via accepting _either_ a `Changeset` or a function as the 3rd parameter (see https://hexdocs.pm/ecto/Ecto.Multi.html#insert/4). `Oban.insert` provides a wrapper for doing this with a `Changeset`, but not with a function.

### Describe the Solution You'd Like

Add an `Oban.insert/4` method that accepts a function to provide the `Changeset` (instead of the changeset itself) following same pattern as `Multi.insert`

### Describe Alternatives You've Considered

I worked around the issue by using `Multi.insert` instead of `Oban.insert` but I assume that means I miss out on some of the unique oban guarantees for that job?

### Additional Context

I've just started using Oban but it looks very well thought out, great work! Thank you.

jakesgordon:
Thought I better add an example client code... Example creating a new user, creating an email activation token then using a background job to send the email to the user I need the user id and activation token that was part of the same transaction...

```
Multi.new                                                                            
|> Multi.insert(:user,           User.create_changeset(user_params))            
|> Multi.insert(:activate_token, fn %{ user: user } -> UserToken.create(user, :activate) end)
|> Oban.insert(:activate_email,  fn %{ user: user, activate_token: token } -> Mailer.new(%{ user_id: user.id, token: token }) end)
|> Repo.transaction                                                                                                                                                                
```

sorentwo:
Thanks for the example. This is definitely a use case we should support.

I’ve labeled this as “help wanted” in case somebody wants to pick it up.

jakesgordon:
Thank you!

PR #260, opened 2020-06-04
feat: Oban.insert/4 support adding a function to run as part of the multi

milmazz:
Fixes: #258

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -203,7 +203,7 @@ defmodule Oban do
 
       Ecto.Multi.new()
       |> Oban.insert("job-1", MyApp.Worker.new(%{id: 1}))
-      |> Oban.insert("job-2", MyApp.Worker.new(%{id: 2}))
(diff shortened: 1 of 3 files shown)


lib/oban.ex
```diff
@@ -203,23 +203,32 @@ defmodule Oban do
...
+      |> Oban.insert("job-2", fn _ -> MyApp.Worker.new(%{id: 2}) end)
       |> MyApp.Repo.transaction()
   """
   @doc since: "0.7.0"
   @spec insert(
           name :: atom(),
           multi :: Multi.t(),
           multi_name :: Multi.name(),
-          changeset :: Changeset.t(Job.t())
+          changeset_or_fun :: Changeset.t(Job.t()) | fun()
```
sorentwo, line 214: I think we can parameterize the `fun`, though the spec for `Ecto.Multi.insert/4` looks wrong to me. The function receives a map of changes, not a changeset or schema.

PR #262, opened 2020-06-05
Clarify oban_config

jamilabreu:


README.md
```diff
@@ -250,7 +250,7 @@ defmodule MyApp.Application do
   defp oban_config do
     opts = Application.get_env(:my_app, Oban)
 
-    # Prevent running queues or scheduling jobs from an iex console.
+    # Prevent running queues or scheduling jobs from an iex console or when starting app with `iex -S mix`
```
sorentwo, line 253: These are effectively the same thing, I think we can change it to an example of using the iex console.

```suggestion
    # Prevent running queues or scheduling jobs from an iex console, i.e. when starting app with `iex -S mix`
```

sorentwo:
I applied my suggestion automatically. Thanks for improving the docs! 💛

PR #264, opened 2020-06-10
Updated readme for realse 2.0.0-rc.0

pankaj-ag:
Updated readme version from  `{:oban, "~> 1.2"}` to `{:oban, "~> 2.0.0-rc.0"}`.

As we have replaced the pruner with a plugin system the following plugin option in test config throws an error. 

```elixir
# config/test.exs
config :my_app, Oban, crontab: false, queues: false, plugins: false
```

So I changed the current version in the readme to keep the consistency. I hope that is worth it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ dependencies in `mix.exs`:
 ```elixir
 def deps do
   [
-    {:oban, "~> 1.2"}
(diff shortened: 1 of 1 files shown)


pankaj-ag:
I struggle a bit with this error while experimenting with the Oban.

Issue #90, opened 2019-11-06
Oban notifier eats up postgres connections


andreasknoepfle:
### Environment

- Oban Version: v0.10.0
- PostgreSQL Version 11.4
- Elixir & Erlang/OTP Versions 1.9.0 22.0.4

### Current Behavior

We observed in our running system that all instances were failing, since non of the instances could get postgres connections anymore. After debugging into the issue we did a quick activity check in postgres to see who owns all the connections: 

```
SELECT pid, query, state FROM pg_stat_activity;

    [11660, "LISTEN \"public.oban_signal\"", "idle"],
    [13727, "LISTEN \"public.oban_signal\"", "idle"],
    [13728, "LISTEN \"public.oban_signal\"", "idle"],
    [13729, "LISTEN \"public.oban_signal\"", "idle"],
    [11661, "LISTEN \"public.oban_signal\"", "idle"],
    [13731, "LISTEN \"public.oban_signal\"", "idle"],
    [13732, "LISTEN \"public.oban_signal\"", "idle"],
    [11662, "LISTEN \"public.oban_signal\"", "idle"],
    [13734, "LISTEN \"public.oban_signal\"", "idle"],
    [13735, "LISTEN \"public.oban_signal\"", "idle"],
    [11663, "LISTEN \"public.oban_signal\"", "idle"],
    [13737, "LISTEN \"public.oban_signal\"", "idle"],
    [11664, "LISTEN \"public.oban_signal\"", "idle"],
    [13740, "LISTEN \"public.oban_signal\"", "idle"],
    [13738, "LISTEN \"public.oban_signal\"", "idle"],
    [13741, "LISTEN \"public.oban_signal\"", "idle"],
    [13742, "LISTEN \"public.oban_signal\"", "idle"],
    [11665, "LISTEN \"public.oban_signal\"", "idle"],
    [11666, "LISTEN \"public.oban_signal\"", "idle"],
    [13744, "LISTEN \"public.oban_signal\"", "idle"],
    [13746, "LISTEN \"public.oban_signal\"", "idle"],
    [11667, "LISTEN \"public.oban_signal\"", "idle"],
    [13748, "LISTEN \"public.oban_signal\"", "idle"],
    [13749, "LISTEN \"public.oban_signal\"", "idle"],
    [11668, "LISTEN \"public.oban_signal\"", "idle"],
    [13751, "LISTEN \"public.oban_signal\"", "idle"],
[...]

PR #265, opened 2020-06-12
Fixed connection bomb from issue #90

bimboleta:
Hey there,

Today I faced the same issue from https://github.com/sorentwo/oban/issues/90.

The notifier has a Breaker circuit which will restart the postgres connection on an interval if it crashes for any reason. However, it's triggering the trip_circuit schedule twice. It calls it once because of the error result from Notifications.start_link and once because of the exit trap at the start of the GenServer. This exit is called, not so obviously, because the Notifications.start_link is starting a linked Connection process which emits the signal when the connection fails. The erlang documentation points this out: _"Terminating processes emit exit signals to all linked processes"_

image

image

It's quite simple to see it happening. I added some logs to Notifiers functions and it's quite clear. Here, the same notifier PID is already receiving two resets from the breaker, followed by two exits.
image

The reset routine is doubled each time, similar to a fork bomb. If this happens for a while (db maintenance, for instance), my terminal turns to flood, with high cpu usage and, once the db server is up, thousands of connections wear the postgres out.

image

### DIFF
--- a/lib/oban/breaker.ex
+++ b/lib/oban/breaker.ex
@@ -20,7 +20,11 @@ defmodule Oban.Breaker do
   defmacro trip_errors, do: [DBConnection.ConnectionError, Postgrex.Error]
 
   @spec trip_circuit(Exception.t(), list(), state_struct()) :: state_struct()
-  def trip_circuit(exception, stack, %{circuit: _, conf: conf, name: name, reset_timer: reset_timer} = state) do
+  def trip_circuit(
+        exception,
+        stack,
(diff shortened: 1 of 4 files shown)


bimboleta:
Thank you for your quick response :)

I'm a little confused about this because I see no way of the Postgrex.Notifications initing without a {:ok, pid} or {:stop, err} with its termination (followed by the exit signal).
However, the timer solution works and is solid enough if, by any reason, that ever happens.

I've already fixed the PR.

sorentwo:
Merged manually for a few small fixes and the addition of a changelog entry. Thanks for this 💛

Issue #212, opened 2020-04-23
Support crobtan @reboot option


ludwikbukowski:
### Is your feature request related to a problem? Please describe.

It is possible to schedule cron jobs that are run periodically.
I'd like to have mechanism of running a job for each application startup, only once each time, automatically.
Similarly to [@reboot](https://en.wikipedia.org/wiki/Cron#Overview) option in crontab.

### Describe the Solution You'd Like

We could extend crontab option with reboot, e.g:
```elixir
crontab: [
  {:reboot, MyApp.RebootWorker}
]
```

### Describe Alternatives You've Considered

Currently, in order to run worker on startup, we would have to wait until oban supervision tree is setup and then, somehow insert job. It is not obvious how could we achieve that. It would be nice to have more standardised way.

bglusman:
@sorentwo just looking around at open issues, does this seem like a good ticket I could try and tackle for you, if I have time this weekend perhaps?  Or open to other suggestions, but this makes sense to me and seems (without having read related code yet) probably simple enough.

sorentwo:
@bglusman I'm reworking what is included in the base oban package, so this isn't an optimal thing to contribute to currently. This would be a better first issue https://github.com/sorentwo/oban/issues/210, if you'd like to give that a shot 👍

sorentwo:
Closing in favor of the PR (#268)

ludwikbukowski:
Thank you @sorentwo !

PR #268, opened 2020-06-15
Add reboot handling to crontab

TylerWitt:
Addresses #212 

I wasn't sure if the locking from `lock_and_enqueue` was required or not, and I added a test specifically around running once a node and it seems to be passing.

The reboot option still follows the one minute unique clause.

If either of these are undesirable, or if I should add locking under the same or a slightly different key, let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -580,6 +580,7 @@ The following Cron extensions are supported:
 * `@weekly` — `0 0 * * 0`
 * `@monthly` — `0 0 1 * *`
 * `@yearly` (as well as `@annually`) — `0 0 1 1 *`
+* `@reboot` — Run once at boot across the entire cluster
 
 Some specific examples that demonstrate the full range of expressions:
 
--- a/lib/oban/crontab/cron.ex
+++ b/lib/oban/crontab/cron.ex
@@ -10,7 +10,8 @@ defmodule Oban.Crontab.Cron do
           hours: expression(),
           days: expression(),
           months: expression(),
-          weekdays: expression()
+          weekdays: expression(),
+          reboot: boolean()
         }
 
   @part_ranges %{
@@ -21,12 +22,17 @@ defmodule Oban.Crontab.Cron do
     weekdays: {0, 6}
   }
 
-  defstruct minutes: [:*], hours: [:*], days: [:*], months: [:*], weekdays: [:*]
+  defstruct minutes: [:*], hours: [:*], days: [:*], months: [:*], weekdays: [:*], reboot: false
 
   @spec now?(cron :: t(), datetime :: DateTime.t()) :: boolean()
-  def now?(%__MODULE__{} = cron, datetime \\ DateTime.utc_now()) do
+  def now?(cron, datetime \\ DateTime.utc_now())
+
+  def now?(%__MODULE__{reboot: true}, _datetime), do: false
+
+  def now?(%__MODULE__{} = cron, datetime) do
     cron
     |> Map.from_struct()
+    |> Map.drop([:reboot])
     |> Enum.all?(fn {part, values} ->
       Enum.any?(values, &matches_rule?(part, &1, datetime))
     end)
@@ -62,6 +68,7 @@ defmodule Oban.Crontab.Cron do
   - @daily: Run once a day, "0 0 * * *".
   - @midnight: same as @daily
   - @hourly: Run once an hour, "0 * * * *".
+  - @reboot: Run once at boot
 
   ## Examples
 
@@ -82,6 +89,7 @@ defmodule Oban.Crontab.Cron do
   def parse!("@midnight"), do: parse!("@daily")
   def parse!("@daily"), do: parse!("0 0 * * *")
   def parse!("@hourly"), do: parse!("0 * * * *")
+  def parse!("@reboot"), do: struct!(__MODULE__, reboot: true)
 
   def parse!(input) when is_binary(input) do
     input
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -580,6 +580,7 @@ The following Cron extensions are supported:
 * `@weekly` — `0 0 * * 0`
 * `@monthly` — `0 0 1 * *`
 * `@yearly` (as well as `@annually`) — `0 0 1 1 *`
+* `@reboot`
```
sorentwo, line 583: It doesn't match the format, but I think a description is important here.

```suggestion
* `@reboot — Run once at boot across the entire cluster`
```

lib/oban/crontab/cron.ex
```diff
@@ -83,6 +90,10 @@ defmodule Oban.Crontab.Cron do
   def parse!("@daily"), do: parse!("0 0 * * *")
   def parse!("@hourly"), do: parse!("0 * * * *")
 
+  def parse!("@reboot") do
+    struct!(__MODULE__, reboot: true)
+  end
```
sorentwo, line 95: This fits nicely inline with the other irregular expressions:

```suggestion
  def parse!("@reboot"), do: struct!(__MODULE__, reboot: true)
```

lib/oban/crontab/cron.ex
```diff
@@ -129,4 +140,8 @@ defmodule Oban.Crontab.Cron do
   defp expand({_type, value}, min, max) do
     raise ArgumentError, "Unexpected value #{inspect(value)} outside of range #{min}..#{max}"
   end
+
+  @spec reboot?(cron :: t()) :: boolean()
+  def reboot?(%__MODULE__{reboot: true}), do: true
+  def reboot?(_cron), do: false
```
sorentwo, line 146:
```suggestion
  def reboot?(%__MODULE__{reboot: reboot}), do: reboot
```

lib/oban/crontab/scheduler.ex
```diff
@@ -51,6 +51,7 @@ defmodule Oban.Crontab.Scheduler do
 
   @impl GenServer
   def handle_continue(:start, state) do
+    enqueue_reboot(state.conf)
```
sorentwo, line 54: Let's call these `insert` instead of `enqueue`, for consistency with the query it's calling.

```suggestion
    insert_reboot(state.conf)
```

test/integration/crontab_test.exs
```diff
@@ -96,6 +96,36 @@ defmodule Oban.Integration.CrontabTest do
     refute_receive {:ok, 2}
   end
 
+  test "reboot jobs are enqueued on startup" do
```
sorentwo, line 99: Looking at these tests now I realize that we don't need to execute the job to make assertions at all. Using `queues: false` and asserting that the job is in the database is even better.

test/integration/crontab_test.exs
```diff
@@ -96,6 +96,36 @@ defmodule Oban.Integration.CrontabTest do
...
+
+    start_supervised_oban!(base_opts ++ [name: ObanA])
+    start_supervised_oban!(base_opts ++ [name: ObanB])
+
+    assert_receive {:ok, 1}
+
+    :ok = stop_supervised(ObanA)
+    :ok = stop_supervised(ObanB)
+
+    assert 1 == Job |> select(count()) |> Repo.one()
```
sorentwo, line 126: Same here; we don't need to execute the job to verify this behavior.

test/oban/crontab/crontab_test.exs
```diff
@@ -118,6 +120,45 @@ defmodule Oban.Crontab.CronTest do
...
+      crontab = %Cron{
+        minutes: [:*],
+        hours: [:*],
+        days: [:*],
+        months: [:*],
+        weekdays: [:*],
+        reboot: true
+      }
+
+      refute Cron.now?(crontab)
```
sorentwo, line 134: Wildcard values are the default, you only need to specify the `reboot` part:

```suggestion
      refute Cron.now?(%Cron{reboot: true})
```

test/oban/crontab/crontab_test.exs
```diff
@@ -118,6 +120,45 @@ defmodule Oban.Crontab.CronTest do
...
+      crontab = %Cron{
+        minutes: [:*],
+        hours: [:*],
+        days: [:*],
+        months: [:*],
+        weekdays: [:*],
+        reboot: false
+      }
+
+      refute Cron.reboot?(crontab)
```
sorentwo, line 160: Likewise here, only the `reboot` part is necessary:

```suggestion
      assert Cron.reboot?(%Cron{reboot: true})
      refute Cron.reboot?(%Cron{reboot: false})
```

lib/oban/crontab/scheduler.ex
```diff
@@ -51,6 +51,7 @@ defmodule Oban.Crontab.Scheduler do
 
   @impl GenServer
   def handle_continue(:start, state) do
+    insert_reboot(state.conf)
```
sorentwo, line 54: Reading back through the initial conversation I see that you were unsure whether this needed to be inside a `lock_and_enquque` block. It should, otherwise there's a chance of inserting duplicate jobs. Additionally, for consistency it would be nice to use `insert` consistently, i.e. `lock_and_insert`, `insert_jobs`.

TylerWitt: Should it use the same key as well, since one node may be inserting reboot jobs, and another node may be enqueuing scheduled ones? Or is it fine to assume that the node handling reboot jobs will also hit the scheduled ones if the other nodes are locked out due to reboot?

sorentwo: It is fine to assume that the node handling reboot jobs will get the scheduled ones as well. Ideally we'd run both forms of `insert` from a single lock transaction.

sorentwo:
Closed with e8bde2e

PR #269, opened 2020-06-16
put/get config via persistent term

milmazz:
It seems that `Oban.Config` is a good candidate for `persistent_term`, given that is only established at the beginning while setting up the supervision tree, and then is read it all over the place.

One benefit of this change is the performance improvement for reading operations. I made a quick benchmark using `Benchee` and here are the results:

```console
$ mix run bench/benchmarks/config_bench.exs
Operating System: macOS
CPU Information: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
Number of Available Cores: 12
Available memory: 16 GB
Elixir 1.10.2
Erlang 22.3.4

Benchmark suite executing with the following configuration:
warmup: 2 s
time: 10 s
memory time: 2 s
parallel: 1
inputs: none specified
Estimated total run time: 28 s

Benchmarking via_agent...
Benchmarking via_persistent_term...

Name                          ips        average  deviation         median         99th %
via_persistent_term        7.64 M       0.131 μs   ±294.53%           0 μs           1 μs
via_agent                  0.49 M        2.05 μs  ±1307.05%           2 μs           4 μs

Comparison:
via_persistent_term        7.64 M
via_agent                  0.49 M - 15.69x slower +1.92 μs

Memory usage statistics:

Name                        average  deviation         median         99th %
via_persistent_term             0 B     ±0.00%            0 B            0 B
via_agent                  336.00 B     ±0.16%          336 B          336 B

Comparison:
via_persistent_term             0 B
via_agent                  336.00 B - ∞ x memory usage +336.00 B
```

One caveat of this change is that `persistent_term` is available since Erlang/OTP 21.2. But, from the CI configuration, I can infer that the minimum Erlang/OTP version supported is 22.2. So, based on that, we should be good.

### DIFF
--- a/bench/benchmarks/config_bench.exs
+++ b/bench/benchmarks/config_bench.exs
@@ -1,0 +1,29 @@
+defmodule ConfigViaAgent do
+  use Agent
+
+  def start_link(opts) when is_list(opts) do
+    {conf, opts} = Keyword.pop(opts, :conf)
+    Agent.start_link(fn -> conf end, opts)
+  end
+
+  def get(name), do: Agent.get(name, & &1)
+end
+
+defmodule ConfigViaPersistentTerm do
+  def put(name, conf), do: :persistent_term.put(name, conf)
+  def get(name), do: :persistent_term.get(name)
+end
+
+defmodule Repo do
+  def __adapter__, do: :ok
+end
+
+name = Bench
+conf = Oban.Config.new(repo: Repo)
+ConfigViaAgent.start_link(conf: conf, name: name)
+ConfigViaPersistentTerm.put(name, conf)
+
+Benchee.run(%{
+  "via_agent" => fn -> ConfigViaAgent.get(name) end,
+  "via_persistent_term" => fn -> ConfigViaPersistentTerm.get(name) end
+})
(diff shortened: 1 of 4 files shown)


milmazz:
> Where it _does_ make a big difference is in preventing errors during shutdown. A persistent_term config will still be present regardless of the app's shutdown order, where an agent process may get stopped sooner than processes that make calls to `Oban`.

I don't recall seeing any issues related to that, but you're right, that could be a possible scenario, even in the current case where the Config process should be the last in the supervision tree to be stopped. And, it's also true that the persistent term will be present regardless of the app's shutdown, which is great on the previous scenario.

> The only blocker is that to make this change I believe we need to bump the minimum Elixir version to 1.9. I'm a little hesitant to do that as part of the 2.0 release itself and would like to get a little more feedback.

Yes, that's why I mentioned that we need at least Erlang/OTP 21.2. AFAIK Elixir 1.9 supports OTP 20-22, and Elixir 1.10 supports OTP 21-23.

I agree that we should wait for more feedback. This proposal is part of my adventures diving into the Oban codebase, so, no worries about putting this PR on hold or completely discard the idea.

sorentwo:
@milmazz Let's hold off on this until Elixir 1.11 is released later this year, around October. At that point I think we can bump the minimum version from 1.8 to 1.9. Before that it feels too aggressive, considering the performance increase we're looking at.

PR #272, opened 2020-06-17
Update FUNDING.yml to reflect new domain

zph:
I'm reading up on Oban and it reminds me of a custom internal tool we built in my last team. Except Oban looks far more polished :).

Thank you for building the library and when I come across a current need for this approach in Elixir+PG, I'll reach for Oban and Oban Pro.


sorentwo:
Nicely spotted. Thanks for this 💛

PR #273, opened 2020-06-17
Configurable :age for FixedPruner

balexand:
This makes the age configurable in `FixedPruner`. I realize that Oban Pro will include a `DynamicPruner`. I'm a paying Oban Pro customer and I'm looking forward to using `DynamicPruner`. I understand that in order to differentiate Oban Pro from the open source version, you'll need to leave certain features out of the open source version.

I believe this feature belongs in open source Oban because it was a feature of 1.x that was removed. It seems that the goal of Oban Pro is to develop sophisticated new functionality, not restore features that were taken away in 2.0.

Thanks!

### DIFF
--- a/lib/oban/plugins/fixed_pruner.ex
+++ b/lib/oban/plugins/fixed_pruner.ex
@@ -15,7 +15,7 @@ defmodule Oban.Plugins.FixedPruner do
   defmodule State do
     @moduledoc false
 
-    defstruct [:conf, :interval, :name]
+    defstruct [:age, :conf, :interval, :name]
   end
 
   @spec start_link(Keyword.t()) :: GenServer.on_start()
@@ -24,6 +24,7 @@ defmodule Oban.Plugins.FixedPruner do
 
     opts =
       opts
+      |> Keyword.put_new(:age, @prune_age)
       |> Keyword.put_new(:interval, @prune_interval)
       |> Map.new()
 
@@ -42,7 +43,7 @@ defmodule Oban.Plugins.FixedPruner do
     %Config{repo: repo, log: log} = conf
 
     Telemetry.span(:prune, fn ->
-      repo.transaction(fn -> acquire_and_prune(conf) end, log: log)
+      repo.transaction(fn -> acquire_and_prune(conf, state.age) end, log: log)
     end)
 
     Process.send_after(self(), :prune, state.interval)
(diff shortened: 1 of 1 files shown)


sorentwo:
I appreciate the thought and consideration you put into this PR.

>  It seems that the goal of Oban Pro is to develop sophisticated new functionality, not restore features that were taken away in 2.0.

The _primary_ goal of Oban Pro is to develop sophisticated new functionality. A secondary goal is to make Pro a desirable and convenient upgrade, and part of that is limiting the base functionality—even if it was available in an earlier version. Is there something specific you're hoping to achieve by restoring that functionality, or is it down to principal?

balexand:
I have jobs that need to be unique with a period greater than 60 seconds. If the pruner deletes the jobs after 60 seconds then this won't work. Also, I like to keep the prune age higher so I can view the completed jobs for debugging purposes.

Since my employer is an Oban Pro customer, I can just use `DynamicPruner` when it is released and stable. But it would be nice to have the option to use the basic `FixedPruner` with unique jobs.

Either way it's not a big deal because of the new plugin system, which is a great design. If I ever need to use unique jobs in a project where Oban Pro is not available then I can just write my own `FixedPruner`. I find Oban Web to be invaluable in my work and I'll pay for it on any non-trivial project. But I wouldn't pay for Oban Pro just to be able to configure a single `:age` field.

sorentwo:
> I have jobs that need to be unique with a period greater than 60 seconds. If the pruner deletes the jobs after 60 seconds then this won't work. Also, I like to keep the prune age higher so I can view the completed jobs for debugging purposes.

That's entirely reasonable, and essentially the reason that Oban retains jobs in the first place 🙂 

> But it would be nice to have the option to use the basic `FixedPruner` with unique jobs.

That's still possible, albeit in a slightly reduced capacity. In fact, without historic "completed" jobs to check against it has the same limited unique semantics as Sidekiq does.

> Either way it's not a big deal because of the new plugin system, which is a great design. If I ever need to use unique jobs in a project where Oban Pro is not available then I can just write my own `FixedPruner`.

That's the idea. If people want to manage retention themselves they can write a plugin or disable pruning entirely and do it manually. However, if people want to get something really customizable and proven then they can go for the plugins in Oban Pro.

Thanks for sharing your thought process. For now I'm going to close this and keep the `FixedPruner` as it is.

sorentwo:
I've reversed my position on this with 5eb6abf. The `FixedPruner` plugin is renamed to `Pruner` and it accepts a `max_age` value. Customizing the value is also documented in the README.

PR #276, opened 2020-06-23
Update minimum ecto_sql requirement

gabrielpra1:
Oban uses the syntax 

```elixir
|> where([j], j.id in subquery(subset))
```

in `Oban.Query.fetch_available_jobs/4`, but this is only supported from [ecto_sql version 3.4.3](https://github.com/elixir-ecto/ecto_sql/blob/master/CHANGELOG.md#v343-2020-04-27).

When trying to install it as a dependency in a project running a version prior to 3.4.3, oban fails to compile:

```
➜ mix deps.compile oban
==> oban
Compiling 20 files (.ex)

== Compilation error in file lib/oban/query.ex ==
** (Ecto.Query.CompileError) `subquery(^subset)` is not a valid query expression.

* If you intended to call a database function, please check the documentation
  for Ecto.Query to see the supported database expressions

* If you intended to call an Elixir function or introduce a value,
  you need to explicitly interpolate it with ^

    expanding macro: Ecto.Query.where/3
    lib/oban/query.ex:31: Oban.Query.fetch_available_jobs/4
    expanding macro: Ecto.Query.select/3
    lib/oban/query.ex:32: Oban.Query.fetch_available_jobs/4
    (elixir 1.10.2) expanding macro: Kernel.|>/2
    lib/oban/query.ex:34: Oban.Query.fetch_available_jobs/4
could not compile dependency :oban, "mix compile" failed. You can recompile this dependency with "mix deps.compile oban", update it with "mix deps.update oban" or clean it with "mix deps.clean oban"
```

By bumping the minimum required version, the error would be clearer, saying that ecto needs to be updated


sorentwo:
This will definitely help some people, thanks! 💛

PR #279, opened 2020-06-24
fix readme, :stack -> :stacktrace change in 2.0.0-rc.0

taylordowns2000:
Sorry, should have caught this with the last PR! The error reporting example needs the new `:stacktrace` key in `meta`.


sorentwo:
Don't be sorry, I should have caught that.

PR #280, opened 2020-06-24
fix: handle cases where the given cron step is one

milmazz:
The cron configuration presented on the `#oban` slack channel was the following:

```elixir
{"0 0 */1 * *", Worker.MarkerScheduler}
```

That configuration produced an `ArgumentError` exception while Oban tried to expand the parsed result by `nimble_parsec` saying that we got an `Unexpected value 1 outside of range 1..31`.

In this case, given that the step is 1, one quick workaround is replacing the original configuration with:

```elixir
{"0 0 * * *", Worker.MarkerScheduler}
```

And it will produce the same result, which is to run the specified job at 00:00 on every day-of-month.

But, given that the original configuration is valid, we need to support that as well. So, here is my attempt to fix that issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- [Oban.Crontab.Cron] Do not raise an `ArgumentError` exception when the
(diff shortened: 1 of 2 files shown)


milmazz:
@sorentwo Done

sorentwo:
@milmazz Merged manually. I made some modifications to this, including overhauling the property tests to cover this case. (It still has you as the author 😉)

Issue #281, opened 2020-06-26
Public func to convert module to worker string


chulkilee:
### Goal

I need to query jobs for specific worker module to cancel the job. To do that, I need to pass worker name as string.

e.g. `MyApp.Worker.HardWork` -> `"MyApp.Worker.HardWork"`.

### Describe the Solution You'd Like

Expose a function to convert module to string value can be used for `worker` column

Pick a good name and expose following private function

https://github.com/sorentwo/oban/blob/v2.0.0-rc.2/lib/oban/job.ex#L297-L301

chulkilee:
I can create PR for this - any thought on the naming?

BTW another workaround to get the value without copying the code is:

```elixir
MyWorker.new(%{}).changes.worker
```

sorentwo:
After experimenting a little I believe the best way to handle this is actually with `inspect/1`, because it does it automatically:

```
iex(2)> to_string(MyApp.MyModule)
"Elixir.MyApp.MyModule"
iex(3)> inspect(MyApp.MyModule)
"MyApp.MyModule"
```

I think the best option here is to document querying the jobs table with this module to string coercion. WDYT?

chulkilee:
For implementation - maybe inspect is better.

However I still think it's better to have func on oban for explicitness, instead of mentioning it only at doc.

Also I'd like to avoid adding oban internal specific code in my code if possible.

sorentwo:
The fact that `Atom.to_string/1`, and thereby `to_string/1` prefixes module names with `Elixir.` is more Elixir specific than Oban specific. Ideally I'd like to see a `Module.to_string/1` implementation that did this naturally rather than having to expose a helper from Oban itself.

In lieu of a `Module.to_string/1` function there may be value in a `Oban.Worker.to_string/1`, but it feels extraneous to me. The difference would look like this:

```elixir
# Using inspect
from j in Oban.Job, where: j.worker == ^inspect(MyApp.MyWorker)

# Using Worker.to_string
from j in Oban.Job, where: j.worker == ^Oban.Worker.to_string(MyApp.MyWorker)
```

chulkilee:
It's more verbose, but more clear and explicit to use oban's helper function instead of using the underlying assumption from documentation.

Also please note that it is technically possible to overwrite the protocol implementation, so `Kernel.inspect/2` is not stable. Thus using `Atom.to_string/1` and transform it for Oban is the best way.

```elixir
defmodule Foo do
end

inspect(Foo)
[...]

PR #284, opened 2020-07-02
Introduce Worker.to_string/1

chulkilee:
### Goal

See #281

### Changes

- Introduce `Oban.Worker.to_string/1` with docs and tests
- Replace `String/trim_leading` with binary pattern match

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -161,8 +161,8 @@ defmodule Oban.Job do
       opts
       |> Keyword.put(:args, args)
       |> Map.new()
-      |> coerce_field(:queue)
-      |> coerce_field(:worker)
+      |> coerce_field(:queue, &to_clean_string/1)
+      |> coerce_field(:worker, &Oban.Worker.to_string/1)
       |> normalize_tags()
 
     %__MODULE__{}
@@ -215,13 +215,13 @@ defmodule Oban.Job do
     |> Map.new()
   end
 
-  defp coerce_field(params, field) do
+  defp coerce_field(params, field, fun) do
     case Map.get(params, field) do
       value when is_atom(value) and not is_nil(value) ->
-        update_in(params, [field], &to_clean_string/1)
+        update_in(params, [field], fun)
 
       value when is_binary(value) ->
-        update_in(params, [field], &to_clean_string/1)
+        update_in(params, [field], fun)
 
       _ ->
         params
@@ -314,8 +314,9 @@ defmodule Oban.Job do
   end
 
   defp to_clean_string(value) do
-    value
-    |> to_string()
-    |> String.trim_leading("Elixir.")
+    case to_string(value) do
+      "Elixir." <> val -> val
+      val -> val
+    end
   end
 end
--- a/lib/oban/worker.ex
+++ b/lib/oban/worker.ex
@@ -188,6 +188,8 @@ defmodule Oban.Worker do
   """
   @moduledoc since: "0.1.0"
 
+  import Kernel, except: [to_string: 1]
+
   alias Oban.Job
 
   @type t :: module()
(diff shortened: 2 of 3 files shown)


lib/oban/job.ex
```diff
@@ -295,8 +295,9 @@ defmodule Oban.Job do
   end
 
   defp to_clean_string(value) do
-    value
-    |> to_string()
-    |> String.trim_leading("Elixir.")
+    case value |> to_string() do
```
sorentwo, line 298:
```suggestion
    case to_string(value) do
```

chulkilee: I found this single step pipe in other files as well.

Do you want to drop such single step pipe? Then I'll create another MR for that using [`Credo.Check.Readability.SinglePipe`](https://github.com/rrrene/credo/blob/v1.4.0/lib/credo/check/readability/single_pipe.ex)

sorentwo: Nicely spotted. I feel like there was an issue with that in nimble_parsec in the past, but it isn't an ussue now. I've fixed it with 0014bdf

sorentwo:
@chulkilee Bump. Are you still working on this?

chulkilee:
@sorentwo oh it's ready - not sure why CI failed. Rebased the branch onto the latest master to trigger new CI build.

sorentwo:
Oh, I see. There are a couple of TODOs above about adding docs and tests. A couple of doctests would be a great fit for something like this.

lib/oban/job.ex
```diff
@@ -215,13 +215,13 @@ defmodule Oban.Job do
     |> Map.new()
   end
 
-  defp coerce_field(params, field) do
+  defp coerce_field(params, field, func) do
```
sorentwo, line 218: For consistency with other anonymous function passing let's use `fun` instead of `func` here:

```suggestion
  defp coerce_field(params, field, fun) do
```

lib/oban/worker.ex
```diff
@@ -336,4 +338,12 @@ defmodule Oban.Worker do
   defp validate_opt!(option) do
     raise ArgumentError, "unknown option provided #{inspect(option)}"
   end
+
+  @spec to_string(module() | String.t()) :: String.t()
+  def to_string(worker) when is_atom(worker) and not is_nil(worker),
+    do: worker |> Kernel.to_string() |> to_string()
```
sorentwo, line 344: Again, for consistency let's use a `do/end` block, there aren't any multi-line `do:` functions elsewhere in the code:

```suggestion
  def to_string(worker) when is_atom(worker) and not is_nil(worker) do
    worker
    |> Kernel.to_string()
    |> to_string()
  end
```

chulkilee:
@sorentwo oh, forgot to add docs/tests. Added them - and made changes for your comments (marked them as resolved).

lib/oban/worker.ex
```diff
@@ -336,4 +338,18 @@ defmodule Oban.Worker do
   defp validate_opt!(option) do
     raise ArgumentError, "unknown option provided #{inspect(option)}"
   end
+
+  @doc """
+  Return a string representation of a worker module.any()
```
sorentwo, line 343: This is a perfect fit for a doctest:

```suggestion
  Return a string representation of a worker module. This is particularly useful for normalizing
  worker names when building custom Ecto queries.

  ## Examples
  
  iex> Oban.Worker.to_string(MyApp.SomeWorker)
  "MyApp.SomeWorker"
  
  iex> Oban.Worker.to_string("Elixir.MyApp.SomeWorker")
  "MyApp.SomeWorker"
```

chulkilee: Oh, I didn't know it would work without declaring the module. Added!

lib/oban/worker.ex
```diff
@@ -336,4 +338,32 @@ defmodule Oban.Worker do
   defp validate_opt!(option) do
     raise ArgumentError, "unknown option provided #{inspect(option)}"
   end
+
+  @doc """
+  Return a string representation of a worker module.
+
+  This is particularly useful for normalizing
+  worker names when building custom Ecto queries.
```
sorentwo, line 346:
```suggestion
  This is particularly useful for normalizing worker names when building custom Ecto queries.
```

sorentwo:
Thanks for seeing this through @chulkilee!

PR #287, opened 2020-07-04
Add installation guide to Hexdocs

JonRowe:
Currently, there are no installation instructions on Hexdocs for Oban and the only reference at all to the migrations is in the isolating docs, and theres nothing to say "hey go read the readme" (which as obvious as it might seem, was not obvious to me that there would be a detailed unpublished readme!).

This is not an ideal solution I'll admit, as its not automated from the readme like the main Oban module docs. I could prepare an alternative (if you prefer) that documents the migration module via pull that section from the read me, and link to it from the main module docs?

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,9 @@ enqueuing scheduled jobs and job dispatching altogether when testing:
 config :my_app, Oban, crontab: false, queues: false, plugins: false
 ```
 
+See the installation instructions in the README or on the Hexdocs guide for details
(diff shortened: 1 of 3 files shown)


sorentwo:
@JonRowe This is a great start on extracting the basic guides out of the README, thank you. As you undoubtedly noticed the README is no longer published on hexdocs, with the intention being to transition to guides just like this.

JonRowe:
Thank you!

PR #290, opened 2020-07-10
Clear up worker snooze documentation

novaugust:
I was confused by the worker snooze documentation and asked @milmazz for some clarification. I think (?) this summarizes its behavior.

It's unclear to me in what way a job is marked as a "success" when it's snoozed. Instead, it's simply snoozed, right?

I rewrote the second sentence because it seemed like leaking an implementation detail into the documentation. I hope this maintains the intent - reassure the user that they can snooze w/o negatively impacting their retry logic - w/o confusing them with how exactly snoozing is implemented.


sorentwo:
I think this helps to get the point across without sharing too much about the implementation. Thanks!

PR #292, opened 2020-07-10
Add missing argument in README

nathanl:
If this is not correct, just close the PR. :)


sorentwo:
That would be correct for Oban 1.2, but not for Oban 2.0 =)

PR #301, opened 2020-07-23
fix test assertion for modules with multiple behaviours

### LINKED ISSUE #300: Oban.Testing.implements_worker?/1 fails if the worker implements another @behaviour first
### Environment

- Oban Version `2.0.0`
- PostgreSQL Version `11.5`
```
Erlang/OTP 22 [erts-10.7.2.1] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]

Elixir 1.8.2 (compiled with Erlang/OTP 22)
```

### Current Behavior

If I have a worker module that looks like this:

```elixir
defmodule MyApp.FooWorker do
  use MyApp.Worker
  use Oban.Worker
end

defmodule MyApp.Worker do
  defmacro __using__(_opts) do
    quote do
      @behaviour unquote(__MODULE__)
    end
  end
end
```

Then calling `perform_job(MyApp.FooWorker, %{})` fails with

```
      Expected worker to be a module that implements the Oban.Worker behaviour, got:
     
     MyApp.FooWorker

```

This seems to be due to the fact that `Oban.Testing.implements_worker?/1` relies on `Oban.Worker` being the first `@behaviour` defined in a worker module. In this contrived example, I could just flip the `use` ordering and all would be well, but in my actual use case that's not an option.

### Expected Behavior

`@behaviour` ordering shouldn't matter.

superhawk610:
Fixes #300.

When a module implements multiple `@behaviour`s, calling `__info__(:attributes)` returns something like

```elixir
[
  vsn: [12345],
  behaviour: [MyBehaviour],
  behaviour: [Oban.Worker]
]
```

This PR takes all `:behaviour` keys into account when checking for `Oban.Worker`.

I wasn't sure of the best way to implement the test so I just put in a quick placeholder.

### DIFF
--- a/lib/oban/testing.ex
+++ b/lib/oban/testing.ex
@@ -339,7 +339,8 @@ defmodule Oban.Testing do
   defp implements_worker?(worker) do
     :attributes
     |> worker.__info__()
-    |> Keyword.get(:behaviour, [])
+    |> Keyword.get_values(:behaviour)
+    |> Enum.flat_map(& &1)
     |> Enum.member?(Oban.Worker)
   end
 
--- a/test/oban/testing_test.exs
+++ b/test/oban/testing_test.exs
@@ -20,12 +20,23 @@ defmodule Oban.TestingTest do
     def perform(%{args: %{"action" => "bad_snooze"}}), do: {:snooze, true}
   end
 
+  defmodule MultiBehaviourWorker do
+    @behaviour SomeOtherBehaviour
+    use Oban.Worker
+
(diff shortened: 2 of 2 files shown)


sorentwo:
Sorry, I pushed a fix for this already. I should have waited to see if you were following up the issue with a PR! Thanks for the report and the attempted fix 💛

superhawk610:
No worries, I should have mentioned it in the issue. Thanks for the quick fix!

PR #308, opened 2020-08-07
Wrap worker crashes and tuples in exceptions

### LINKED ISSUE #305: Unhelpful error when job fails
### Environment

- Oban Version

2.0.0

- PostgreSQL Version

12.3

Erlang/OTP 23 [erts-11.0.2] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]

Elixir 1.10.4 (compiled with Erlang/OTP 21)

### Current Behavior

We're receiving an unhelpful error when a job fails to execute.

```
** (ErlangError) Erlang error: %{}\n    (elixir 1.10.4) lib/process.ex:765: Process.info/2\n    (oban 2.0.0) lib/oban/queue/executor.ex:223: Oban.Queue.Executor.current_stacktrace/0\n    (oban 2.0.0) lib/oban/queue/executor.ex:170: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:152: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:69: Oban.Queue.Executor.call/1\n    (elixir 1.10.4) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2\n    (elixir 1.10.4) lib/task/supervised.ex:35: Task.Supervised.reply/5\n
```

We attempt to send errors to Sentry using a telemetry instrumenter but whenever we see this error nothing gets sent to Sentry.

### Expected Behavior

Error messages/stacktraces are helpful and get sent to external error services via the telemetry integration

### Additional Information

Job record

```
%Oban.Job{
  __meta__: #Ecto.Schema.Metadata<:loaded, "integration_shopify", "oban_jobs">,
  args: %{ ... },
  attempt: 5,
  attempted_at: ~U[2020-08-03 19:30:36.864663Z],
  attempted_by: ["integration_shopify@10.244.1.152", "events", "84kquhuh"],
  completed_at: nil,
  discarded_at: ~U[2020-08-03 19:30:36.881854Z],
  errors: [
    %{
      "at" => "2020-08-03T19:29:03.980467Z",
      "attempt" => 1,
      "error" => "** (ErlangError) Erlang error: %{}\n    (elixir 1.10.4) lib/process.ex:765: Process.info/2\n    (oban 2.0.0) lib/oban/queue/executor.ex:223: Oban.Queue.Executor.current_stacktrace/0\n    (oban 2.0.0) lib/oban/queue/executor.ex:170: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:152: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:69: Oban.Queue.Executor.call/1\n    (elixir 1.10.4) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2\n    (elixir 1.10.4) lib/task/supervised.ex:35: Task.Supervised.reply/5\n"
    },
    %{
      "at" => "2020-08-03T19:29:21.881709Z",
      "attempt" => 2,
      "error" => "** (ErlangError) Erlang error: %{}\n    (elixir 1.10.4) lib/process.ex:765: Process.info/2\n    (oban 2.0.0) lib/oban/queue/executor.ex:223: Oban.Queue.Executor.current_stacktrace/0\n    (oban 2.0.0) lib/oban/queue/executor.ex:170: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:152: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:69: Oban.Queue.Executor.call/1\n    (elixir 1.10.4) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2\n    (elixir 1.10.4) lib/task/supervised.ex:35: Task.Supervised.reply/5\n"
    },
    %{
      "at" => "2020-08-03T19:29:41.807279Z",
      "attempt" => 3,
      "error" => "** (ErlangError) Erlang error: %{}\n    (elixir 1.10.4) lib/process.ex:765: Process.info/2\n    (oban 2.0.0) lib/oban/queue/executor.ex:223: Oban.Queue.Executor.current_stacktrace/0\n    (oban 2.0.0) lib/oban/queue/executor.ex:170: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:152: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:69: Oban.Queue.Executor.call/1\n    (elixir 1.10.4) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2\n    (elixir 1.10.4) lib/task/supervised.ex:35: Task.Supervised.reply/5\n"
    },
    %{
      "at" => "2020-08-03T19:30:04.986102Z",
      "attempt" => 4,
      "error" => "** (ErlangError) Erlang error: %{}\n    (elixir 1.10.4) lib/process.ex:765: Process.info/2\n    (oban 2.0.0) lib/oban/queue/executor.ex:223: Oban.Queue.Executor.current_stacktrace/0\n    (oban 2.0.0) lib/oban/queue/executor.ex:170: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:152: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:69: Oban.Queue.Executor.call/1\n    (elixir 1.10.4) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2\n    (elixir 1.10.4) lib/task/supervised.ex:35: Task.Supervised.reply/5\n"
    },
    %{
      "at" => "2020-08-03T19:30:36.881864Z",
      "attempt" => 5,
      "error" => "** (ErlangError) Erlang error: %{}\n    (elixir 1.10.4) lib/process.ex:765: Process.info/2\n    (oban 2.0.0) lib/oban/queue/executor.ex:223: Oban.Queue.Executor.current_stacktrace/0\n    (oban 2.0.0) lib/oban/queue/executor.ex:170: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:152: Oban.Queue.Executor.perform_inline/1\n    (oban 2.0.0) lib/oban/queue/executor.ex:69: Oban.Queue.Executor.call/1\n    (elixir 1.10.4) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2\n    (elixir 1.10.4) lib/task/supervised.ex:35: Task.Supervised.reply/5\n"
    }
  ],
  id: 13,
  inserted_at: ~U[2020-08-03 19:29:03.844930Z],
  max_attempts: 5,
  priority: 0,
  queue: "events",
  scheduled_at: ~U[2020-08-03 19:30:35.986078Z],
  state: "discarded",
  tags: [],
  unique: nil,
  unsaved_error: nil,
  worker: "MyApp.Workers.CreateOrder"
}
```

sorentwo:
This makes several significant changes to how crahes, errors and timeouts are reported from `perform/1` calls:

* Timeouts are wrapped in `Oban.TimeoutError`
* Error and discard tuples are wrapped in `Oban.PerformError`
* Exits and throws are wrapped in `Oban.CrashError`
* Stacktraces are only included from code that is rescued or caught, not
  from error tuples or timeouts.

The goal is improve error formatting within a job's error array and to make error reporting to external services like Sentry entirely consistent.

Fixes #305

/cc @anthonator @coladarci

### DIFF
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -3,7 +3,7 @@ defmodule Oban.Queue.Executor do
 
   require Logger
 
-  alias Oban.{Breaker, Config, Job, Query, Worker}
+  alias Oban.{Breaker, Config, CrashError, Job, PerformError, Query, TimeoutError, Worker}
 
   @type success :: {:success, Job.t()}
   @type failure :: {:failure, Job.t(), Worker.t(), atom(), term()}
@@ -30,7 +30,6 @@ defmodule Oban.Queue.Executor do
     :conf,
     :error,
     :job,
-    :kind,
     :meta,
     :snooze,
     :start_mono,
@@ -39,6 +38,7 @@ defmodule Oban.Queue.Executor do
     :worker,
     safe: true,
     duration: 0,
+    kind: :error,
     queue_time: 0,
     stacktrace: [],
     state: :unset
@@ -147,9 +147,14 @@ defmodule Oban.Queue.Executor do
 
   defp perform_inline(%{safe: true} = exec) do
     perform_inline(%{exec | safe: false})
+  rescue
+    error ->
+      %{exec | state: :failure, error: error, stacktrace: __STACKTRACE__}
   catch
     kind, value ->
-      %{exec | state: :failure, kind: kind, error: value, stacktrace: __STACKTRACE__}
+      error = CrashError.exception({kind, value, __STACKTRACE__})
+
+      %{exec | state: :failure, error: error, stacktrace: __STACKTRACE__}
   end
 
   defp perform_inline(%{worker: worker, job: job} = exec) do
@@ -161,13 +166,13 @@ defmodule Oban.Queue.Executor do
         %{exec | state: :success}
 
       :discard ->
-        %{exec | state: :discard, kind: :error, error: "None", stacktrace: current_stacktrace()}
+        %{exec | state: :discard, error: PerformError.exception({worker, :discard})}
 
       {:discard, reason} ->
-        %{exec | state: :discard, kind: :error, error: reason, stacktrace: current_stacktrace()}
+        %{exec | state: :discard, error: PerformError.exception({worker, reason})}
 
-      {:error, error} ->
-        %{exec | state: :failure, kind: :error, error: error, stacktrace: current_stacktrace()}
+      {:error, reason} ->
+        %{exec | state: :failure, error: PerformError.exception({worker, reason})}
 
       {:snooze, seconds} ->
         %{exec | state: :snoozed, snooze: seconds}
@@ -204,7 +209,9 @@ defmodule Oban.Queue.Executor do
         reply
 
       nil ->
-        %{exec | state: :failure, kind: :error, error: :timeout, stacktrace: current_stacktrace()}
+        error = TimeoutError.exception({exec.worker, timeout})
+
+        %{exec | state: :failure, error: error}
     end
   end
 
@@ -224,12 +231,6 @@ defmodule Oban.Queue.Executor do
       Map.merge(exec.meta(), %{kind: exec.kind, error: exec.error, stacktrace: exec.stacktrace})
 
     :telemetry.execute([:oban, :job, :exception], measurements, meta)
-  end
-
-  defp current_stacktrace do
-    self()
-    |> Process.info(:current_stacktrace)
-    |> elem(1)
   end
 
   defp event_metadata(conf, job) do
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -151,8 +151,8 @@ defmodule Oban.Queue.Executor do
     error ->
       %{exec | state: :failure, error: error, stacktrace: __STACKTRACE__}
   catch
-    kind, value ->
-      error = CrashError.exception({kind, value, __STACKTRACE__})
+    kind, reason ->
+      error = CrashError.exception({kind, reason, __STACKTRACE__})
 
       %{exec | state: :failure, error: error, stacktrace: __STACKTRACE__}
   end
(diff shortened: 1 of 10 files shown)


anthonator:
Should this include a README.md update on how to report errors to an external service like Sentry?

sorentwo:
> Should this include a README.md update on how to report errors to an external service like Sentry?

Yes it should! Reverting my recent change.

sorentwo:
I'll leave this open for a little bit in case people have some feedback.

lib/oban/queue/executor.ex
```diff
@@ -147,9 +147,14 @@ defmodule Oban.Queue.Executor do
 
   defp perform_inline(%{safe: true} = exec) do
     perform_inline(%{exec | safe: false})
+  rescue
+    error ->
+      %{exec | state: :failure, error: error, stacktrace: __STACKTRACE__}
```
coladarci, line 152: I am correct that if the worker throws an exception directly (function mismatch, etc), then it'll just pass it on raw, not wrapped in `PerformError.exception`?

sorentwo: That's right. This won't wrap a raised exception. Though the way it will wrap an exception returned in an error tuple, which may not be desirable.

coladarci: i.e if I return `{:error, %{something: :strange, but: :important}}` from my worker, I will lose that structured data and instead just get the exception w/ `%{something: :strange, but:important}` encoded in the message?

sorentwo: No, you'll get the raw value you return. Only the formatted error stored in `errors` and the value passed to telemetry.

coladarci: > No, you'll get the raw value you return.

Where? The telemetry error handler only gets the wrapped error, right? I'm not sure where else that error would surface.

sorentwo: I meant that the value is still returned from perform/1, which would only apply in testing. I see what you mean now.

The exception struct itself could have a `reason` field added that stored the original value. What do you think?

coladarci: 100%. I don't see why you'd want to hide that structured data when it's easy to pass on. Maybe something like `worker_response` to make it explicit? And have it be the actual `{:error, :not_found}` `etc` so there's less "magic"?

sorentwo: @coladarci I ended up using a `reason` field for consistency between the various Oban exceptions and it uses the exact value that was returned. Meaning it uses `{:error, :not_found}` or `{:discard, :some_reason}` for `PerformError`, and the value returned from caught exceptions for `CrashError`.

PR #312, opened 2020-08-17
Fix Telemetry ErrorReporter example doc

rockwood:
Oban 2.0 changed `meta.stack` to `meta.stacktrace`. The main docs were updated, but it looks like this one was missed.


sorentwo:
Nice catch, thanks for the fix!

PR #314, opened 2020-08-20
Add support for sub-arg uniquness with :keys`

### LINKED ISSUE #283: Unique worker arg fields

We are scraping news articles and enqueue each url we find to a separate downloader worker. The issue is that we sometimes find the same url on different places but we only want to enqueue it once… however we include more data than just the url in the args which makes the unique feature of Oban not work.

Ideally an additional unique field, like `:keys` that you'd specify in the worker:

```
use Oban.Worker, unique: [period: :infinity, keys: [:url]]
```

Using a partial unique index for the specific unique args but that gets cumbersome since you need to add migration scripts for each worker.

sorentwo:
This makes it possible to only use a few values from a job to determine uniqueness. For example, a job that includes a lot of metadata but must be considered unique based on a `:url` field can specify `keys: [:url]`.

Closes #283

/cc @hfjallemark


lib/oban/query.ex
```diff
@@ -254,6 +255,34 @@ defmodule Oban.Query do
...
+    queue = Changeset.get_field(changeset, :queue)
+
+    dynamic([j], j.queue == ^queue and ^acc)
+  end
+
+  defp unique_field({changeset, :worker, _}, acc) do
+    worker = Changeset.get_field(changeset, :worker)
+
+    dynamic([j], j.worker == ^worker and ^acc)
+  end
```
fuelen, line 284: as an alternative to these 3 definitions
```
  defp unique_field({changeset, field, _}, acc) when field in [:args, :queue, :worker] do
    value = Changeset.get_field(changeset, field)

    dynamic([j], field(j, ^field) == ^value and ^acc)
  end
```

sorentwo: Yes! The `field/2` syntax is what I was looking for earlier. Updated without the field guard, those are already validated by this point.

README.md
```diff
@@ -506,6 +506,10 @@ level using the following options:
 * `:fields` — The fields to compare when evaluating uniqueness. The available
   fields are `:args`, `:queue` and `:worker`, by default all three are used.
 
+* `:keys` — A specific subset of the `:args` to consider when comparing against
+  historic jobs. This allows a job with multiple key/value pairs in the args to
+  be compared using only one or more of the key/values.
```
milmazz, line 511:
```suggestion
  be compared using only a subset of them.
```

PR #322, opened 2020-09-01
Adds support for weekday ranges in crontab

jmnsf:
As the title indicates, adds support for [`MON-FRI`](https://crontab.guru/#*_*_*_*_MON-FRI)-like day of week ranges.

I wasn't able to get tests passing on `master`, but this doesn't seem to add any new errors. I do have postgres up locally, happy to look into this further if the readme is out of date and there's some missing step to get them to work.


sorentwo:
Looks perfect! Thanks for the fix. Don't worry about the failing master test, just some race conditions that I'll try and iron out. 💛

PR #329, opened 2020-09-11
Allow replacing the arguments of a unique job

Gazler:
This commit allows queing a unique job and replacing the args in a
subsequent insert. For example, given the following:

```
%{some_value: 1, id: "123"}
|> MyJob.new(schedule_in: 10, unique: [keys: [:id]])
|> Oban.insert()
```

Calling

```
%{some_value: 2, id: "123"}
|> MyJob.new(schedule_in: 10, replace_args: true unique: [keys: [:id]])
|> Oban.insert()
```

Would replace the job args with `%{some_value: 2, id: "123"}`

I haven't really considered the API too much, I just wanted to start a discussion around this feature. The use case for me is that there is a relatively expensive operation that I don't want to call too frequently, however I do want it to always be called with the most recent arguments.

There are also some scheduling issues with the above. For example, if you wanted it to "debounce" then it would be necessary to allow `schedule_in` to be modified.

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -41,6 +41,7 @@ defmodule Oban.Job do
           | {:priority, pos_integer()}
           | {:queue, atom() | binary()}
           | {:schedule_in, pos_integer()}
+          | {:replace_args, boolean()}
           | {:scheduled_at, DateTime.t()}
           | {:tags, tags()}
           | {:unique, [unique_option()]}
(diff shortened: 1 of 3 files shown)


sorentwo:
@Gazler Beautiful! The implementation looks excellent, and this is something that was discussed/requested on Slack recently.

I'm not too concerned about the scheduling issue. It seems secondary to this solution, and unique jobs already provide a "throttle" rather than a "debounce".

sorentwo:
@Gazler This has conflicts now due to a huge PR that was just merged. Can you fit in a rebase?

sorentwo (approved):

sorentwo:
Thanks @Gazler, great stuff!

Issue #326, opened 2020-09-09
Add support for dynamic repos


sasa1977:
### Is your feature request related to a problem? Please describe.

The team I'm working with needs to use dynamic databases. Multiple instances of the same repo will be started (and these can change at runtime), and we want to manage an oban queue in each database.

The Ecto approach for dynamic repos is to use [put_dynamic_repo](https://hexdocs.pm/ecto/Ecto.Repo.html?#c:put_dynamic_repo/1) to configure the instance pid for the caller process. From what I can tell Oban currently doesn't support this.

### Describe the Solution You'd Like

I'm not sure about the best course of actions, but clearly we need some way of passing the repo instance pid to Oban. At the moment I'm inclined to propose an option which requires one oban instance per dynamic repo instance. The question is how can we pass the repo pid to the oban instance? One option is to pass the pid via oban startup options. In this case, the oban instance would be tightly coupled with the repo instance. The client code would then have to start the repo instance first, get the instance pid, and then start the oban instance passing it the repo pid. Oban would have to link itself to the repo to make sure it is stopped if the instance stops. 

### Describe Alternatives You've Considered

A variation of this is to have clients pass some logical instance name (e.g. `{:my_tenant, tenant_id}`) and a callback function which can find the repo pid given the name. A theoretical benefit is that we don't need to restart the oban instance if the repo is restarted, but I feel that this benefit is academic, because if the repo is permanently stopped we still want to stop the oban instance, so we don't really remove any work from the client code.

Another option is to use a global oban instance, and expose interface functions such as `add_repo_instance` and `remove_repo_instance`.  As soon as repo instance is added, Oban would start the set of required processes (such as pollers, cleaners, etc.). The client code would need to pass the repo instance as an option when invoking functions such as `insert`. I'm not sure that this would simplify things for clients, and I'm worried it would complicate the implementation of oban significantly, so it's not an option I'm currently inclined to propose.

### Additional Context

Glancing at the code, I feel that Oban is a bit rigid with respect to dynamic instances. If we want to run multiple oban instances we need to provide a locally registered alias as an atom, which is not perfect for dynamic instances. Since the client code needs to provide a unique name, it would have to pass something like `:"tenant_#{tenant.id}" as the name option, which can overflow the atom table and lead to the system crash. 

Admittedly this feels more like a theoretical issue that is unlikely to cause problems in production. According to the current oban code and BEAM system limit defaults, it would take about 250k of different names generated over time to crash the beam.

[...]

PR #331, opened 2020-09-12
Replace local names with registry

sasa1977:
This replaces locally registered names with Registry based names, which allows us to use arbitrary terms as oban instance names, e.g. `{Oban, name: {:tenant, 1}}`. This is the preparation for supporting running dynamic Oban instances which can work with dynamic Ecto repos, as discussed in #326.

In terms of function signatures, the change should be backwards compatible. However, if the client code relied on the previous name registration, it will break.

Summary of changes:

- Oban now has its own supervision tree with a registry which is used for process discovery.
- Any term can be used as name. Note that name still has to be unique, i.e. it's not possible to start two instances with the same name.
- Configuration is also managed in the registry.
- To improve confidence, integration tests are adapted to use a unique non-atom name.
- Name is also used as the default child id in `Oban.child_spec`.

### DIFF
--- a/lib/oban/registry.ex
+++ b/lib/oban/registry.ex
@@ -1,9 +1,9 @@
 defmodule Oban.Registry do
   @moduledoc false
 
-  @type role :: term
-  @type key :: Oban.name() | {Oban.name(), role}
-  @type value :: term
+  @type role :: term()
+  @type key :: Oban.name() | {Oban.name(), role()}
+  @type value :: term()
 
   def child_spec(_arg) do
     Supervisor.child_spec(
@@ -18,10 +18,10 @@ defmodule Oban.Registry do
     config
   end
 
-  @spec whereis(Oban.name(), role) :: pid | nil
+  @spec whereis(Oban.name(), role()) :: pid() | nil
   def whereis(oban_name, role \\ nil), do: GenServer.whereis(via(oban_name, role))
 
-  @spec via(Oban.name(), role, value) :: {:via, Registry, {__MODULE__, key}}
+  @spec via(Oban.name(), role(), value()) :: {:via, Registry, {__MODULE__, key()}}
   def via(oban_name, role \\ nil, value \\ nil)
   def via(oban_name, role, nil), do: {:via, Registry, {__MODULE__, key(oban_name, role)}}
   def via(oban_name, role, value), do: {:via, Registry, {__MODULE__, key(oban_name, role), value}}
--- a/lib/oban/registry.ex
+++ b/lib/oban/registry.ex
@@ -3,6 +3,7 @@ defmodule Oban.Registry do
 
   @type role :: term
   @type key :: Oban.name() | {Oban.name(), role}
+  @type value :: term
 
   def child_spec(_arg) do
     Supervisor.child_spec(
@@ -11,11 +12,19 @@ defmodule Oban.Registry do
     )
   end
 
+  @spec config(Oban.name()) :: Oban.Config.t()
+  def config(oban_name) do
+    [{_pid, config}] = Registry.lookup(__MODULE__, oban_name)
+    config
+  end
+
   @spec whereis(Oban.name(), role) :: pid | nil
   def whereis(oban_name, role \\ nil), do: GenServer.whereis(via(oban_name, role))
 
-  @spec via(Oban.name(), role) :: {:via, Registry, {__MODULE__, key}}
-  def via(oban_name, role \\ nil), do: {:via, Registry, {__MODULE__, key(oban_name, role)}}
+  @spec via(Oban.name(), role, value) :: {:via, Registry, {__MODULE__, key}}
+  def via(oban_name, role \\ nil, value \\ nil)
+  def via(oban_name, role, nil), do: {:via, Registry, {__MODULE__, key(oban_name, role)}}
+  def via(oban_name, role, value), do: {:via, Registry, {__MODULE__, key(oban_name, role), value}}
 
   defp key(oban_name, nil), do: oban_name
   defp key(oban_name, role), do: {oban_name, role}
--- a/lib/oban/registry.ex
+++ b/lib/oban/registry.ex
@@ -1,6 +1,9 @@
 defmodule Oban.Registry do
   @moduledoc false
 
+  @type role :: term
+  @type key :: Oban.name() | {Oban.name(), role}
+
   def child_spec(_arg) do
     Supervisor.child_spec(
       Registry.child_spec(keys: :unique, name: __MODULE__),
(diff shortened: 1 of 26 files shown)


lib/oban/telemetry.ex
```diff
@@ -205,6 +205,26 @@ defmodule Oban.Telemetry do
 
   defp duration(start_mono), do: System.monotonic_time() - start_mono
 
+  @doc false
+  def execute(event_name, measurements, meta) do
+    :telemetry.execute(event_name, measurements, normalize_meta(meta))
+  end
+
+  defp normalize_meta(%{name: {:via, Registry, {Oban.Registry, {_pid, name}}}} = meta) do
```
sorentwo, line 213: It took me a little reading through the telemetry docs to realize where the `name` is even used. Nice catch 👍

test/integration/controlling_test.exs
```diff
@@ -12,33 +12,33 @@ defmodule Oban.Integration.ControllingTest do
     end
 
     test "starting individual queues dynamically" do
-      start_supervised_oban!(queues: [alpha: 10])
+      name = start_supervised_oban!(queues: [alpha: 10]).name
```
sorentwo, line 15: The look of this may take me some getting used to. The fact that it is a unique reference for every test is fantastic though.

test/test_helper.exs
```diff
@@ -44,12 +44,15 @@ defmodule Oban.Case do
...
+      |> Keyword.put_new(:name, make_ref())
       |> Keyword.put_new(:repo, Repo)
       |> Keyword.put_new(:poll_interval, 25)
       |> Keyword.put_new(:shutdown_grace_period, 1)
 
-    start_supervised!({Oban, opts}, id: opts[:name])
+    name = opts[:name]
+    pid = start_supervised!({Oban, opts}, id: name)
+
+    %{name: name, pid: pid}
```
sorentwo, line 55: I may have missed it, but I didn't notice anywhere that the `pid` is used. It could clean up a lot of test noise if this only returned the name.

test/test_helper.exs
```diff
@@ -44,12 +44,15 @@ defmodule Oban.Case do
...
       opts
-      |> Keyword.put_new(:name, Oban)
+      |> Keyword.put_new(:name, make_ref())
       |> Keyword.put_new(:repo, Repo)
       |> Keyword.put_new(:poll_interval, 25)
       |> Keyword.put_new(:shutdown_grace_period, 1)
 
-    start_supervised!({Oban, opts}, id: opts[:name])
+    name = opts[:name]
+    pid = start_supervised!({Oban, opts}, id: name)
```
sorentwo, line 53: Not necessary after the `child_spec/1` override:

```suggestion
    pid = start_supervised!({Oban, opts})
```

test/integration/uniqueness_test.exs
```diff
@@ -36,71 +36,93 @@ defmodule Oban.Integration.UniquenessTest do
...
-             |> Oban.insert(:job_1, UniqueWorker.new(%{id: 1}))
-             |> Oban.insert(:job_2, UniqueWorker.new(%{id: 2}))
-             |> Oban.insert(:job_3, UniqueWorker.new(%{id: 1}))
-             |> Repo.transaction()
+  test "inserting unique jobs within a multi transaction", context do
+    multi = Multi.new()
+    multi = Oban.insert(context.name, multi, :job_1, UniqueWorker.new(%{id: 1}))
+    multi = Oban.insert(context.name, multi, :job_2, UniqueWorker.new(%{id: 2}))
+    multi = Oban.insert(context.name, multi, :job_3, UniqueWorker.new(%{id: 1}))
+    assert {:ok, %{job_1: job_1, job_2: job_2, job_3: job_3}} = Repo.transaction(multi)
```
sorentwo, line 125: What a shame, this was a nice pipe chain 😢

test/integration/isolation_test.exs
```diff
@@ -6,18 +6,18 @@ defmodule Oban.Integration.IsolationTest do
...
   test "multiple supervisors can be run simultaneously" do
-    start_supervised_oban!(name: ObanA, queues: [alpha: 1], plugins: [Oban.Plugins.Pruner])
-    start_supervised_oban!(name: ObanB, queues: [gamma: 1], plugins: [Oban.Plugins.Pruner])
+    name1 = start_supervised_oban!(queues: [alpha: 1], plugins: [Oban.Plugins.Pruner]).name
+    start_supervised_oban!(queues: [gamma: 1], plugins: [Oban.Plugins.Pruner])
 
-    insert!(ObanA, %{ref: 1, action: "OK"}, [])
+    insert!(name1, %{ref: 1, action: "OK"}, [])
 
     assert_receive {:ok, 1}
```
sorentwo, line 14: This test loses a bit of its pizazz now that _all_ of the names are dynamic. I don't believe we need to insert a job to prove that it works. Simply starting up the separate supervision trees proves that there aren't conflicts.

```elixir
name1 = start_supervised_oban!(queues: [gamma: 1], plugins: [Oban.Plugins.Pruner])
name2 = start_supervised_oban!(queues: [gamma: 1], plugins: [Oban.Plugins.Pruner])

assert name1 != name2
```

sorentwo (commented):
Thanks for this **amazing work**. The global registry, storing config, using the name as the id automatically, and the dramatically improved isolation guarantees are brilliant.

I've left some nitpicks, nearly all of which are in the name of consistency and to satisfy my neurosis. This is a great start to a v2.2.0 🎉

lib/oban/registry.ex
```diff
@@ -0,0 +1,31 @@
...
+    )
+  end
+
+  @spec config(Oban.name()) :: Oban.Config.t()
+  def config(oban_name) do
+    [{_pid, config}] = Registry.lookup(__MODULE__, oban_name)
+    config
+  end
+
+  @spec whereis(Oban.name(), role) :: pid | nil
```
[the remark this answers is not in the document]
sasa1977: `nil` is definitely not implied as it is a concrete type (atom of value `:nil`). Omitting it would mean that `whereis` will always return `pid`, so it is needed here. See also [GenServer.whereis spec](https://hexdocs.pm/elixir/GenServer.html#whereis/1).

lib/oban/telemetry.ex
sasa1977: It was caught by the tests :-) Good job on making them exhaustive :+1:

test/integration/controlling_test.exs
sasa1977: Hopefully it's a bit simpler now, after I applied your proposal of only returning the name.

test/integration/uniqueness_test.exs
sasa1977: What can I say. You win some, you lose some :-)

sasa1977:
I think I addressed everything. Let me know if something else is needed.

sorentwo:
That was everything. This is exciting stuff, puno hvala! 💛

Issue #323, opened 2020-09-02
Storing job metadata


v0idpwn:
### Is your feature request related to a problem? Please describe.

It's not exactly a problem but, when jobs fail, I wish they had more context.

### Describe the Solution You'd Like

A new :map field for Oban.Job called `metadata`. This would allow me to store all the info I want on a place dedicated to this purpose during the insertion.

### Describe Alternatives You've Considered

- Passing metadata as a param, which I personally think is inadequate
- Using the tag system, which I found out that doesn't work well for this purpose
- Storing on a new table, which I think is sub-optimal

### Additional Context

Example use-case: storing the request id which triggered the job to make debugging easier

gugahoa:
The example use-case is exactly the same I also have.
Right now we use OpenTelemetry at work, and to propagate trace context correctly I'm adding to the job arguments the relevant propagation values, and on `[:oban, :job, :start]` telemetry event I'm injecting that context back.

It would be awesome to be able to somehow add the `metadata` on job insert.
I'll share more on how I'm approaching this tomorrow.

sorentwo:
When I initially introduced the `tags` array I considered adding a `meta` jsonb column instead, but decided against it. My thinking at the time was that anything people would put in `meta` that wasn't some tag-like data could be put into `args`. 

The `args` field is a free form map and as of Oban 2.0 it doesn't have any special handling. You can match on any fields of the `args` map, or ignore the ones you don't care about.

I'm willing to hear more about your requirements and use case. However, adding a new column would require a new migration, which I'm _very_ hesitant to do at this point.

gugahoa:
Almost forgot to post how I'm handling it:

I created a wrapper around `Oban.Worker` to inject the trace context inside `args`, when calling to `new/2`
```elixir
defmodule Tracing.ObanWorker do
  @moduledoc """
  Tracing.ObanWorker is a wrapper around Oban.Worker that defines a
  `new/2` function with span propagation logic
  """

  @doc false

  defmacro __using__(opts) do
    quote location: :keep do
      use Oban.Worker, unquote(opts)

      require OpenTelemetry.Tracer

      @impl Oban.Worker
      def new(args, opts) when is_map(args) and is_list(opts) do
        args = add_trace_context(args)
        Oban.Job.new(args, Keyword.merge(__opts__(), opts, &Oban.Worker.resolve_opts/3))
      end

      defp add_trace_context(args) do
        add_trace_context(args, :ot_propagation.http_inject([]))
      end

[...]

PR #332, opened 2020-09-13
V9 migration: adds `meta` and `cancelled` state

v0idpwn:
Starting point for the #323 issue. As this requires a migration I ain't sure if it can be merged.

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -67,6 +67,7 @@ defmodule Oban.Job do
           attempted_at: DateTime.t(),
           completed_at: DateTime.t(),
           discarded_at: DateTime.t(),
+          cancelled_at: DateTime.t(),
           unique: %{fields: [unique_field()], period: unique_period(), states: [unique_state()]},
           unsaved_error: %{kind: atom(), reason: term(), stacktrace: Exception.stacktrace()}
         }
@@ -86,6 +87,7 @@ defmodule Oban.Job do
     field :attempted_at, :utc_datetime_usec
     field :completed_at, :utc_datetime_usec
     field :discarded_at, :utc_datetime_usec
+    field :cancelled_at, :utc_datetime_usec
     field :inserted_at, :utc_datetime_usec
     field :scheduled_at, :utc_datetime_usec
     field :unique, :map, virtual: true
@@ -100,6 +102,7 @@ defmodule Oban.Job do
     attempted_at
     completed_at
     discarded_at
+    cancelled_at
     errors
     inserted_at
     max_attempts
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -38,6 +38,7 @@ defmodule Oban.Job do
   @type option ::
           {:args, args()}
           | {:max_attempts, pos_integer()}
+          | {:meta, map()}
           | {:priority, pos_integer()}
           | {:queue, atom() | binary()}
           | {:schedule_in, pos_integer()}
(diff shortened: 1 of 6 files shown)


lib/oban/job.ex
```diff
@@ -117,6 +125,7 @@ defmodule Oban.Job do
 
     * `:max_attempts` — the maximum number of times a job can be retried if there are errors
       during execution
+    * `:meta` — a map containing additional information about the job
```
sorentwo, line 128:
```suggestion
    * `:meta` — a map containing additional information about the job that isn't appropriate for `args`
```

lib/oban/migrations.ex
```diff
@@ -441,4 +441,89 @@ defmodule Oban.Migrations do
...
+    def up(prefix) do
+      alter table(:oban_jobs, prefix: prefix) do
+        add_if_not_exists(:meta, :map)
+        add_if_not_exists(:cancelled_at, :utc_datetime_usec)
+      end
+
+      execute """
+      DO $$
+      BEGIN
+      IF EXISTS (SELECT 1 FROM pg_type
```
sorentwo, line 461: For this migration we can safely assume that the `oban_job_state` enum is available. There is no need to do the `IF EXISTS` check within the `DO $$` block.

lib/oban/migrations.ex
```diff
@@ -441,4 +441,89 @@ defmodule Oban.Migrations do
...
+    end
+
+    def down(prefix) do
+      alter table(:oban_jobs, prefix: prefix) do
+        remove_if_exists(:meta, :map)
+        remove_if_exists(:cancelled_at, :utc_datetime_usec)
+      end
+
+      execute """
+      DO $$
```
sorentwo, line 498: Likewise, no need to wrap it in the existence check.

lib/oban/query.ex
```diff
@@ -160,14 +160,33 @@ defmodule Oban.Query do
...
+  @spec cancel_running_job(Config.t(), Job.t()) :: :ok
+  def cancel_running_job(%Config{prefix: prefix, repo: repo, log: log}, job) do
+    updates = [
+      set: [state: "cancelled", cancelled_at: utc_now()],
+      push: [
+        errors: %{attempt: job.attempt, at: utc_now(), error: format_blamed(job.unsaved_error)}
+      ]
+    ]
+
+    repo.update_all(
```
sorentwo, line 180: It looks like you'll need to rebase. This should use `Repo.update_all` from the new `Oban.Repo` instead.

lib/oban/queue/producer.ex
```diff
@@ -148,7 +148,7 @@ defmodule Oban.Queue.Producer do
             with :ok <- DynamicSupervisor.terminate_child(foreman, pid) do
               uerror = %{kind: :error, reason: "None", stacktrace: []}
 
-              Query.discard_job(conf, %{exec.job | unsaved_error: uerror})
+              Query.cancel_running_job(conf, %{exec.job | unsaved_error: uerror})
```
sorentwo, line 151: I think this can just use `cancel_job`. There isn't any value in the unsaved error here as it doesn't have any information.

sorentwo (changes requested):
I have a few minor requests, nothing big. This is looking good. It couldn't have come at a better time! I've hit a spot where having `meta` would be really useful.

sorentwo:
@v0idpwn I made my suggested changes locally, added `meta` to the telemetry metadata and merged this into master. Nice work! 💛

PR #333, opened 2020-09-13
Add `Oban.retry_job/2`

### LINKED ISSUE #309: Add function to restart jobs manually

we would like to restart discarded jobs that were attempted multiple times but because reasons couldn't complete, but after restart would complete successfully

right now i'm doing this, but would like to see this in the lib, and will be willing to contribute it if you accept, i know this is not enough in order to be accepted by you, but just seeing if this is something you would like to havein oban

```elixir
  def restart_oban_job(oban_job) do
	from(job in Oban.Job, where: job.id == ^oban_job.id)
    |> Repo.update_all(set: [state: "available", attempt: 0, discarded_at: nil])
  end
```

v0idpwn:
This function should turn a job available, increasing max_attempts
accordingly. It should ignore executing jobs.

This is a revival of #194 and closes #309

### DIFF
--- a/lib/oban/query.ex
+++ b/lib/oban/query.ex
@@ -105,6 +105,25 @@ defmodule Oban.Query do
     )
   end
 
+  @spec retry_job(Config.t(), pos_integer()) :: :ok
+  def retry_job(%Config{prefix: prefix, repo: repo, log: log}, id) do
+    Job
+    |> where([j], j.id == ^id)
+    |> where([j], j.state not in ["available", "executing", "scheduled"])
+    |> update([j],
+      set: [
+        state: "available",
+        max_attempts: fragment("GREATEST(?, ? + 1)", j.max_attempts, j.attempt),
+        scheduled_at: ^utc_now(),
+        completed_at: nil,
+        discarded_at: nil
+      ]
+    )
+    |> repo.update_all([], log: log, prefix: prefix)
+
+    :ok
+  end
+
   @spec complete_job(Config.t(), Job.t()) :: :ok
   def complete_job(%Config{prefix: prefix, repo: repo, log: log}, %Job{id: id}) do
     repo.update_all(
--- a/test/integration/controlling_test.exs
+++ b/test/integration/controlling_test.exs
@@ -239,6 +239,31 @@ defmodule Oban.Integration.ControllingTest do
     assert_receive {:ok, 1}
   end
 
+  test "retrying jobs from multiple states" do
+    name = start_supervised_oban!(queues: [])
+
+    job_a = insert!(%{ref: 1}, state: "discarded", max_attempts: 20, attempt: 20)
+    job_b = insert!(%{ref: 2}, state: "completed", max_attempts: 20)
+    job_c = insert!(%{ref: 3}, state: "available", max_attempts: 20)
+    job_d = insert!(%{ref: 4}, state: "retryable", max_attempts: 20)
+    job_e = insert!(%{ref: 5}, state: "executing", max_attempts: 1, attempt: 1)
+    job_f = insert!(%{ref: 6}, state: "scheduled", max_attempts: 1, attempt: 1)
+
+    assert :ok = Oban.retry_job(name, job_a.id)
+    assert :ok = Oban.retry_job(name, job_b.id)
+    assert :ok = Oban.retry_job(name, job_c.id)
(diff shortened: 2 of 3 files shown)


lib/oban/query.ex
```diff
@@ -105,6 +105,27 @@ defmodule Oban.Query do
...
+        max_attempts: fragment("GREATEST(?, ? + 1)", j.max_attempts, j.attempt),
+        completed_at: nil,
+        discarded_at: nil
+      ]
+    )
+    |> repo.update_all([])
+
+    # log: log,
+    # prefix: prefix
+
```
v0idpwn, line 125: I'm still figuring out this, I'll amend ASAP

v0idpwn: fix'd

lib/oban/query.ex
```diff
@@ -105,6 +105,24 @@ defmodule Oban.Query do
...
+    |> update([j],
+      set: [
+        state: "available",
+        max_attempts: fragment("GREATEST(?, ? + 1)", j.max_attempts, j.attempt),
+        completed_at: nil,
+        discarded_at: nil
+      ]
+    )
+    |> repo.update_all([], log: log, prefix: prefix)
+
```
v0idpwn, line 122: Questions: 
1. do I need to use any lock?
2. should I avoid `available` jobs to avoid the useless update?

sorentwo: 1. No need for a separate lock, the row-level lock is enough here.
2. It should avoid `scheduled`, `available` and `executing` jobs.

lib/oban/query.ex
```diff
@@ -105,6 +105,24 @@ defmodule Oban.Query do
...
 
+  @spec retry_job(Config.t(), pos_integer()) :: :ok
+  def retry_job(%Config{prefix: prefix, repo: repo, log: log}, id) do
+    Job
+    |> where([j], j.id == ^id)
+    |> where([j], j.state != "executing")
+    |> update([j],
+      set: [
+        state: "available",
+        max_attempts: fragment("GREATEST(?, ? + 1)", j.max_attempts, j.attempt),
```
sorentwo, line 116: 👍

test/integration/controlling_test.exs
```diff
@@ -241,6 +241,30 @@ defmodule Oban.Integration.ControllingTest do
...
+    job_e = insert!(%{ref: 5}, state: "executing", attempt: 1, max_attempts: 1)
+
+    assert :ok = Oban.retry_job(job_a.id)
+    assert :ok = Oban.retry_job(job_b.id)
+    assert :ok = Oban.retry_job(job_c.id)
+    assert :ok = Oban.retry_job(job_d.id)
+    assert :ok = Oban.retry_job(job_e.id)
+
+    assert %Job{state: "available", max_attempts: 21} = Repo.reload(job_a)
+    assert %Job{state: "available", max_attempts: 20} = Repo.reload(job_b)
```
sorentwo, line 260: This relies on the somewhat arbitrary 20 retry limit. Let's set `max_attempts` for all of the various jobs instead to be explicit about it.

lib/oban/query.ex
sorentwo: There is a loose end for `retryable` jobs here as well. I think this should set `scheduled_at` to the current time as well.

test/integration/controlling_test.exs
```diff
@@ -241,6 +241,30 @@ defmodule Oban.Integration.ControllingTest do
...
+    assert %Job{state: "available", max_attempts: 20} = Repo.reload(job_c)
+    assert %Job{state: "available", max_attempts: 20} = Repo.reload(job_d)
+    assert %Job{state: "executing", max_attempts: 1} = Repo.reload(job_e)
+
+    :ok = stop_supervised(Oban)
```
sorentwo, line 265: No need for this shutdown.

sorentwo (changes requested):
Thanks for taking a shot at this. I left some feedback based on your notes. This will also require a rebase after the large PR that was just merged.

v0idpwn:
I probably should've rebased instead of merging this, but we can squash when merging to master, right? This would be a single commit anyway.

sorentwo (approved):

sorentwo:
Thanks @v0idpwn 💛

PR #334, opened 2020-09-13
Add tags as part of the job event metadata

milmazz:
Now the `[:oban, :job, {:start|:stop|:exception}]` telemetry events include the Job tags as part of their metadata.

### DIFF
--- a/test/oban/queue/executor_test.exs
+++ b/test/oban/queue/executor_test.exs
@@ -72,6 +72,22 @@ defmodule Oban.Queue.ExecutorTest do
     end
   end
 
+  describe "new/2" do
+    test "tags are included in the job metadata" do
+      now = DateTime.utc_now()
+
+      job = %Job{
+        args: %{"mode" => "sleep"},
+        worker: to_string(Worker),
+        tags: ["my_tag"],
+        attempted_at: DateTime.add(now, 30, :millisecond),
+        scheduled_at: now
+      }
+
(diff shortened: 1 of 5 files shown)


test/oban/queue/executor_test.exs
```diff
@@ -72,6 +72,22 @@ defmodule Oban.Queue.ExecutorTest do
     end
   end
 
+  describe "new/2" do
+    test "tags are included in the job metadata" do
```
sorentwo, line 76: This test should be included in `test/integration/telemetry_test.exs` instead. Take a look at the assertion around line 50.

sorentwo (changes requested):
I'm surprised this didn't come up before. That may indicate how infrequently people use tags. Anyhow, the changes look great. Get that test swapped over and this is ready to go.

milmazz:
> I'm surprised this didn't come up before. That may indicate how infrequently people use tags. Anyhow, the changes look great. Get that test swapped over and this is ready to go.

I forgot to mention the use case, the team I'm working with is collecting telemetry metrics, we're publishing the worker, node, and a few other entries as tags for those events, and we would like to be able to include the `tags` from the Oban Job too.

Regarding your request, I'll start working asap on this.

milmazz:
@sorentwo I think I addressed your comment. Please let me know what do you think.

sorentwo (approved):

sorentwo:
Thanks @milmazz 💛

PR #335, opened 2020-09-14
enable selectable states for pruning

blisscs:
The pull request will allow the pruning to pick either one of `completed` or `discarded` states to be pruned, instead of using the default i.e., both `completed` and `discarded` to be pruned.

I am proposing this merge request because I think some people might prefer to not delete both completed and discarded jobs but would like to keep discarded jobs to see what jobs have been failed to complete.

### DIFF
--- a/lib/oban/plugins/pruner.ex
+++ b/lib/oban/plugins/pruner.ex
@@ -7,7 +7,7 @@ defmodule Oban.Plugins.Pruner do
 
   import Ecto.Query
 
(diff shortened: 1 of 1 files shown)


sorentwo:
Thanks for the PR, but the basic `Pruner` module is intentionally limited. This is provided by Oban Pro's [DynamicPruner](https://hexdocs.pm/oban/dynamic_pruning.html#providing-overrides).

PR #347, opened 2020-10-01
Support dynamic repos

### LINKED ISSUE #326: Add support for dynamic repos

sasa1977:
resolves #326 

This is a WIP PR which tackles the remaining work for adding dynamic repos support. I didn't include tests, specs, function docs, and `@doc since` tags, because I'd like to first verify if you're ok with this approach from the technical point of view.

I hacked a quick experiment on the project where this is supposed to be used, and it worked properly. Assuming that no significant API changes are requested in this PR, I'd first like to try this branch properly in our branch, to see if everything makes sense. If I see no significant issues, I'll proceed to finish this PR.

Most of the work in this PR revolves around wrapping repo operations. I introduced `Oban.Repo` module which wraps all repo functions currently used in Oban, except for `Ecto.Repo.load`, because that one doesn't require setting prefix or dynamic repo instance pid.

This module is intended to be a part of the public API, because it has to be used by plugins (otherwise they will access the wrong instance, or more likely fail b/c they will try to use the global instance which does not exist).

With those changes in place, adding the support for dynamic repos was straightforward, as can be seen in the last commit.


sorentwo:
Thank you! This looks **fantastic**. It's exactly what we discussed and looks just about perfect.

> I didn't include tests, specs, function docs, and `@doc since` tags, because I'd like to first verify if you're ok with this approach from the technical point of view.

I'm absolutely fine with the approach 👍

sasa1977:
@sorentwo I added some basic tests in 9293b8a7b38f3a30a67a1ee6934493ef1175d857 and I'd like to get some feedback on that. These are very basic tests, just verifying that functionality such as job execution, draining, cron, and pruning works in a dynamic repo. They are not exhaustive, b/c such tests are already done with static repos, so the goal of dynamic repo test is to verify that the basic flow works in a dynamic setting.

What do you think about this approach, and do you see anything else that needs to be tested? Once we complete the tests, I can start wrapping this up by adding specs & docs.

test/integration/dynamic_repo_test.exs
```diff
@@ -0,0 +1,101 @@
...
+    {:ok, repo_pid: repo_pid}
+  end
+
+  test "job execution", context do
+    name = start_oban!(context.repo_pid, queues: [alpha: 1])
+    job_ref = insert_job!(name).ref
+    assert_receive {:ok, ^job_ref}
+  end
+
+  test "draining", context do
```
sorentwo, line 29: The `Drainer` module uses the same `Query` calls as a queue producer and this may not exercise any additional `Repo` calls. Do you think we need it as an additional case?

sorentwo (commented):
As a whole the testing approach looks good to me. You've hit on the major interaction points: queues, pruner plugin, cron scheduler, which has good coverage. I agree that adding dynamic examples to the other integration tests would be overkill and not worth the effort, or the runtime.

I've left a few nitpick suggestions, nothing major at all. Nice job tracking down all of the various test setup/assertions from the various integration tests!

test/integration/dynamic_repo_test.exs
sasa1977: I wasn't sure about the internals, so I decided to include it in tests. If you think it's redundant let me know and I'll remove it.

sasa1977:
I addressed all your comments, except the one about drain test. If you think this is not needed, let me know. Added basic documentation & typespecs. Let me know if you think something else is needed.

README.md
```diff
@@ -993,6 +993,23 @@ def start(_type, _args) do
 end
 ```
 
+### Dynamic repositories
+
+Oban supportd [Ecto dynamic repositories](https://hexdocs.pm/ecto/replicas-and-dynamic-repositories.html#dynamic-repositories) through the `:get_dynamic_repo` option. To make this work, you need to run a separate oban instance per each dynamic repo instance. Most often it's worth bundling each Oban and repo instance under the same supervisor:
```
sorentwo, line 998:
```suggestion
Oban supports [Ecto dynamic repositories](https://hexdocs.pm/ecto/replicas-and-dynamic-repositories.html#dynamic-repositories) through the `:get_dynamic_repo` option. To make this work, you need to run a separate Oban instance per each dynamic repo instance. Most often it's worth bundling each Oban and repo instance under the same supervisor:
```

lib/oban/repo.ex
```diff
@@ -0,0 +1,184 @@
+defmodule Oban.Repo do
+  @moduledoc """
+  Wrappers around `Ecto.Repo` callbacks.
+
+  These functions should be used when working with Ecto repo inside a plugin. These functions will
```
sorentwo, line 5:
```suggestion
  These functions should be used when working with an Ecto repo inside a plugin. These functions will
```

lib/oban/repo.ex
```diff
@@ -0,0 +1,184 @@
...
+  resolve the correct repo instance, and set the schema prefix and the log level, according to the
+  Oban configuration.
+  """
+
+  @type config ::
+          %{
```
sorentwo, line 11: Nicely done. That is a _lot_ of typespec work! 👏

test/integration/dynamic_repo_test.exs
```diff
@@ -0,0 +1,76 @@
+defmodule Oban.Integration.DynamicRepoTest do
+  use Oban.Case
+  import Ecto.Query
+  alias Oban.Test.DynamicRepo
```
sorentwo, line 4:
```suggestion
  use Oban.Case
  
  import Ecto.Query
  
  alias Oban.Test.DynamicRepo
```

test/integration/dynamic_repo_test.exs
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  test "draining", context do
+    name = start_oban!(context.repo_pid, queues: false)
+    job_ref = insert_job!(name).ref
+
+    assert Oban.drain_queue(name, queue: :alpha) == %{failure: 0, success: 1}
+    assert_receive {:ok, ^job_ref}
+  end
+
```
sorentwo, line 34: Alright. I think we can remove it.

```suggestion
```

sorentwo (approved):
Amazing work. This will make authoring plugins much easier and less error prone. I'll apply my most recent suggestions, merge this and start with the 2.2.0 release. 💛

PR #353, opened 2020-10-08
Fix recursive jobs documentation

aerosol:
Removing bang from `Oban.insert!`, otherwise the job results with a
non-expected return.


v0idpwn (approved):

sorentwo:
Thanks for the fix! The guide was pulled from a blog post that was written before Oban checked the return value.

PR #357, opened 2020-10-14
Fix readme

moogle19:
While trying to find my bug with the 2.2 version I came across some minor things in the `README`
- fixed missing `args` binding in Job example
- bumped the Oban version in `Installation` to 2.2
- changed `enqueuing` to`enqueueing` to make it consistent


sorentwo:
Thanks for the doc fixes 💛

PR #358, opened 2020-10-14
Fix `listen/2` function in Notifier

moogle19:
Splitted the `listen/2` function to handle `pid` and `atom` separately.
The function handling `atom` resolves the `pid` using the registry's `whereis` function and then calls `listen/2` with the pid.

I am not sure if the fallback to `Oban` in line 140 is neccessary (the documentation states that the default name is always `Oban` so it could be omitted).

I also removed `is_server` guard because it isn't used anymore.

### DIFF
--- a/lib/oban/notifier.ex
+++ b/lib/oban/notifier.ex
@@ -128,7 +128,7 @@ defmodule Oban.Notifier do
       Oban.Notifier.listen([:gossip, :insert, :signal])
   """
   @spec listen(GenServer.server(), channels :: list(channel())) :: :ok
-  def listen(server \\ __MODULE__, channels)
+  def listen(server \\ Oban, channels)
 
   def listen(pid, channels) when is_pid(pid) and is_list(channels) do
     :ok = validate_channels!(channels)
@@ -136,10 +136,9 @@ defmodule Oban.Notifier do
     GenServer.call(pid, {:listen, channels})
   end
 
-  def listen(server, channels) when is_atom(server) and is_list(channels) do
-    Oban.config()
-    |> Map.get(:name, Oban)
-    |> Oban.Registry.whereis(server)
+  def listen(oban_name, channels) when is_atom(oban_name) and is_list(channels) do
+    oban_name
+    |> Oban.Registry.whereis(__MODULE__)
     |> listen(channels)
   end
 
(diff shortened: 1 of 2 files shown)


lib/oban/notifier.ex
```diff
@@ -130,10 +128,19 @@ defmodule Oban.Notifier do
       Oban.Notifier.listen([:gossip, :insert, :signal])
   """
   @spec listen(GenServer.server(), channels :: list(channel())) :: :ok
-  def listen(server \\ __MODULE__, channels) when is_server(server) and is_list(channels) do
+  def listen(server \\ __MODULE__, channels)
```
sorentwo, line 131: This needs to use the oban instance name, which defaults to `Oban`.

```suggestion
  def listen(oban_name \\ Oban, channels)
```

lib/oban/notifier.ex
```diff
@@ -130,10 +128,19 @@ defmodule Oban.Notifier do
...
 
-    GenServer.call(server, {:listen, channels})
+    GenServer.call(pid, {:listen, channels})
+  end
+
+  def listen(server, channels) when is_atom(server) and is_list(channels) do
+    Oban.config()
+    |> Map.get(:name, Oban)
+    |> Oban.Registry.whereis(server)
+    |> listen(channels)
```
sorentwo, line 143: Each Oban instance has its own notifier. This needs to accept an `oban_name` and use the registry from there:

```suggestion
  def listen(oban_name, channels) when is_atom(oban_name) and is_list(channels) do
    oban_name
    |> Registry.whereis(__MODULE__)
    |> listen(channels)
```

sorentwo (changes requested):
This is a great start. I left a few suggestions to move this toward using the "oban name" instead. It would be nice to have documentation on using an alternate name.

lib/oban/notifier.ex
moogle19: Okay. I thought you name the notifier and not the Oban instance (just read that you could start more than one). But hat makes sense.

moogle19:
@sorentwo 
I updated the PR according to your suggestions.

sorentwo (approved):
The changes look great. I've pushed fixes for the failing tests and they should be resolved on a re-deploy. Once I have some regression tests in for this I'll release a patch version.

moogle19:
I just rebased and now all tests pass.

sorentwo:
Thanks @moogle19 💛

PR #372, opened 2020-11-05
Fix circuit event descriptions in docs

s22su:
Oban 2.0 changed two telemetry events, but the description blocks in docs
were missed. This commit updates the description blocks to match reality.


sorentwo:
Thanks! 💛

PR #378, opened 2020-11-24
Pruner deletes jobs according to an outdated date configuration 

zaherh8:
We needed more control over the Pruner. As we want it to delete jobs comparing to a specified time (like delete the jobs that are older than 1 day or 1 week..). So in addition of the ``max_age`` configuration, we added an ``outdated`` param that can be passed to the Pruner in order to compare the ``attempted_at`` column with, in the ``Oban.Plugins.Pruner.delete_jobs/3`` (now 4) function.
So now we differentiate between the frequency of the pruner (``max_age``) and the jobs outdated deletion.


sorentwo:
The default pruner is intentionally simple and limited. Advanced pruning controls are a Pro feature and available through a license. Thanks for taking a shot at the feature, though!

PR #380, opened 2020-12-02
Update Worker.from_string/1 to work with modules that are not loaded

wjdix:
Stemming from this conversation in the Elixir slack: https://elixir-lang.slack.com/archives/CRDGH1XCM/p1606924009092100

To summarize: https://github.com/sorentwo/oban/commit/7b53223bf44a6025e93f5eec354a931c4d3a71dd caused Oban to be unable to execute workers that were valid but had not yet been loaded. This caused our test suite to be unable to execute jobs in certain workers.

The change ensures that a module is loaded before checking if it exports `perform/1`.

### DIFF
--- a/test/oban/worker_test.exs
+++ b/test/oban/worker_test.exs
@@ -95,6 +95,16 @@ defmodule Oban.WorkerTest do
     end
   end
 
+  describe "from_string/1" do
+    test "finds workers that are not loaded" do
+      :code.delete(Oban.Integration.Worker)
+
+      {:ok, worker_module} = Worker.from_string("Oban.Integration.Worker")
+
+      assert worker_module == Oban.Integration.Worker
+    end
+  end
+
   test "validating __using__ macro options" do
     assert_raise ArgumentError, ~r/unknown option/, fn ->
       defmodule UnknownOption do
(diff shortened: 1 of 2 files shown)


test/oban/worker_test.exs
```diff
@@ -95,6 +95,16 @@ defmodule Oban.WorkerTest do
     end
   end
 
+  describe "from_string/1" do
+    test "finds workers that are not loaded" do
+      :code.delete(Oban.Integration.Worker)
```
sorentwo, line 100: I believe this may cause race conditions due to async tests. To be safe, this module should have `async: false`.

test/oban/worker_test.exs
```diff
@@ -95,6 +95,16 @@ defmodule Oban.WorkerTest do
...
+
+      {:ok, worker_module} = Worker.from_string("Oban.Integration.Worker")
+
+      assert worker_module == Oban.Integration.Worker
```
sorentwo, line 104:
```suggestion
      assert {:ok, Oban.Integration.Worker} = Worker.from_string("Oban.Integration.Worker")
```

sorentwo (changes requested):
One small change to prevent race conditions and this is good to go!

sorentwo (approved):

sorentwo:
Thanks @wjdix 💛

PR #387, opened 2020-12-17
Fix migration naming collision in older Postgres versions

### LINKED ISSUE #386: _oban_job_state naming collision in Postgres 9.6.x
### Precheck

- Searched the Elixir Forum and the other issues, did not found any other users affected by this (possibly because newer versions are unaffected)

### Environment

- Oban 2.3.4
- PostgreSQL 9.6.2
- Elixir 1.9.0 & Erlang/OTP 22

### Current Behavior

While installing Oban to an existing application using an older version of Postgres (but still 9.6+), during the migration step, there is an unhandled name collision in the temporary `_oban_job_state` datatype, causing the migration to fail:

```
17:21:33.996 [info]  execute "DO $$\nDECLARE\n  version int;\nBEGIN\n  SELECT current_setting('server_version_num')::int INTO version;\n\n  IF version >= 120000 THEN\n    ALTER TYPE public.oban_job_state ADD VALUE IF NOT EXISTS 'cancelled';\n  ELSE\n    ALTER TYPE public.oban_job_state RENAME TO _oban_job_state;\n\n    CREATE TYPE public.oban_job_state AS ENUM (\n      'available',\n      'scheduled',\n      'executing',\n      'retryable',\n      'completed',\n      'discarded',\n      'cancelled'\n    );\n\n    ALTER TABLE public.oban_jobs RENAME column state TO _state;\n\n    ALTER TABLE public.oban_jobs ADD state public.oban_job_state NOT NULL default 'available';\n\n    UPDATE public.oban_jobs SET state = _state::text::public.oban_job_state;\n\n    ALTER TABLE public.oban_jobs DROP column _state;\n    DROP TYPE public._oban_job_state;\n  END IF;\nEND$$;\n"
** (Postgrex.Error) ERROR 42710 (duplicate_object) type "_oban_job_state" already exists
```

This is likely caused by the [implicit Array Type creation](https://www.postgresql.org/docs/12/sql-createtype.html#id-1.9.3.94.5.9) that has improved in more recent version to avoid this kind of conflicts. 

### Expected Behavior

The migration should succeed.

The [Postgres documentation notes](https://www.postgresql.org/docs/12/sql-createtype.html#SQL-CREATETYPE-NOTES) also advise to avoid prepending data type names with underscores:

> It may be advisable to avoid using type and table names that begin with underscore. While the server will change generated array type names to avoid collisions with user-given names, there is still risk of confusion, particularly with old client software that may assume that type names beginning with underscores always represent arrays.

So the fix is actually very straightforward - by replacing all four occurrences of `_oban_job_state` in `migrations.ex` with something else such as `old_oban_job_state` (since this is a temporary type that is deleted by the end of the operation), it started working for me.

gVirtu:
This PR closes #386 by implementing the simple naming change that avoids the collision described in the issue.

### DIFF
--- a/lib/oban/migrations.ex
+++ b/lib/oban/migrations.ex
@@ -471,7 +471,7 @@ defmodule Oban.Migrations do
         ELSIF version >= 120000 THEN
           ALTER TYPE #{prefix}.oban_job_state ADD VALUE IF NOT EXISTS 'cancelled';
         ELSE
-          ALTER TYPE #{prefix}.oban_job_state RENAME TO _oban_job_state;
+          ALTER TYPE #{prefix}.oban_job_state RENAME TO old_oban_job_state;
 
           CREATE TYPE #{prefix}.oban_job_state AS ENUM (
             'available',
@@ -489,7 +489,7 @@ defmodule Oban.Migrations do
           UPDATE #{prefix}.oban_jobs SET state = _state::text::#{prefix}.oban_job_state;
 
           ALTER TABLE #{prefix}.oban_jobs DROP column _state;
-          DROP TYPE #{prefix}._oban_job_state;
+          DROP TYPE #{prefix}.old_oban_job_state;
         END IF;
       END$$;
       """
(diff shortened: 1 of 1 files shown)


v0idpwn (approved):

sorentwo:
Thank you @gVirtu! 💛

PR #392, opened 2020-12-30
Added init telemetry event and tweaked existing telemetry metadata

akoutmos:
Firstly, thanks for the awesome library! This was my first time using Oban and I must say it was super easy to get started and be productive :).

Next, I am currently working on the PromEx Oban plugin and while developing the plugin, I came across some situations where I needed additional metadata related to telemetry events. I added those additional bits to the job stop metadata. I also needed an `init` event (similarly to how Ecto has an 'init' event for repos) so that I can populate a drop down in Grafana for the Oban instance that you want to filter metrics for.

For reference if you want to see how I am using the Oban telemetry events: https://github.com/akoutmos/prom_ex/pull/21. More changes coming to that shortly as I will be using my fork for development.

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -11,7 +11,7 @@ defmodule Oban do
   use Supervisor
 
   alias Ecto.{Changeset, Multi}
-  alias Oban.{Config, Job, Midwife, Notifier, Query, Telemetry, Registry}
+  alias Oban.{Config, Job, Midwife, Notifier, Query, Registry, Telemetry}
   alias Oban.Crontab.Scheduler
   alias Oban.Queue.{Drainer, Producer}
   alias Oban.Queue.Supervisor, as: QueueSupervisor
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -11,7 +11,7 @@ defmodule Oban do
   use Supervisor
 
   alias Ecto.{Changeset, Multi}
-  alias Oban.{Config, Job, Midwife, Notifier, Query, Registry}
+  alias Oban.{Config, Job, Midwife, Notifier, Query, Telemetry, Registry}
   alias Oban.Crontab.Scheduler
   alias Oban.Queue.{Drainer, Producer}
   alias Oban.Queue.Supervisor, as: QueueSupervisor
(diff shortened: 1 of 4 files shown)


lib/oban.ex
```diff
@@ -185,9 +185,24 @@ defmodule Oban do
...
 
+  defp execute_init(conf) do
+    measurements = %{
+      system_time: System.system_time()
+    }
+
+    metadata = %{
+      pid: self(),
+      config: conf
+    }
```
sorentwo, line 201: The data looks good, but let's tighten up the spacing in this already large module:

```suggestion
    measurements = %{system_time: System.system_time()}
    metadata = %{pid: self(), config: conf}
```

lib/oban/telemetry.ex
```diff
@@ -14,7 +29,7 @@ defmodule Oban.Telemetry do
   provide the error type, the error itself, and the stacktrace. The following chart shows which
   metadata you can expect for each event:
 
-  | event        | measures                   | metadata                                                                                                  |
+  | event        | measures                   | metadata                                    |
```
sorentwo, line 32: 👍

lib/oban/queue/executor.ex
```diff
@@ -245,6 +251,7 @@ defmodule Oban.Queue.Executor do
     |> Map.take([:id, :args, :queue, :worker, :attempt, :max_attempts, :tags])
     |> Map.put(:job, job)
     |> Map.put(:prefix, conf.prefix)
+    |> Map.put(:name, conf.name)
```
sorentwo, line 254: After constantly playing catch-up with exposing various job fields I eventually put the `job` in there directly. The telemetry events happen in the same process, so there isn't any data copying. 

What if we do the same thing with the config? We can put the config in and soft deprecate `prefix` by removing it from the telemetry docs.

sorentwo (changes requested):
Thanks for opening this! I'm so pleased to see the new PromEx dashboard coming together and the changes here look great. Just a couple of suggestions and this is good to go 👍

lib/oban/queue/executor.ex
akoutmos: That sounds good. Updated the code and the tests to reflect that change :+1:

sorentwo (approved):

sorentwo:
This is excellent, thanks @akoutmos 💛

PR #394, opened 2021-01-02
Added config to producer & circuit metrics metadata

akoutmos:
Continuing my work on the PromEx plugin, I came across a situation where I was unable to create accurate metrics for produced jobs when I had multiple Oban supervisors running since some of the queue names overlap, but I don't have the configuration available to determine which Oban supervisor this producer event was for.

This PR adds the config value to the producer event :).

I also added it to the circuit breaker events as it is nice to have that metric label there as well for reporting.


sorentwo:
@akoutmos This is great for consistency, I'm glad you added it 💛

PR #400, opened 2021-01-19
Adding telemetry events to the plugins

akoutmos:
Currently the PR only has telemetry events for the Cron plugin, but the Stager and Pruner changes are under way.

### DIFF
--- a/lib/oban/plugins/cron.ex
+++ b/lib/oban/plugins/cron.ex
@@ -29,6 +29,21 @@ defmodule Oban.Plugins.Cron do
 
   [tzdata]: https://hexdocs.pm/tzdata
   [perjob]: oban.html#module-periodic-jobs
+
+  ## Instrumenting with Telemetry
+  Oban emits the following telemetry event whenever the Cron plugin executes:
+
+  * `[:oban, :plugin, :cron, :start]` — at the point that the `Cron` plugin begins evaluating what cron jobs to enqueue
+  * `[:oban, :plugin, :cron, :stop]` — after the `Cron` plugin has enqueued any cron jobs
+  * `[:oban, :plugin, :cron, :exception]` — after the `Cron` plugin fails to insert cron jobs
+
+  The following chart shows which metadata you can expect for each event:
+
+  | event        | measures       | metadata                                     |
+  | ------------ | ---------------| ---------------------------------------------|
+  | `:start`     | `:system_time` | `:config, :interval, :name, :crontab`        |
+  | `:stop`      | `:duration`    | `:jobs, :config, :interval, :name, :crontab` |
+  | `:exception` | `:duration`    | `:error, :kind, :error, :stacktrace`         |
   """
 
   use GenServer
@@ -107,9 +122,32 @@ defmodule Oban.Plugins.Cron do
   def handle_info(:evaluate, %State{} = state) do
     state = schedule_evaluate(state)
 
-    Query.with_xact_lock(state.conf, state.lock_key, fn ->
-      insert_jobs(state.conf, state.crontab, state.timezone)
-    end)
+    start_metadata = %{
+      config: state.conf,
+      crontab: state.crontab,
+      name: state.name,
+      interval: state.interval
+    }
+
+    :telemetry.span(
+      [:oban, :plugin, :cron],
+      start_metadata,
+      fn ->
+        state.conf
+        |> Query.with_xact_lock(state.lock_key, fn ->
+          insert_jobs(state.conf, state.crontab, state.timezone)
+        end)
+        |> case do
+          {:ok, inserted_jobs} when is_list(inserted_jobs) ->
+            metadata = %{jobs: extract_jobs_from_insert(inserted_jobs)}
+            {:ok, Map.merge(start_metadata, metadata)}
+
+          error ->
+            metadata = %{error: error}
+            {:error, Map.merge(start_metadata, metadata)}
+        end
+      end
+    )
 
     {:noreply, state}
   end
@@ -123,6 +161,12 @@ defmodule Oban.Plugins.Cron do
   end
 
   # Parsing & Validation Helpers
+
+  defp extract_jobs_from_insert(jobs) do
+    Enum.map(jobs, fn {_, job} ->
+      job
+    end)
+  end
 
   defp parse_crontab(%State{crontab: crontab} = state) do
     parsed =
--- a/lib/oban/plugins/cron.ex
+++ b/lib/oban/plugins/cron.ex
@@ -29,6 +29,12 @@ defmodule Oban.Plugins.Cron do
 
   [tzdata]: https://hexdocs.pm/tzdata
   [perjob]: oban.html#module-periodic-jobs
+
+  ## Instrumenting with Telemetry
+
+  The `Oban.Plugins.Cron` plugin adds the following metadata to the `[:oban, :plugin, :stop]` event:
+
+  * :jobs - a list of jobs that were inserted into the database for processes
   """
 
   use GenServer
--- a/lib/oban/plugins/cron.ex
+++ b/lib/oban/plugins/cron.ex
@@ -128,19 +128,18 @@ defmodule Oban.Plugins.Cron do
 
     start_metadata = %{config: state.conf, plugin: __MODULE__}
 
-    :telemetry.span(
-      [:oban, :plugin],
-      start_metadata,
-      fn ->
-        case lock_and_insert_jobs(state) do
-          {:ok, inserted_jobs} when is_list(inserted_jobs) ->
-            {:ok, Map.put(start_metadata, :jobs, inserted_jobs)}
+    :telemetry.span([:oban, :plugin], start_metadata, fn ->
+      case lock_and_insert_jobs(state) do
+        {:ok, inserted_jobs} when is_list(inserted_jobs) ->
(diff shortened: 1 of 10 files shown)


lib/oban/plugins/cron.ex
```diff
@@ -29,6 +29,21 @@ defmodule Oban.Plugins.Cron do
 
   [tzdata]: https://hexdocs.pm/tzdata
   [perjob]: oban.html#module-periodic-jobs
+
+  ## Instrumenting with Telemetry
+  Oban emits the following telemetry event whenever the Cron plugin executes:
+
+  * `[:oban, :plugin, :cron, :start]` — at the point that the `Cron` plugin begins evaluating what cron jobs to enqueue
```
sorentwo, line 36: After adding instrumentation for four plugins I arrived at the conclusion that I really preferred a single plugin event that was differentiated by metadata. In this case there would be a single `[:oban, :plugin, :start]` event that included a `:plugin` in the metadata.

The Pro telemetry already uses this pattern (https://hexdocs.pm/oban/lifeline.html#instrumenting-with-telemetry). What do you think?

lib/oban/plugins/cron.ex
```diff
@@ -107,9 +122,32 @@ defmodule Oban.Plugins.Cron do
...
+      [:oban, :plugin, :cron],
+      start_metadata,
+      fn ->
+        state.conf
+        |> Query.with_xact_lock(state.lock_key, fn ->
+          insert_jobs(state.conf, state.crontab, state.timezone)
+        end)
+        |> case do
+          {:ok, inserted_jobs} when is_list(inserted_jobs) ->
+            metadata = %{jobs: extract_jobs_from_insert(inserted_jobs)}
```
sorentwo, line 142: The other plugins (pruner, stager) don't load the jobs. For the pruner that is impossible and for the stager it's kind of wasteful. They don't need to match necessarily, but what do you think about having the complete jobs here and only counts for pruning/staging?

lib/oban/plugins/cron.ex
```diff
@@ -107,9 +122,32 @@ defmodule Oban.Plugins.Cron do
...
     state = schedule_evaluate(state)
 
-    Query.with_xact_lock(state.conf, state.lock_key, fn ->
-      insert_jobs(state.conf, state.crontab, state.timezone)
-    end)
+    start_metadata = %{
+      config: state.conf,
+      crontab: state.crontab,
+      name: state.name,
+      interval: state.interval
```
sorentwo, line 129: The `interval` is only configurable for testing and the module should only be looked up through the registry. I don't think the `name` or `interval` should be included, and the `crontab` is kind of questionable as well.

lib/oban/plugins/cron.ex
akoutmos: That sounds good to me!

Just so that I am consistent, is the `:plugin` value the module name (i.e `__MODULE__`) or is it a downcased atom `:cron`?

lib/oban/plugins/cron.ex
akoutmos: That's actually the approach I was going to take hah. Great minds think alike :P.

lib/oban/plugins/cron.ex
akoutmos: Sounds good. I wasn't sure if that information was useful or not. I'll remove the 3 you mentioned and add the `:plugin` field instead.

lib/oban/plugins/cron.ex
```diff
@@ -107,9 +126,27 @@ defmodule Oban.Plugins.Cron do
...
+      [:oban, :plugin],
+      start_metadata,
+      fn ->
+        state.conf
+        |> Query.with_xact_lock(state.lock_key, fn ->
+          insert_jobs(state.conf, state.crontab, state.timezone)
+        end)
+        |> case do
+          {:ok, inserted_jobs} when is_list(inserted_jobs) ->
+            metadata = %{jobs: extract_jobs_from_insert(inserted_jobs)}
```
sorentwo, line 141: You can change the return value of `insert_jobs` from a list of ok tuples to a list of jobs and avoid the separate extraction step.

lib/oban/plugins/cron.ex
```diff
@@ -107,9 +126,27 @@ defmodule Oban.Plugins.Cron do
...
+            {:ok, Map.merge(start_metadata, metadata)}
+
+          error ->
+            metadata = %{error: error}
+            {:error, Map.merge(start_metadata, metadata)}
```
sorentwo, line 146:
```suggestion
          {:ok, inserted_jobs} when is_list(inserted_jobs) ->
            {:ok, Map.put(start_metadata, :jobs, inserted_jobs)}

          error ->
            {:error, Map.put(start_metadata, :error, error)}
```

lib/oban/plugins/pruner.ex
```diff
@@ -78,9 +97,27 @@ defmodule Oban.Plugins.Pruner do
...
+        |> case do
+          {:ok, {jobs_pruned_count, _}} ->
+            metadata = %{jobs_pruned_count: jobs_pruned_count}
+            {:ok, Map.merge(start_metadata, metadata)}
+
+          error ->
+            metadata = %{error: error}
+            {:error, Map.merge(start_metadata, metadata)}
+        end
+      end
```
sorentwo, line 119: This will return `false` if another process has the advisory lock already. That's an expected situation and shouldn't fall through to the error case. The nested functions and `case` piping seems a little awkward to me, perhaps we can make a helper function to `lock_and_delete_jobs` in this case?

```suggestion
      fn ->
        case lock_and_delete_jobs(state) do
          {:ok, {count, _}} ->
            {:ok, Map.put(start_metadata, :pruned_count, count)}

          false ->
            {:ok, Map.put(start_metadata, :pruned_count, 0)}

          error ->
            {:error, Map.merge(start_metadata, :error, error)}
        end
      end
```

lib/oban/query.ex
```diff
@@ -91,7 +91,7 @@ defmodule Oban.Query do
 
     case Repo.update_all(conf, query, set: [state: "available"]) do
       {0, _} -> []
```
sorentwo, line 93: This can also return a count tuple, `{0, []}` to prevent special casing.

lib/oban/plugins/cron.ex
```diff
@@ -107,9 +126,21 @@ defmodule Oban.Plugins.Cron do
...
+      fn ->
+        case lock_and_insert_jobs(state) do
+          {:ok, inserted_jobs} when is_list(inserted_jobs) ->
+            {:ok, Map.put(start_metadata, :jobs, inserted_jobs)}
+
+          error ->
+            {:error, Map.put(start_metadata, :error, error)}
+        end
+      end
+    )
```
sorentwo, line 143: This has the same `{:ok, false}` response due to the use of xact lock. I think we can save an extra three lines of vertical space rather cleanly:

```suggestion
    :telemetry.span([:oban, :plugin], start_metadata, fn ->
      case lock_and_insert_jobs(state) do
        {:ok, inserted_jobs} when is_list(inserted_jobs) ->
          {:ok, Map.put(start_metadata, :jobs, inserted_jobs)}

        {:ok, false} when is_list(inserted_jobs) ->
          {:ok, Map.put(start_metadata, :jobs, [])}

        error ->
          {:error, Map.put(start_metadata, :error, error)}
      end
    end)
```

lib/oban/plugins/pruner.ex
```diff
@@ -78,15 +97,36 @@ defmodule Oban.Plugins.Pruner do
...
   def handle_info(:prune, %State{} = state) do
-    Query.with_xact_lock(state.conf, state.lock_key, fn ->
-      delete_jobs(state.conf, state.max_age, state.limit)
-    end)
+    start_metadata = %{config: state.conf, plugin: __MODULE__}
+
+    :telemetry.span(
+      [:oban, :plugin],
+      start_metadata,
+      fn ->
```
sorentwo, line 105: The function looks perfect now. Can we collapse the vertical space, as suggested above?

lib/oban/plugins/stager.ex
```diff
@@ -55,18 +74,38 @@ defmodule Oban.Plugins.Stager do
 
   @impl GenServer
   def handle_info(:stage, %State{} = state) do
+    start_metadata = %{config: state.conf, plugin: __MODULE__}
+
+    :telemetry.span(
```
sorentwo, line 79: Same comment here about the vertical space. It also needs to handle the `{:ok, false}` xact lock return value.

lib/oban/query.ex
```diff
@@ -90,8 +90,8 @@ defmodule Oban.Query do
       |> select([j], j.queue)
 
     case Repo.update_all(conf, query, set: [state: "available"]) do
-      {0, _} -> []
-      {_, queues} -> queues
+      {0, []} -> []
```
sorentwo, line 93: Doesn't this need to return `{0, []}`?

test/support/plugin_telemetry_handler.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Oban.PluginTelemetryHandler do
```
sorentwo, line 1: Nice extraction 👍

sorentwo (approved):
Beautiful!

sorentwo:
Thanks, this is perfect 💛

PR #406, opened 2021-01-24
Fix typo in "mix ci" alias

preciz:
I think this command was silently failing with exit code 0 due to this typo.

Thank you for reviewing.


sorentwo:
Nice catch. Thanks! 💛

PR #410, opened 2021-01-25
Support Oban.insert_all/4 with a changes function

halostatue:
- Ecto.Multi supports `insert_all` where the changeset list can be provided as a function/1 that reads from the changes. This adds the same to `Oban.insert_all/4`.

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -343,7 +343,7 @@ defmodule Oban do
 
   ## Example
 
-      changesets = Enum.map(0..100, MyApp.Worker.new(%{id: &1}))
+      changesets = Enum.map(0..100, &MyApp.Worker.new(%{id: &1}))
 
       Ecto.Multi.new()
       |> Oban.insert_all(:jobs, changesets)
@@ -354,7 +354,8 @@ defmodule Oban do
           name(),
           multi :: Multi.t(),
           multi_name :: Multi.name(),
-          changesets_or_wrapper :: [job_changeset()] | %{changesets: [job_changeset()]}
+          changesets_or_wrapper ::
+            [job_changeset()] | %{changesets: [job_changeset()]} | (map -> [job_changeset()])
         ) :: Multi.t()
   def insert_all(name \\ __MODULE__, multi, multi_name, changesets_or_wrapper)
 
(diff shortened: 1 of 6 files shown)


halostatue:
The implementation here is based on digging through `Ecto.Multi.insert_all`.

sorentwo:
@halostatue Thanks for the PR! It hit in the middle of a release (and some follow-up fixes), it hasn't been missed.

v0idpwn (approved):

halostatue:
Not a problem. If I have a problem it’s that I upgraded to 2.3 one day before 2.4 was released. :D

lib/oban.ex
```diff
@@ -354,15 +354,17 @@ defmodule Oban do
           name(),
           multi :: Multi.t(),
           multi_name :: Multi.name(),
-          changesets_or_wrapper :: [job_changeset()] | %{changesets: [job_changeset()]}
+          changesets_or_wrapper ::
+            [job_changeset()] | %{changesets: [job_changeset()]} | (map -> [job_changeset()])
```
sorentwo, line 358: This typespec is getting a little out of hand. What if we pull `changesets_or_wrapper` into a type and reference it here instead?

lib/oban/query.ex
```diff
@@ -71,13 +71,24 @@ defmodule Oban.Query do
     end
   end
 
-  @spec insert_all_jobs(Config.t(), Multi.t(), Multi.name(), [Changeset.t()]) :: Multi.t()
+  @spec insert_all_jobs(
+          Config.t(),
+          Multi.t(),
+          Multi.name(),
+          [Changeset.t()] | (map -> [Changeset.t()])
```
sorentwo, line 78: Likewise here. Can we pull this out as a type and keep the spec more compact?

sorentwo (changes requested):
This is a welcome addition! A couple of nitpick changes requested here, mostly to clean up typespecs and no logical changes. Ready to go once those changes are in 👍

sorentwo:
@halostatue Ping. If you don't have time to wrap this up I'll make the requested changes locally.

halostatue:
Sorry, I missed the original comments somehow (didn’t see the notifications). I won’t be able to look at it before tonight, so if you want to make the changes please feel free to do so (I’ve allowed edits by maintainers).

sorentwo:
No worries, I'm not in a rush. Whenever you can get to it is fine by me 👍

lib/oban/query.ex
halostatue: I’m not sure that what I’ve done is more compact, but I think it’s _better_ for reading.

halostatue (commented):
The changes are a little larger than previously done because of type extraction, but I think that the types are a little more noticeable than before. I’m running dialyzer to make sure that I have not introduced anything new, but I will be pushing up the changes to address these (and rebase against `master`) soon.

sorentwo (approved):
Nice type cleanup! They are much clearer now 🥇

PR #411, opened 2021-01-26
Align migration module with command in changelog

maartenvanvliet:
The command run was `mix ecto.gen.migration upgrade_oban_jobs_to_v10` which would generate `Repo.Migrations.UpdateObanJobsToV10`


sorentwo:
Nice catch, thank you! 💛

PR #441, opened 2021-02-26
Remove obsolete performance note

### LINKED ISSUE #440: README - Is performance note regarding index on args still valid?
Hi! I think I've found an outdated part of README:

### Environment

- Oban Version - 2.4.3

### Current Behavior

The [README](https://github.com/sorentwo/oban/tree/master#performance-note) says

> Performance Note
>If your application makes heavy use of unique jobs you may want to add an index on the args column of the oban_jobs table. The other columns considered for uniqueness are already covered by indexes.

However it seems migration v10 added that index
https://github.com/sorentwo/oban/blob/219375bb159b83e761e0394d9ac59813f2cd8dba/lib/oban/migrations/v10.ex#L39

Doesn't that render the `Performance Note` obsolete? 

### Expected Behavior

Performance Note removed from readme

bblaszkow06:
Removes obsolete part of README

Closes #440

### DIFF
--- a/README.md
+++ b/README.md
@@ -557,12 +557,6 @@ they _do not_ rely on unique constraints in the database. This makes uniqueness
 entirely configurable by application code, without the need for database
 migrations.
 
-#### Performance Note
-
-If your application makes heavy use of unique jobs you may want to add an index
-on the `args` column of the `oban_jobs` table. The other columns considered for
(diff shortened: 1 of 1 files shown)


tomekzaw (approved):

sorentwo:
Thanks 💛

PR #445, opened 2021-03-08
Deal with time overflowing into the morning just before midnight.

LostKobrakai:


sorentwo (changes requested):
Nice fix!

Before merging this I'd like to enhance the property test so it doesn't allow a regression. With the following modification:

```elixir
      check all hour <- integer(0..23),
                minute <- integer(0..59),
                second <- integer(0..59), max_runs: 1000 do
        {:ok, time} = Time.new(hour, minute, second)

        assert Cron.interval_to_next_minute(time) in 1_000..60_000
      end
```

It catches the edge case that we'd expect:

```
  1) property interval_to_next_minute/1 calculated time is always within a short future range (Oban.Plugins.CronTest)
     test/oban/plugins/cron_test.exs:59
     Failed with generated values (after 223 successful runs):

         * Clause:    hour <- integer(0..23)
           Generated: 23

         * Clause:    minute <- integer(0..59)
           Generated: 59

         * Clause:    second <- integer(0..59)
           Generated: 0

     Assertion with in failed
     code:  assert Cron.interval_to_next_minute(time) in 1000..60000
     left:  -86340000
     right: 1000..60000
```

Setting `max_runs` to 1000 is a bit excessive, but I'd like to have the other property test changes included with this patch.

sorentwo:
Merged locally and closed with 95d16ab. Thanks 💛

PR #451, opened 2021-03-24
README example: Blow up if can't fetch Oban config

nathanl:
With `get_env`, if you copypasta and leave it as `:my_app`, the error is more obscure


sorentwo:
Nice improvement. Thanks! 💛

PR #464, opened 2021-04-15
Allow replacing the scheduled execution time of a unique job

monteithpj:
This commit adds the capability to update the scheduled execution time of a unique job, similar to `:replace_args`

Not sure whether this is the nicest way of going about it (if there are other fields that could be replaced in future, this API gets a bit messy) but just went with something very simple to spark discussion.

Any feedback gratefully received 😅

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -36,6 +36,17 @@ defmodule Oban.Job do
           | {:period, unique_period()}
           | {:states, [unique_state()]}
 
+  @type replace_option :: [
+          :args
+          | :max_attempts
+          | :meta
+          | :priority
+          | :queue
+          | :scheduled_at
+          | :tags
+          | :worker
+        ]
+
   @type option ::
           {:args, args()}
           | {:max_attempts, pos_integer()}
@@ -44,7 +55,7 @@ defmodule Oban.Job do
           | {:queue, atom() | binary()}
           | {:schedule_in, pos_integer()}
           | {:replace_args, boolean()}
-          | {:replace_scheduled_at, boolean()}
+          | {:replace, [replace_option()]}
           | {:scheduled_at, DateTime.t()}
           | {:tags, tags()}
           | {:unique, [unique_option()]}
@@ -99,8 +110,7 @@ defmodule Oban.Job do
     field :scheduled_at, :utc_datetime_usec
 
     field :conf, :map, virtual: true
-    field :replace_args, :boolean, virtual: true
-    field :replace_scheduled_at, :boolean, virtual: true
+    field :replace, {:array, :any}, virtual: true
     field :unique, :map, virtual: true
     field :unsaved_error, :map, virtual: true
   end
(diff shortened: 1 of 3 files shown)


sorentwo:
Thanks for opening the PR. This is an interesting use case and it prompts the question: what other things should be replaceable? What about `meta`, or `tags`, or `priority`, etc? I think the API needs to be flexible to support anything that is normally set in `Worker.new`, otherwise we'll slowly create toward a slough of `replace_` keys. WDYT?

monteithpj:
Yeah, that's what I was thinking. I can update to create something like:
```Worker.new(args, replace: [:args, scheduled_at])```

Would all the options other than `:unique`, `:replace_args` and `schedule_in` would make sense to be replaceable?

**Comment by sorentwo** [comment]
That replace syntax looks good!

> Would all the options other than :unique, :replace_args and schedule_in would make sense to be replaceable?

Yes, exactly. We can also soft deprecate the `replace_args` variant and translate that into `replace: [:args]` internally.

File: lib/oban/job.ex [hunk 1]
```diff
@@ -36,6 +36,17 @@ defmodule Oban.Job do
...
 
+  @type replace_option :: [
+          :args
+          | :max_attempts
+          | :meta
+          | :priority
+          | :queue
+          | :scheduled_at
+          | :tags
+          | :worker
```
sorentwo, line 47: Technically the `worker` and the `queue` are part of the unique constraints and viable viable options to `Job.new`. it seems strange to change them on a conflict, but I can't think of a reason why _not_ to do it.

lib/oban/job.ex
```diff
@@ -325,6 +338,28 @@ defmodule Oban.Job do
...
+
+        put_change(changeset, :replace, updated)
+
+      nil ->
+        changeset
+
+      _ ->
+        add_error(changeset, :replace_args, "invalid value")
+    end
+  end
```
sorentwo, line 361: I believe we can simplify this with `update_change`:

```suggestion
  defp put_replace(changeset, value) do
    case value do
      true -> update_change(changeset, :replace, &[:args | &1])
      false -> update_change(changeset, :replace, &(&1 -- [:args]))
      nil -> changeset
      _ -> add_error(changeset, :replace_args, "invalid value")
    end
  end
```

sorentwo (commented):
This looks great, nicely done! One small suggestion that will simplify the `replace_args` support, otherwise this is good to go.

lib/oban/job.ex
monteithpj: I don't think this will work in the case where `:replace` isn't set, as `update_change` only gets applied when the key is set, unless I'm missing something?

monteithpj: Good point! Will tidy this up a bit

sorentwo (approved):

sorentwo:
This is great! Thanks for working with me on it 💛

PR #475, opened 2021-05-01
Make notifier configurable

### LINKED ISSUE #467: Abstract notifier to allow alternate implementations

Not all databases can make use of `LISTEN/NOTIFY`, typically due to the use of a pooler like pg_bouncer in transaction mode. In that situation it is impossible to use the Gossip plugin, which prevents any use of the dashboard.

Add a notifier behavior with swappable callback modules. The default will remain PostgreSQL based, but could be swapped for something like pg (process groups), redis, phoenix pubsub, etc.

v0idpwn:
Closes #467. 

Opening as a draft because I have some questions:

a) It seems to me that, if the `Oban.Notifier` API is being used outside, my implementation introduce a compatibility break. A solution would be adding clauses that don't depend on `config` on `Oban.Notifier`, these would call `Oban.PostgresNotifier`. WDYT?

b) Is there anything currently on `PostgresNotifier` that should be moved to `Notifier`? I kept the base notifier as thin as possible

P.S.: I'll reinsert all the documentation reviewed for these changes once we agree on the API.

### DIFF
--- a/lib/oban/notifier.ex
+++ b/lib/oban/notifier.ex
@@ -62,9 +62,16 @@ defmodule Oban.Notifier do
   @type option :: {:name, module()} | {:conf, Config.t()}
   @type channel :: :gossip | :insert | :signal
 
+  @doc "Starts a notifier"
   @callback start_link([option]) :: GenServer.on_start()
+
+  @doc "Register current process to receive messages from some channels"
   @callback listen(server(), channels :: list(channel())) :: :ok
+
+  @doc "Unregister current process from channels"
   @callback unlisten(server(), channels :: list(channel())) :: :ok
+
+  @doc "Broadcast a notification in a channel"
   @callback notify(server(), channel :: channel(), payload :: map() | [map()]) :: :ok
 
   @mappings %{
--- a/lib/oban/notifier.ex
+++ b/lib/oban/notifier.ex
@@ -56,9 +56,22 @@ defmodule Oban.Notifier do
     |> conf.notifier.unlisten(channels)
   end
 
-  @doc false
-  @spec notify(Config.t(), channel :: channel(), payload :: map() | [map()]) :: :ok
+  @doc """
+  Sends a notification to a channel
+
+  Using notify/3 with a config is soft deprecated. Use a server as the first
+  argument instead
+  """
+  @spec notify(Config.t() | server(), channel :: channel(), payload :: map() | [map()]) :: :ok
   def notify(%Config{} = conf, channel, payload) when is_channel(channel) do
+    conf.name
+    |> Registry.whereis(Oban.Notifier)
+    |> conf.notifier.notify(channel, payload)
+  end
+
+  def notify(server, channel, payload) when is_channel(channel) do
+    conf = Oban.config(server)
+
     conf.name
     |> Registry.whereis(Oban.Notifier)
     |> conf.notifier.notify(channel, payload)
--- a/lib/oban/notifier.ex
+++ b/lib/oban/notifier.ex
@@ -1,6 +1,59 @@
 defmodule Oban.Notifier do
   @moduledoc """
-  Behaviour for notifiers
+  The `Notifier` coordinates listening for and publishing notifications for events in predefined
+  channels.
+
+  Every Oban supervision tree contains a notifier process, registered as `Oban.Notifier`, which
+  can be any implementation of the `Oban.Notifier` behaviour. The default one is 
+  `Oban.PostgresNotifier`, which relies on Postgres `LISTEN/NOTIFY`. All incoming notifications
+  are relayed through the notifier to other processes.
+
+  ## Channels
+
+  The notifier recognizes three predefined channels, each with a distinct responsibility:
+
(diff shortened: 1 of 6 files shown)


lib/oban/postgres_notifier.ex
```diff
@@ -0,0 +1,201 @@
...
+  end
+
+  def listen(oban_name, [_ | _] = channels) do
+    oban_name
+    |> Registry.whereis(Oban.Notifier)
+    |> listen(channels)
+  end
+
+  @impl Oban.Notifier
+  def unlisten(oban_name \\ Oban, [_ | _] = channels) do
```
sorentwo, line 62: Similarly to `listen/2`, we can simplify this to be `unlisten(server, channels)`.

lib/oban/postgres_notifier.ex
```diff
@@ -0,0 +1,201 @@
...
+    channel = "#{conf.prefix}.#{@mappings[channel]}"
+
+    Repo.query(
+      conf,
+      "SELECT pg_notify($1, payload) FROM json_array_elements_text($2::json) AS payload",
+      [channel, Enum.map(payload, &Jason.encode!/1)]
+    )
+
+    :ok
+  end
```
sorentwo, line 83: For the sake of consistency it would be nice to make this `notify(server, channel, payload)`. Other implementations may rely on server state. We can fetch the `conf` from the local server state with something fast like this:

```elixir
with %State{conf: conf} <- GenServer.call(server, :get_state) do
  # use the conf in here
end
```

That pulls the server state into the local process and avoids bottlenecking outgoing notifications.

lib/oban/postgres_notifier.ex
```diff
@@ -0,0 +1,201 @@
...
+
+  # Helpers
+
+  for {atom, name} <- @mappings do
+    defp reverse_channel(unquote(name)), do: unquote(atom)
+  end
+
+  defp validate_channels!([]), do: :ok
+  defp validate_channels!([head | tail]) when is_channel(head), do: validate_channels!(tail)
+  defp validate_channels!([head | _]), do: raise(ArgumentError, "unexpected channel: #{head}")
```
sorentwo, line 178: These validations should probably live in the `Notifier` module. Notification channels are either valid or invalid for all implementations.

sorentwo (commented):
This is a wonderful start 👍 . The config changes look spot on, and the naming is great. I'd like to get a more consistent behaviour and push the server resolution into the `Oban.Notifier` module itself. That should work for a wider range of adapters.

lib/oban/postgres_notifier.ex
v0idpwn: Do you think we should do this for every function in the notifier? Seems like a more consistent approach. I think it would also be 100% compatible with the old implementation.

sorentwo:
Ideally each of the functions would be consistent. The API stays compatible and the servers are still testable in isolation.

lib/oban/notifier.ex
```diff
@@ -84,232 +22,57 @@ defmodule Oban.Notifier do
...
-  end
-
-  def handle_info(:reset_circuit, %State{circuit: :disabled} = state) do
-    state =
-      state
-      |> open_circuit()
-      |> connect_and_listen()
-
-    {:noreply, state}
+  def notify(%Config{} = conf, channel, payload) when is_channel(channel) do
```
v0idpwn, line 61: Removing this config would introduce a compatibility break. Shall I?

sorentwo: We can soft-deprecate the old version and transparently map it to the new one.

v0idpwn:
Anything else we should change in here? :)

lib/oban/postgres_notifier.ex
```diff
@@ -0,0 +1,176 @@
+defmodule Oban.PostgresNotifier do
+  @doc """
+  Postgres Listen/Notify based Notifier
```
sorentwo, line 3: This should contain the old caveats section. It's still the default notifier most apps will use.

lib/oban/notifier.ex
```diff
@@ -66,16 +56,17 @@ defmodule Oban.Notifier do
...
-  alias Oban.{Config, Registry, Repo}
-  alias Postgrex.Notifications
+  alias Oban.{Config, Registry}
 
+  @type server :: GenServer.server()
   @type option :: {:name, module()} | {:conf, Config.t()}
   @type channel :: :gossip | :insert | :signal
 
+  @callback start_link([option]) :: GenServer.on_start()
+  @callback listen(server(), channels :: list(channel())) :: :ok
```
sorentwo, line 66: These callbacks populate the docs for the notifier implementations. They need to describe what the function is doing a little, at least as a stub.

sorentwo (commented):
The structure looks perfect. There are still a couple of docs to tie things together and then this is good to go 👍

lib/oban/postgres_notifier.ex
v0idpwn: True! Removed from one place and forgot to add in the other

sorentwo:
@v0idpwn Wonderful! Thank you for all the iterations 💛

PR #480, opened 2021-05-07
Fix dialyzer warning related to less-specific changeset type

jeroenvisser101:
This ensures that dialyzer no longer complains about "insert_job!/2 will never return"

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -200,7 +200,7 @@ defmodule Oban.Job do
       |> Oban.insert()
   """
   @doc since: "0.1.0"
-  @spec new(args(), [option]) :: Ecto.Changeset.t()
+  @spec new(args(), [option()]) :: changeset()
   def new(args, opts \\ []) when is_map(args) and is_list(opts) do
     params =
       opts


lib/oban/job.ex
```diff
@@ -200,7 +200,7 @@ defmodule Oban.Job do
       |> Oban.insert()
   """
   @doc since: "0.1.0"
-  @spec new(args(), [option]) :: Ecto.Changeset.t()
+  @spec new(args(), [option]) :: changeset
```
sorentwo, line 203: Purely for the sake of consistency:

```suggestion
  @spec new(args(), [option()]) :: changeset()
```

jeroenvisser101: Sure! I kept it like is because in some other places it's still like that. No problem at all :)

jeroenvisser101:
@sorentwo I think the same issue exists in `Oban.Pro.Relay`

sorentwo (approved):

sorentwo:
@jeroenvisser101 Thanks for the fix (and it's great to hear that somebody is using the Relay plugin 😉).

PR #482, opened 2021-05-15
fix a grammatical error in Oban.Config module doc

sushant12:


lib/oban/config.ex
```diff
@@ -3,7 +3,7 @@ defmodule Oban.Config do
   The Config struct validates and encapsulates Oban instance state.
 
   Options passed to `Oban.start_link/1` are validated and stored in a config struct. Internal
-  modules and plugins are always passed the config with a `:conf` key.
+  modules and plugins are always passed in the config with a `:conf` key.
```
v0idpwn, line 6: I don't think it is an actual grammatical error. It seems to me that "modules are passed the config" and not that "modules are passed in the config".

This happens, for instance, here: https://github.com/sorentwo/oban/blob/d5cc908ef440dbe8532bb462db02ba7ee72c7b5f/lib/oban/midwife.ex#L26

sushant12: Oh ok I see now. So internal modules like the Notifier and Midwife are passed the configuration as `:conf`

```
children = [
      {Notifier, conf: conf, name: Registry.via(conf.name, Notifier)},
      {Midwife, conf: conf, name: Registry.via(conf.name, Midwife)}
    ]
```

Issue #483, opened 2021-05-17
Optionally return an error tuple when inserting a duplicate job


sorentwo:
### Is your feature request related to a problem? Please describe.

Users frequently ask how to tell if a unique job was inserted or if its a job that already existed in the database. Calls to `Oban.insert` will always return `{:ok, job}` whether it inserted or not.

### Describe the Solution You'd Like

Add an option to `unique` that allows the end user to control whether they get an `{:ok, job}` tuple or an error tuple indicating that the job already exists. Putting the option into `unique` opts is backward compatible and easily detected at the insertion point.

### Describe Alternatives You've Considered

Directing users to use other values, or the `replace` syntax to change things when they are updated.

chasers:
Thanks for this ... yes for me understanding explicitly if a job was added or not is useful and returning an error tuple if a job was not added felt better.

After digging, it looks like `inserted_at` is `nil` if a job is new so I have that for now!

chasers:
Mentioned by someone else is Slack ... returning `{:duplicate, job}` which I also like better.

capitalist:
>After digging, it looks like inserted_at is nil if a job is new so I have that for now!
This is a useful workaround, thanks for adding it here.

sorentwo:
Closing in favor of the PR. Thanks all!

PR #484, opened 2021-05-18
Check when inserted unique job previously existed

v0idpwn:
Adds the `duplicate?` virtual field, which is true when an inserted job already existed in the database.

I don't think that this is a good name, as I wouldn't say that the `:replace` upsert scenario is a dupe.

Related to #483, but doesn't use a tuple as this would be weird with the bang variant of `Oban.insert`

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -112,6 +112,7 @@ defmodule Oban.Job do
     field :conf, :map, virtual: true
     field :replace, {:array, :any}, virtual: true
     field :unique, :map, virtual: true
+    field :conflict?, :boolean, virtual: true, default: false
     field :unsaved_error, :map, virtual: true
   end
 
--- a/lib/oban/query.ex
+++ b/lib/oban/query.ex
@@ -101,14 +101,18 @@ defmodule Oban.Query do
 
     with {:ok, query, lock_key} <- unique_query(changeset),
          :ok <- acquire_lock(conf, lock_key, query_opts),
-         {:ok, job} <- unprepared_one(conf, query, query_opts) do
-      return_or_replace(conf, query_opts, job, changeset)
+         {:ok, job} <- unprepared_one(conf, query, query_opts),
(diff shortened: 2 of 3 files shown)


lib/oban/job.ex
```diff
@@ -112,6 +112,7 @@ defmodule Oban.Job do
     field :conf, :map, virtual: true
     field :replace, {:array, :any}, virtual: true
     field :unique, :map, virtual: true
+    field :duplicate?, :boolean, virtual: true, default: false
```
sorentwo, line 115: How about `conflict?` instead? That is less ambiguous than `duplicate?` when combined with the `replace` option.

lib/oban/query.ex
```diff
@@ -101,14 +101,18 @@ defmodule Oban.Query do
 
     with {:ok, query, lock_key} <- unique_query(changeset),
          :ok <- acquire_lock(conf, lock_key, query_opts),
-         {:ok, job} <- unprepared_one(conf, query, query_opts) do
-      return_or_replace(conf, query_opts, job, changeset)
+         {:ok, job} <- unprepared_one(conf, query, query_opts),
+         {:ok, job} <- return_or_replace(conf, query_opts, job, changeset) do
+      {:ok, Map.put(job, :duplicate?, true)}
```
sorentwo, line 106: Since it is a virtual field we know it will be present. Let's use the struct update syntax instead of Map.put here.

sorentwo (commented):
This turned out to be so tiny! Let's nail down the name and this is all set.

v0idpwn:
Did the suggested fixes. Maybe this should be documented somewhere?

sorentwo:
> Maybe this should be documented somewhere?

Absolutely. The `Job.t` type needs to be updated, a note about the field in `Job.new`, and some info in the unique section of the README would be perfect.

sorentwo:
Merged locally. Thanks! 💛

Issue #466, opened 2021-04-19
Unify job execution during runtime and testing


sorentwo:
### Is your feature request related to a problem? Please describe.

There are differences between how a job is executed by `perform_job` and during runtime through the `Executor` module. Most notably, `perform_job` doesn't emit any telemetry events.

### Describe the Solution You'd Like

Use the `Executor` module to exercise `perform` functions.

### Describe Alternatives You've Considered

Accept that there are subtle differences and document them.

milmazz:
Hey :wave:  @sorentwo I just started looking at this issue, and it seems like an interesting problem. 

The current implementation of `perform_job/3` is like this:

https://github.com/sorentwo/oban/blob/d5cc908ef440dbe8532bb462db02ba7ee72c7b5f/lib/oban/testing.ex#L202-L220

So, if I got your proposal right, you're suggesting to change the way we create the `result` binding, using the `Executor` instead of doing a direct call to `worker.perform()`. But to use the `Executor.call/1` function we need to build first an `Executor` struct using for example `Executor.new/2`, where the first argument is an `Oban.Config` struct, so, my first question is, how do we capture that config at _runtime_ without introducing a _backward incompatilbility_?

Our new `result` could be something like this I guess:

```elixir
      result =
        [repo: MyRepo]
        |> Oban.Config.new()
        |> Executor.new(job)
        |> Executor.call()
```

But then, we need to capture the `Oban.Config` struct first. Any suggestion?

Also, I think we need to change the `changeset` to introduce the `attempted_at`, and `scheduled_at` to avoid issues here with `DateTime.diff/3`:

https://github.com/sorentwo/oban/blob/d5cc908ef440dbe8532bb462db02ba7ee72c7b5f/lib/oban/queue/executor.ex#L121-L127

Please let me know your opinion about this issue. Also, I don't know if someone else is working on this at the moment.

milmazz:
@sorentwo Thanks for your guidance, I'll start working based on your feedback. I'll report back soon<sup>TM</sup>

PR #486, opened 2021-05-21
feat: `perform_job/{2,3}` now emits telemetry events

milmazz:
Fixes: #466

### DIFF
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -219,24 +219,7 @@ defmodule Oban.Queue.Executor do
         %{exec | result: result, state: :snoozed, snooze: seconds}
 
       returned ->
-        Logger.warn(fn ->
-          """
-          Expected #{worker}.perform/1 to return:
-
-          - `:ok`
-          - `:discard`
-          - `{:ok, value}`
-          - `{:error, reason}`,
-          - `{:discard, reason}`
-          - `{:snooze, seconds}`
-
-          Instead received:
-
-          #{inspect(returned, pretty: true)}
-
-          The job will be considered a success.
-          """
-        end)
+        maybe_log_warning(exec, returned)
 
         %{exec | state: :success}
     end
@@ -298,5 +281,29 @@ defmodule Oban.Queue.Executor do
     unsaved_error = %{kind: exec.kind, reason: exec.error, stacktrace: exec.stacktrace}
 
     %{exec.job | unsaved_error: unsaved_error}
+  end
+
+  defp maybe_log_warning(exec, returned)
+  defp maybe_log_warning(%__MODULE__{safe: false}, _returned), do: :noop
+
+  defp maybe_log_warning(%__MODULE__{worker: worker}, returned) do
+    Logger.warn(fn ->
+      """
+      Expected #{worker}.perform/1 to return:
+
+      - `:ok`
+      - `:discard`
+      - `{:ok, value}`
+      - `{:error, reason}`,
+      - `{:discard, reason}`
+      - `{:snooze, seconds}`
+
+      Instead received:
+
+      #{inspect(returned, pretty: true)}
+
+      The job will be considered a success.
+      """
+    end)
   end
 end
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -128,30 +128,46 @@ defmodule Oban.Queue.Executor do
 
   @spec report_finished(t()) :: t()
   def report_finished(%__MODULE__{} = exec) do
+    exec
+    |> ack_event()
+    |> emit_event()
+  end
+
+  @spec ack_event(t()) :: t()
+  def ack_event(%__MODULE__{} = exec) do
     case exec.state do
       :success ->
         Engine.complete_job(exec.conf, exec.job)
-
-        execute_stop(exec)
+        exec
 
       :failure ->
         job = job_with_unsaved_error(exec)
 
         Engine.error_job(exec.conf, job, backoff(exec.worker, job))
 
-        execute_exception(%{exec | job: job})
+        %{exec | job: job}
 
       :snoozed ->
         Engine.snooze_job(exec.conf, exec.job, exec.snooze)
-
-        execute_stop(exec)
+        exec
 
       :discard ->
         job = job_with_unsaved_error(exec)
 
         Engine.discard_job(exec.conf, job)
 
-        execute_stop(%{exec | job: job})
+        %{exec | job: job}
+    end
+  end
+
+  @spec emit_event(t()) :: t()
+  def emit_event(%__MODULE__{} = exec) do
+    case exec.state do
+      :failure ->
+        execute_exception(exec)
+
+      state when state in [:success, :snoozed, :discard] ->
+        execute_stop(exec)
     end
 
     exec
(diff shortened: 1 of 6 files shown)


lib/oban/testing.ex
```diff
@@ -116,7 +116,13 @@ defmodule Oban.Testing do
...
 
   alias Ecto.Changeset
-  alias Oban.{Job, Repo, Worker}
+
+  alias Oban.{
+    Job,
+    Queue.Executor,
+    Repo,
+    Worker
+  }
```
sorentwo, line 125: It's purely a preference, but let's keep this on a single line:

```suggestion
  alias Oban.{Job, Queue.Executor, Repo, Worker}
```

lib/oban/queue/executor.ex
```diff
@@ -177,17 +193,32 @@ defmodule Oban.Queue.Executor do
...
+        %{
+          exec
+          | result: result,
+            state: :failure,
+            error: PerformError.exception({worker, {:error, reason}})
+        }
 
-      {:snooze, seconds} ->
-        %{exec | state: :snoozed, snooze: seconds}
+      {:snooze, seconds} = result when is_integer(seconds) and seconds > 0 ->
```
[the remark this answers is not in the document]
sorentwo: Another great catch from this change 👍

lib/oban/testing.ex
```diff
@@ -200,6 +208,7 @@ defmodule Oban.Testing do
   @spec perform_job(worker :: Worker.t(), args :: Job.args(), opts :: [Job.option()]) ::
           Worker.result()
   def perform_job(worker, args, opts \\ []) when is_atom(worker) and is_map(args) do
+    {repo, opts} = Keyword.pop(opts, :repo)
```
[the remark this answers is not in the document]
sorentwo: That should be a safe assumption. This function, like the others in here, is meant to be called locally. That's how the docs are written as well, so I think it's safe.

There isn't any need to have the `opts \\ []` default now. The function is always called as `perform_job/3` now.

lib/oban/testing.ex
```diff
@@ -515,4 +531,12 @@ defmodule Oban.Testing do
     |> where(^[{key, value}])
     |> apply_where_clauses(rest)
   end
+
+  defp create_job(changeset) do
+    now = DateTime.utc_now()
+
+    changeset
+    |> Changeset.apply_action!(:insert)
+    |> Map.merge(%{attempted_at: now, scheduled_at: DateTime.add(now, -3)})
```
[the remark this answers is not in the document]
sorentwo: They can both be the same timestamp. The `queue_time` will always be 0, which makes sense to me.

```suggestion
    |> Map.merge(%{attempted_at: now, scheduled_at: now})
```

test/integration/telemetry_test.exs
```diff
@@ -3,31 +3,17 @@ defmodule Oban.Integration.TelemetryTest do
 
   import ExUnit.CaptureLog
 
-  alias Oban.{Config, PerformError, Telemetry}
+  alias Oban.{
+    Config,
+    PerformError,
+    Telemetry,
+    TelemetryHandler
+  }
```
sorentwo, line 11: Same here. Let's compact these aliases:

```suggestion
  alias Oban.{Config, PerformError, Telemetry, TelemetryHandler}
```

test/oban/testing_test.exs
```diff
@@ -67,18 +71,46 @@ defmodule Oban.TestingTest do
...
+      end
     end
 
     test "returning the value of worker's perform/1 function" do
       assert :ok = perform_job(Worker, %{ref: 1, action: "OK"})
       assert :discard = perform_job(Worker, %{ref: 1, action: "DISCARD"})
       assert {:error, _} = perform_job(Worker, %{ref: 1, action: "ERROR"})
     end
+
+    test "validating that emits telemetry events" do
```
sorentwo, line 103:
```suggestion
    test "emitting appropriate telemetry events" do
```

test/oban/testing_test.exs
```diff
@@ -67,18 +71,46 @@ defmodule Oban.TestingTest do
...
+
+        #{inspect(result, pretty: true)}
+
+        The job will be considered a success.
+        """
+
+        assert capture_log(fn ->
+                 assert_perform_error(MisbehavedWorker, %{"action" => action}, message)
+               end) =~ log
+      end
```
sorentwo, line 94: This isn't ideal. The `perform_job` function will raise an error anyhow, and I wouldn't expect to see an error. Both `perform_job` and `drain_queue` use `safe: false`. Perhaps we raise an error rather than log a warning when `safe` is `false`?

sorentwo (changes requested):
Great first pass! Only a few items to fix up, and this is ready to go 👍

lib/oban/queue/executor.ex
```diff
@@ -128,32 +128,46 @@ defmodule Oban.Queue.Executor do
...
 
-        Engine.discard_job(exec.conf, job)
+    %{exec | job: job}
+  end
 
-        execute_stop(%{exec | job: job})
-    end
+  @spec emit_event(t()) :: t()
+  def emit_event(%__MODULE__{state: :failure} = exec) do
+    execute_exception(exec)
```
sorentwo, line 165: Let's pull the body of `execute_exception` inline. There's no need to put it in a separate private function now.

lib/oban/queue/executor.ex
```diff
@@ -128,32 +128,46 @@ defmodule Oban.Queue.Executor do
...
+    exec
+  end
 
+  def emit_event(%__MODULE__{state: state} = exec) when state in [:success, :snoozed, :discard] do
+    execute_stop(exec)
```
sorentwo, line 170: Same here.

lib/oban/queue/executor.ex
```diff
@@ -219,24 +219,7 @@ defmodule Oban.Queue.Executor do
...
-          - `{:snooze, seconds}`
-
-          Instead received:
-
-          #{inspect(returned, pretty: true)}
-
-          The job will be considered a success.
-          """
-        end)
+        maybe_log_warning(exec, returned)
```
sorentwo, line 222: Beautiful!

lib/oban/queue/executor.ex
```diff
@@ -219,24 +219,7 @@ defmodule Oban.Queue.Executor do
         %{exec | result: result, state: :snoozed, snooze: seconds}
 
       returned ->
-        Logger.warn(fn ->
```
milmazz, line 193: This `case` is long enough to extract this block into a separate helper function for readability purposes and handle the case when we don't want to log a warning because we're running on _unsafe_ mode. Please let me know if you're ok with it.

test/oban/queue/executor_test.exs
```diff
@@ -47,13 +44,6 @@ defmodule Oban.Queue.ExecutorTest do
                |> Executor.resolve_worker()
     end
 
-    test "warning on unexpected return values" do
-      message = capture_log(fn -> %{state: :success} = call_with_mode("warn") end)
-
-      assert message =~ "Expected #{__MODULE__}.Worker.perform/1"
-      assert message =~ "{:bad, :this_will_warn}"
-    end
```
milmazz, line 55: I think this doesn't make sense now that we don't log warnings when running on _unsafe_ mode.

lib/oban/queue/executor.ex
sorentwo: Totally ok with it 👍

test/oban/queue/executor_test.exs
sorentwo: Agreed 👍

test/oban/testing_test.exs
milmazz: > Perhaps we raise an error rather than log a warning when `safe` is `false`?

Yes, it's true, `perform_job/3` already raises an error if the `perform/1` returns an unexpected result, so, there is no point in adding more noise with the warning.

I just added a `maybe_log_warning/2` based on the `safe` field from the `Executor` struct :)

milmazz:
@sorentwo Should be ready for another review ;)

sorentwo (approved):
This looks perfect. Nicely done 💛

PR #488, opened 2021-05-21
Use `drop_if_exists` in 2.6 upgrade doc

bryannaegele:
This prevents errors if somebody runs `mix ecto.migrate` and then tries to run a test suite.


sorentwo:
Excellent! Thanks 💛

PR #489, opened 2021-05-21
feat: prune only some states

aschiavon91:
I don't know if it can be helpful for others, but I have been using this to avoid prune discarded jobs.


sorentwo:
The basic pruner is intentionally simple. The DynamicPruner in Pro has many more complex options. Otherwise, you're welcome to build something tailored to your application.

Issue #442, opened 2021-03-03
Setting a dynamic repo breaks transaction behavior


al2o3cr:
### Environment

- Oban 2.2.0
- PostgreSQL 11
- Elixir 1.11.2 / OTP 23

We're currently using dynamic repos to ensure that a glut of jobs can't starve our Phoenix handlers of connections.

### Current Behavior

When Oban is configured to use a dynamic repo (via the `get_dynamic_repo` setting), calling `Oban.insert` uses THAT repo instead of the caller's.

This is a problem when the caller has a transaction open; the writes to `Oban.Job` are visible immediately and do not roll back with the containing transaction.

### Expected Behavior

Functions like `Oban.insert` that are used by client code should not switch the dynamic repo, to prevent splitting writes across multiple connections.

sorentwo:
This is a bit tricky because `insert` and `insert_all` are used by client code as well as plugins, e.g. Cron. It would be appropriate not to switch repos for client calls, but it still needs to switch for plugins.

Perhaps the `with_dynamic_repo` function in `Oban.Repo` can use `in_transaction?` to check whether there is already a transaction in progress and skip switching?

Any additional advice, investigation or a PR is welcome.

al2o3cr:
@sorentwo agreed - it's going to be tricky to sort out some of those callers. I'll continue investigating.

For the immediate fix, we've copied the couple functions from `Oban` that we use in application code to our `Repo` and modified them to clear the `get_dynamic_repo` value before calling the functions from `Oban.Query` (similar to how `Oban.Query.fetch_or_insert_job/4` [overrides `repo`](https://github.com/sorentwo/oban/blob/2ee44412f8ed3b5940017131eb29bf5296a7cf55/lib/oban/query.ex#L59)). This is obviously Not Good from a maintainability standpoint, but it should stop the phantom jobs. 👻

al2o3cr:
One immediate thing that jumped out - the typing in `Oban.Repo` is a little weird; `Oban.Repo.config` is:
```
  @type config :: %{
          :repo => module,
          optional(:get_dynamic_repo) => (() -> pid | atom),
          optional(:log) => false | Logger.level(),
          optional(:prefix) => binary(),
          optional(any) => any
        }
```

but the functions that take it as an argument actually pass in an `Oban.Config` struct where `get_dynamic_repo` is either `nil` or a function. 🤔

al2o3cr:
[...]

PR #490, opened 2021-05-25
bugfix: keep transaction behavior using dynamic repo

milmazz:
One of the fundamental advantages of using Oban is that it runs on top
of PostgreSQL, this setup allow us to enqueue jobs along with multiple
repo operations, ensuring that everything is committed or rolled back
atomically using transactions.

The previous behavior fails when you set a dynamic repo as part of the
Oban configuration via the `get_dynamic_repo` setting. The Oban dynamic
repo always takes precedence even when the caller is running a
transaction.

This PR is an attempt to fix the previous issue.

Fixes: #442

### DIFF
--- a/lib/oban/repo.ex
+++ b/lib/oban/repo.ex
@@ -191,6 +191,17 @@ defmodule Oban.Repo do
   end
 
   defp in_transaction?(conf, instance) when is_pid(instance), do: conf.repo.in_transaction?()
+
+  defp in_transaction?(conf, instance) when is_atom(instance) do
+    case GenServer.whereis(instance) do
+      pid when is_pid(pid) ->
+        in_transaction?(conf, pid)
+
+      _ ->
+        false
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


lib/oban/repo.ex
```diff
@@ -173,19 +173,26 @@ defmodule Oban.Repo do
...
+    unless in_transaction?(conf, prev_instance) do
+      conf.repo.put_dynamic_repo(instance)
+    end
+
+    fun.()
+  after
+    conf.repo.put_dynamic_repo(prev_instance)
+  end
+
+  defp in_transaction?(conf, instance) when is_pid(instance), do: conf.repo.in_transaction?()
```
milmazz, line 193: * [ ] TODO: Consider the case when `instance` is a `name`, which is possible with `put_dynamic_repo/1`

sorentwo (approved):
Wonderful! Exactly the fix I hoped for 👍

PR #495, opened 2021-05-31
Fix links to Pro / Web changelogs

jaimeiniesta:


sorentwo:
Doc linking strikes again! Thanks 💛

PR #498, opened 2021-06-05
add documentation for snoozing jobs

### LINKED ISSUE #476: Make snooze not affect default retry backoff base
Hello!

I was a bit surprised by this behaviour, and not sure about the best way to report it.
I don't think it's a bug per se, so I created a feature request.

If you think the current behaviour should stay, I think it warrants a note in the Worker documentation, though.

Thank you!

I caught me by surprise that by using `{:snooze, seconds}` to postpone a job, if an error is later encountered, the snooze attempts are used to calculate the backoff.

My scenario was that I had a job X that needed to wait for a condition in order to run, in the `perform/1` function of the worker, I check for the condition, and if it's not met yet, I snooze the job. After the condition is met, I need to call a third-party API. That third party API failed, and the retry was scheduled for ~5 days in the future.

I think any snoozed attempt should not count for computing backoff by default. Would keeping track of `snooze count` in the `meta` field make sense?

I'm fully aware that I could:
- implement this without snooze
- implement the `backoff` callback myself

nirev:
Add more documentation regarding snoozing jobs and provides an example for handling backoff callback when a job is snoozed.

Should close #476

### DIFF
--- a/lib/oban/worker.ex
+++ b/lib/oban/worker.ex
@@ -48,7 +48,7 @@ defmodule Oban.Worker do
     An error is recorded using the optional reason, though the job is still successful
   * `{:error, error}` — the job failed, record the error and schedule a retry if possible
   * `{:snooze, seconds}` — mark the job as `snoozed` and schedule it to run again `seconds` in the
-    future. Snoozing a job does not change the number of retries remaining on the job.
+    future. See [Snoozing](#module-snoozing-jobs) for more details.
 
   In addition to explicit return values, any _unhandled exception_, _exit_ or _throw_ will fail
   the job and schedule a retry if possible.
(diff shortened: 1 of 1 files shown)


lib/oban/worker.ex
```diff
@@ -160,6 +160,33 @@ defmodule Oban.Worker do
         end
       end
 
+  ## Snoozing jobs
+
+  When returning `{:snooze, snooze_time}` in `c:perform/1`, the job will be postponed for at
+  least `snooze_time` seconds. Snoozing is done by incrementing the job `max_attempts` field and
```
sorentwo, line 166:
```suggestion
  least `snooze_time` seconds. Snoozing is done by incrementing the job's `max_attempts` field and
```

lib/oban/worker.ex
```diff
@@ -160,6 +160,33 @@ defmodule Oban.Worker do
         end
       end
 
+  ## Snoozing jobs
+
+  When returning `{:snooze, snooze_time}` in `c:perform/1`, the job will be postponed for at
```
sorentwo, line 165:
```suggestion
  When returning `{:snooze, snooze_time}` in `c:perform/1`, the job is postponed for at
```

lib/oban/worker.ex
```diff
@@ -160,6 +160,33 @@ defmodule Oban.Worker do
...
+  least `snooze_time` seconds. Snoozing is done by incrementing the job `max_attempts` field and
+  scheduling execution for `snooze_time` seconds in the future.
+
+  Snoozing does not change the number of retries remaining on the job, but it does increment the `attempt`
+  number each time the job is snooze, which will affect the default backoff exponential retry
```
sorentwo, line 170:
```suggestion
  number each time the job snoozes, which will affect the default backoff
```

sorentwo (commented):
This is excellent! There are only a couple of typos and nitpicks to fix up 👍

nirev:
Fixed! Thank you!

sorentwo (approved):

sorentwo:
Thanks for the docs and the fix 💛

PR #505, opened 2021-06-16
Add jitter when scheduling queue refreshes

### LINKED ISSUE #504: Apply jitter to queue refresh intervals

Queues all start around the same time, which means they'll refresh around the same time too. This can tie up multiple Ecto connections and starve the pool, leading to timeouts and errors.

Apply broad jitter to the refresh time to space `refresh` calls apart.

There isn't an alternative to spacing the refresh interval.

v0idpwn:
Closes #504.
Adds a 100% deviation over cooldown when scheduling queue refreshes.

### DIFF
--- a/lib/oban/breaker.ex
+++ b/lib/oban/breaker.ex
@@ -48,13 +48,13 @@ defmodule Oban.Breaker do
 
   @spec jitter(time :: pos_integer(), opts :: Keyword.t()) :: pos_integer()
   def jitter(time, opts \\ []) do
-    mode = opts[:mode] || @default_jitter_mode
-    mult = opts[:mult] || @default_jitter_mult
+    mode = Keyword.get(opts, :mode, @default_jitter_mode)
+    mult = Keyword.get(opts, :mult, @default_jitter_mult)
 
     diff = trunc(:rand.uniform() * mult * time)
 
     case mode do
-      :add ->
+      :inc ->
         time + diff
 
       :dec ->
--- a/lib/oban/breaker.ex
+++ b/lib/oban/breaker.ex
@@ -15,7 +15,8 @@ defmodule Oban.Breaker do
 
   @max_retries 10
   @min_delay 100
-  @jitter_mult 0.10
+  @default_jitter_mult 0.10
+  @default_jitter_mode :both
 
   defmacro trip_errors, do: [DBConnection.ConnectionError, Postgrex.Error]
 
@@ -45,6 +46,29 @@ defmodule Oban.Breaker do
     %{state | circuit: :enabled}
   end
 
+  @spec jitter(time :: pos_integer(), opts :: Keyword.t()) :: pos_integer()
+  def jitter(time, opts \\ []) do
+    mode = opts[:mode] || @default_jitter_mode
+    mult = opts[:mult] || @default_jitter_mult
+
+    diff = trunc(:rand.uniform() * mult * time)
+
+    case mode do
+      :add ->
+        time + diff
+
+      :dec ->
+        time - diff
+
+      :both ->
+        if :rand.uniform() >= 0.5 do
+          time + diff
+        else
+          time - diff
+        end
+    end
+  end
+
   @spec with_retry(fun(), integer()) :: term()
   def with_retry(fun, retries \\ 0)
 
@@ -57,11 +81,12 @@ defmodule Oban.Breaker do
   end
 
   defp lazy_retry(fun, retries) do
-    base = @min_delay * :math.pow(2, retries)
-    diff = base * @jitter_mult
-    sleep = Enum.random(trunc(base - diff)..trunc(base + diff))
+    time = @min_delay * :math.pow(2, retries)
 
-    Process.sleep(sleep)
+    time
+    |> trunc()
+    |> jitter()
+    |> Process.sleep()
 
     with_retry(fun, retries + 1)
   end
--- a/lib/oban/queue/producer.ex
+++ b/lib/oban/queue/producer.ex
@@ -201,7 +201,7 @@ defmodule Oban.Queue.Producer do
   end
 
   defp schedule_refresh(%State{} = state) do
-    cooldown = Breaker.jitter(state.refresh_interval, mode: :add, mult: 1)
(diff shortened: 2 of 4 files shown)


lib/oban/queue/producer.ex
```diff
@@ -201,11 +203,18 @@ defmodule Oban.Queue.Producer do
...
 
   defp schedule_refresh(%State{} = state) do
-    timer = Process.send_after(self(), :refresh, state.refresh_interval)
+    timer = Process.send_after(self(), :refresh, jitter(state.refresh_interval))
 
     %{state | refresh_timer: timer}
   end
 
+  defp jitter(time) do
+    deviation = :rand.uniform() * @max_refresh_deviation * time
```
sorentwo, line 212: This is the second or third jitter implementation we have sprinkled around. What do you think about adding a helper function for it to Breaker?

v0idpwn: Great point, I didn't notice. I added a jitter implementation on Broker and changed to use it everywhere. I tested other implementations and this one was slightly faster.

lib/oban/queue/producer.ex
```diff
@@ -18,6 +18,7 @@ defmodule Oban.Queue.Producer do
       :refresh_timer,
       dispatch_cooldown: 5,
       refresh_interval: :timer.seconds(30),
+      refresh_max_interval: :timer.seconds(60),
```
v0idpwn, line 21: How the interface for setting these two configurations should be?

sorentwo: Let's not change this field. The implementation is an internal detail, we don't need to expose any controls over the jitter or imply how the interval is actually used.

lib/oban/breaker.ex
```diff
@@ -45,6 +45,17 @@ defmodule Oban.Breaker do
     %{state | circuit: :enabled}
   end
 
+  @spec jitter(time :: pos_integer()) :: pos_integer()
+  def jitter(time) do
+    diff = trunc(:rand.uniform() * @jitter_mult * time)
```
sorentwo, line 50: What if we pass the `mult` through as an option with a default? That will let us tweak the range from outside the module. The `jitter/2` function can then be rewritten using options; rather than a separate clause, it's an option that controls whether the jitter is applied to the leading, trailing, or both edges. WDYT?

sorentwo:
@v0idpwn Do you have time to finish this one off or would you like me to pick it up?

lib/oban/breaker.ex
```diff
@@ -45,6 +46,29 @@ defmodule Oban.Breaker do
     %{state | circuit: :enabled}
   end
 
+  @spec jitter(time :: pos_integer(), opts :: Keyword.t()) :: pos_integer()
+  def jitter(time, opts \\ []) do
+    mode = opts[:mode] || @default_jitter_mode
+    mult = opts[:mult] || @default_jitter_mult
```
sorentwo, line 52: Purely for consistency with defaults elsewhere:

```suggestion
    mode = Keyword.get(opts, :mode, @default_jitter_mode)
    mult = Keyword.get(opts, :mult, @default_jitter_mult)
```

lib/oban/breaker.ex
```diff
@@ -45,6 +46,29 @@ defmodule Oban.Breaker do
...
+
+    diff = trunc(:rand.uniform() * mult * time)
+
+    case mode do
+      :add ->
```
sorentwo, line 57: I'd expect pairs to be `:add / :sub` or `:inc / :dec`, rather than a mixture of the two.

```suggestion
      :inc ->
```

lib/oban/queue/producer.ex
```diff
@@ -201,7 +201,8 @@ defmodule Oban.Queue.Producer do
   end
 
   defp schedule_refresh(%State{} = state) do
-    timer = Process.send_after(self(), :refresh, state.refresh_interval)
+    cooldown = Breaker.jitter(state.refresh_interval, mode: :add, mult: 1)
```
sorentwo, line 204: The refresh interval needs to use `:dec` to be _at least as frequent as the interval_. Otherwise, with a tight configuration, a plugin like Lifeline could erroneously rescue jobs from a running producer.

```suggestion
    cooldown = Breaker.jitter(state.refresh_interval, mode: :dec, mult: 0.5)
```

lib/oban/worker.ex
```diff
@@ -303,10 +302,8 @@ defmodule Oban.Worker do
         round(attempt / max_attempts * @max_for_backoff)
       end
 
-    base = :math.pow(2, clamped_attempt) + @backoff_base
-    diff = base * @backoff_jitter
-
-    Enum.random(trunc(base - diff)..trunc(base + diff))
+    time = trunc(:math.pow(2, clamped_attempt) + @backoff_base)
+    Breaker.jitter(time)
```
sorentwo, line 306:
```suggestion
    time = trunc(:math.pow(2, clamped_attempt) + @backoff_base)

    Breaker.jitter(time)
```

test/oban/breaker_test.exs
```diff
@@ -0,0 +1,18 @@
+defmodule Oban.BreakerTest do
+  use Oban.Case, async: true
+  use ExUnitProperties
+
+  alias Oban.Breaker
+
+  property "jitter creates time deviations within interval" do
+    check all mode <- one_of([:add, :dec, :both]),
```
sorentwo, line 8:
```suggestion
    check all mode <- one_of([:inc, :dec, :both]),
```

test/oban/breaker_test.exs
```diff
@@ -0,0 +1,18 @@
+defmodule Oban.BreakerTest do
+  use Oban.Case, async: true
+  use ExUnitProperties
+
+  alias Oban.Breaker
+
+  property "jitter creates time deviations within interval" do
```
sorentwo, line 7: Love the property test here 👍

sorentwo (approved):
This is fantastic! Thanks for making all those changes. I've left a few nitpicks for consistency and I'll merge the suggestions in myself.

lib/oban/queue/producer.ex
v0idpwn: Oh, I thought the interval was expected to be the minimum time :)

PR #508, opened 2021-06-23
Fix typo in down version docs

jc00ke:
Yeah, we're _that_ far behind on one of our apps :grimacing:


sorentwo:
Your migration secret is safe 🤫

PR #510, opened 2021-07-01
Add ability to include unit for :schedule_in

andersonmcook:
This adds the ability for users to specify `schedule_in` as a tuple of `{integer, unit}` when creating a job.

The time units and implementation are carried over from Oban Pro.

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -59,9 +59,7 @@ defmodule Oban.Job do
              | :day
              | :days
              | :week
-             | :weeks
-             | :month
-             | :months}
+             | :weeks}
 
   @type option ::
           {:args, args()}
@@ -325,8 +323,6 @@ defmodule Oban.Job do
     days
     week
     weeks
-    month
-    months
   )a
 
   defguardp is_timestampable(value)
@@ -432,8 +428,6 @@ defmodule Oban.Job do
   defp to_timestamp({days, :days}), do: to_timestamp({days, :day})
   defp to_timestamp({weeks, :week}), do: to_timestamp(weeks * 7 * 24 * 60 * 60)
   defp to_timestamp({weeks, :weeks}), do: to_timestamp({weeks, :week})
-  defp to_timestamp({months, :month}), do: to_timestamp(months * 30 * 24 * 60 * 60)
-  defp to_timestamp({months, :months}), do: to_timestamp({months, :month})
 
   defp validate_unique_opts(unique) do
     Enum.reduce_while(unique, :ok, fn {key, val}, _acc ->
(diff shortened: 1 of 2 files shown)


lib/oban/job.ex
```diff
@@ -47,13 +47,29 @@ defmodule Oban.Job do
...
+             | :minute
+             | :minutes
+             | :hour
+             | :hours
+             | :day
+             | :days
+             | :week
+             | :weeks
+             | :month
+             | :months}
```
sorentwo, line 64: Due to the variable number of days in a month and the expected accuracy of `scheduled_at,` I don't think we should support months. Of course, weeks may differ between calendars as well, but as this is just syntax sugar, I think using weeks is fine.

lib/oban/job.ex
```diff
@@ -293,10 +314,29 @@ defmodule Oban.Job do
...
+    day
+    days
+    week
+    weeks
+    month
+    months
+  )a
+
+  defguardp is_timestampable(value)
+            when is_integer(value) or
```
sorentwo, line 333: Lovely use of a private guard. Something I noticed, between this and the property test, is that we're accepting negative values. Enforcing positive integers now would be a backward-incompatible change, but it struck me.

test/oban/job_test.exs
```diff
@@ -11,10 +12,20 @@ defmodule Oban.JobTest do
       assert changeset.changes[:scheduled_at]
     end
 
+    property "scheduling a job for the future using a tuple" do
+      check all schedule_in <- tuple({integer(), time_unit()}) do
+        changeset = Job.new(%{}, worker: Fake, schedule_in: schedule_in)
+        assert changeset.changes[:state] == "scheduled"
+        assert changeset.changes[:scheduled_at]
```
sorentwo, line 19: Love the property test here! It would be nice to limit the value to positive integers and then make an assertion that `scheduled_at` was both a DateTime and in the future, rather than a plain truthy check.

sorentwo (changes requested):

lib/oban/job.ex
andersonmcook: I noticed that too.  The Stager test makes use of the ability to schedule jobs in the past.

test/oban/job_test.exs
```diff
@@ -107,4 +121,19 @@ defmodule Oban.JobTest do
...
+      :seconds,
+      :minute,
+      :minutes,
+      :hour,
+      :hours,
+      :day,
+      :days,
+      :week,
+      :weeks
+    ])
```
sorentwo, line 137: Without month variants this fits nicely into a single line:

```suggestion
    one_of([:second, :seconds, :minute, :minutes, :hour, :hours, :day, :days, :week, :weeks])
```

sorentwo (approved):
This is excellent!

PR #511, opened 2021-07-02
Include repo option in Oban.Testing.perform_job spec

michallepicki:
Fixes "Function ... has no local return" dialyzer warnings on call site


sorentwo:
Thanks for the fix! 💛

Issue #503, opened 2021-06-15
Forward all relevant queue options to start_queue/2,3


sorentwo:
### Is your feature request related to a problem? Please describe.

The `start_queue` functions only accept a queue name and a limit. There are other options available for basic queues, and an expanding set for other engine's (e.g. `SmartEngine`) queues.

### Describe the Solution You'd Like

Serialize and forward all options passed to `start_queue` through.

### Describe Alternatives You've Considered

There isn't any other way to pass those options through to queues at runtime.

PR #513, opened 2021-07-07
Raise ArgumentError if `:limit` is missing from opts

derekkraan:
Without this check, Oban crashes the queue elsewhere with a weird error:

```[error] GenServer {Oban.Registry, {Oban, {:producer, "my_queue"}}} terminating          
** (ArithmeticError) bad argument in arithmetic expression                             
    :erlang.-(nil, 0)                                                                  
    (oban 2.7.2) lib/oban/queue/basic_engine.ex:54: Oban.Queue.BasicEngine.fetch_jobs/3
```


sorentwo:
@derekkraan Thanks for the PR! I'm currently revamping the queue options as part of #503 and have a fix for this issue included there already.

PR #514, opened 2021-07-07
Add Oban.retry_all_jobs/1

VitorTrin:
This adds a function that retries all jobs that need retrying.
I do not consider completed jobs for this function because usually you want to retry something that failed.
We use this pattern so much that I thought other people would profit from it too.

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -648,6 +648,22 @@ defmodule Oban do
   end
 
   @doc """
+  Behaves like `retry_job/2`, but retries all `discarded` and `retryable` jobs.
+
+  ## Example
+
+      Oban.retry_all_jobs()
+      :ok
+  """
+  @doc since: "2.8.0"
+  @spec retry_all_jobs(name :: atom()) :: :ok
+  def retry_all_jobs(name \\ __MODULE__) do
+    name
+    |> config()
+    |> Query.retry_all_jobs()
+  end
+
+  @doc """
   Cancel an `executing`, `available`, `scheduled` or `retryable` job and mark it as `cancelled` to
   prevent it from running. If the job is currently `executing` it will be killed and otherwise it
   is ignored.
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -658,8 +658,8 @@ defmodule Oban do
       Oban.Job |> Ecto.Query.where(priority: 0) |> Oban.retry_all_jobs()
       :ok
   """
-  @doc since: "2.8.0"
-  @spec retry_all_jobs(name :: atom(), queryable :: Ecto.Queryable.t()) :: :ok
+  @doc since: "2.9.0"
+  @spec retry_all_jobs(name :: atom(), queryable :: Ecto.Queryable.t()) :: {:ok, integer()}
   def retry_all_jobs(name \\ __MODULE__, queryable \\ Job) do
     name
     |> config()
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -666,10 +666,6 @@ defmodule Oban do
     |> Query.retry_all_jobs(queryable)
   end
 
-  # @doc since: "2.8.0"
-  # @spec retry_all_jobs(queryable :: Ecto.Queryable.t()) :: :ok
-  # defdelegate retry_all_jobs(queryable), to: Query
-
   @doc """
   Cancel an `executing`, `available`, `scheduled` or `retryable` job and mark it as `cancelled` to
   prevent it from running. If the job is currently `executing` it will be killed and otherwise it
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -648,23 +648,9 @@ defmodule Oban do
   end
 
   @doc """
-  Retries all discarded and retryable jobs. For more complex mass retries, see `retry_all_jobs/2`.
-
-  ## Example
-
-      Oban.retry_all_jobs()
-      :ok
-  """
-  @doc since: "2.8.0"
-  @spec retry_all_jobs :: :ok
-  def retry_all_jobs() do
-    __MODULE__
-    |> config()
-    |> Query.retry_all_jobs()
-  end
-
-  @doc """
-  Retries all jobs that match on the given queryable.
+  Retries all jobs that match on the given queryable. Please note that no matter the used
+  queryable it will never retry available, executing or scheduled jobs.
+  If no queryable is given, all jobs in states where retry is possible will be retried.
 
   ## Example
 
@@ -674,7 +660,7 @@ defmodule Oban do
   """
   @doc since: "2.8.0"
   @spec retry_all_jobs(name :: atom(), queryable :: Ecto.Queryable.t()) :: :ok
-  def retry_all_jobs(name \\ __MODULE__, queryable) do
+  def retry_all_jobs(name \\ __MODULE__, queryable \\ Job) do
     name
     |> config()
     |> Query.retry_all_jobs(queryable)
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -655,7 +655,11 @@ defmodule Oban do
 
   ## Example
 
-      # Retries all jobs with priority 0
+      # Retries all retryable jobs
+      Oban.retry_all_jobs()
+      {:ok, 9}
+      
+      # Retries all retryable jobs with priority 0
       Oban.Job |> Ecto.Query.where(priority: 0) |> Oban.retry_all_jobs()
       {:ok, 5}
   """
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -648,9 +648,10 @@ defmodule Oban do
   end
 
   @doc """
-  Retries all jobs that match on the given queryable. Please note that no matter the used
-  queryable it will never retry available, executing or scheduled jobs.
-  If no queryable is given, all jobs in states where retry is possible will be retried.
+  Retries all jobs that match on the given queryable. Please note that no matter the
+  queryable constraints, it will never retry `available`, `executing` or `scheduled` jobs.
+  
+  If no queryable is given, Oban will retry all jobs in retryable states.
 
   ## Example
 
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -656,7 +656,7 @@ defmodule Oban do
 
       # Retries all jobs with priority 0
       Oban.Job |> Ecto.Query.where(priority: 0) |> Oban.retry_all_jobs()
-      :ok
+      {:ok, 5}
   """
   @doc since: "2.9.0"
   @spec retry_all_jobs(name :: atom(), queryable :: Ecto.Queryable.t()) :: {:ok, integer()}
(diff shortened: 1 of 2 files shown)


lib/oban/query.ex
```diff
@@ -84,6 +84,27 @@ defmodule Oban.Query do
     :ok
   end
 
+  @spec retry_all_jobs(Config.t()) :: :ok
+  def retry_all_jobs(conf) do
+    query =
+      Job
+      |> where([j], j.state not in ~w<available executing scheduled completed>)
```
v0idpwn, line 91: I think you don't want to retry `cancelled` jobs, also.

Maybe using `in` instead of `not in` would be a better fit here too

sorentwo:
Thanks for the PR!

There is great utility in retrying or canceling multiple jobs at once. Operating on _all_ jobs without any explicit controls isn't a viable option, though. Many applications have millions of jobs sitting around. This operation would lock the table and bring processing to a halt. Even with some warnings on the function, that's too big a risk for us to include as a blessed operation.

VitorTrin:
This isn't a reason for not letting an utility in. There is the term "sharp knifes", tools that in the wrong hands can cause a lot of trouble, but that is not a reason to avoiding giving it to power users. Should Ecto not implement locking because improper usage may led to locking tables for long times and causing outrages?

sorentwo:
@VitorTrin That is absolutely a reason for not letting a utility in, especially something as niche as retrying every job in the database. On the other hand, a function that accepted a batch of job ids to retry provides the same benefit and is both explicit and composable.

Compare this implementation:

```elixir
Oban.Job
|> where([j], j.state in ["discarded", "retryable"])
|> select([j], j.id)
|> MyApp.Repo.all()
|> Oban.retry_all_jobs()
```

to this:

```elixir
Oban.retry_all_jobs()
```

While the second is much terser, the first gives the application complete control over which jobs to retry. What if an application doesn't want to retry discarded jobs? They wouldn't be able to use the second implementation at all.

VitorTrin:
This is another argument entirely, that the function isn't flexible enough, and I expect suggestions. The first implementation you suggested isn't much of an improvement over doing the update yourself and takes 2 operations, but it's a start. Maybe this function can take a queriable (most flexible and only one repo operation, but requires the user to know how jobs are represented in the db) or you can send a list of states as an option.

  > What if an application doesn't want to retry discarded jobs?
  
  Well, what does the application wants to retry at all then? Successfull jobs? Jobs that will be retried anyway if you do nothing?

v0idpwn:
Note: given an implementation flexible enough, Oban.Plugins.Stager could rely on retry_all_jobs

sorentwo:
> This is another argument entirely, that the function isn't flexible enough, and I expect suggestions.

I'm sorry that it came across that way. I thought I suggested an alternative that was composable, but reading through my closing message I see that wasn't apparent.

> The first implementation you suggested isn't much of an improvement over doing the update yourself and takes 2 operations.

Writing a function to batch retry every job in the database in your own application seems reasonable to me, considering how niche that is.

> Maybe this function can take a queriable (most flexible and only one repo operation, but requires the user to know how jobs are represented in the DB) or you can send a list of states as an option.

The [Oban.Job schema is public and documented](https://hexdocs.pm/oban/Oban.Job.html#content), so understanding how to build an Ecto query for jobs seems reasonable. Taking a list of states would be better, though still limiting—what if you want to retry all discarded jobs for a particular worker or queue?

> Well, what does the application wants to retry at all then? Successfull jobs? Jobs that will be retried anyway if you do nothing?

Perhaps they have jobs that won't be retried for a few weeks that they want to retry now. I'm not sure, and time has shown that I can't predict all the things people want to do with their jobs 🙂

VitorTrin:
@sorentwo Could you please reopen the pull request? I changed to receive a queryable but as the pr is closed it does not reflect here

sorentwo:
@VitorTrin re-opened. I'll take a look at your new approach tomorrow 👍

lib/oban.ex
```diff
@@ -647,6 +647,43 @@ defmodule Oban do
...
+  end
+
+  @doc """
+  Retries all jobs that match on the given queryable.
+
+  ## Example
+
+      # Retries all jobs with priority 0
+      Oban.Job |> Ecto.Query.where(priority: 0) |> Oban.retry_all_jobs()
+      :ok
```
v0idpwn, line 673: This example looks a bit dangerous to me. 

It makes me think that only failed jobs will be retried, but actually it also retries completed or even **executing** jobs with prio 0.

v0idpwn (commented):
I quite like the queryable approach, but I think the current form may induce one to retry executing jobs

lib/oban.ex
sorentwo: That's an excellent point. Retrying `executing` jobs would cause a lot of issues. The `retry_all_jobs` function would need to apply a `status` condition, which may defeat the point of passing in a query entirely.

lib/oban.ex
```diff
@@ -647,6 +647,43 @@ defmodule Oban do
...
+  Retries all discarded and retryable jobs. For more complex mass retries, see `retry_all_jobs/2`.
+
+  ## Example
+
+      Oban.retry_all_jobs()
+      :ok
+  """
+  @doc since: "2.8.0"
+  @spec retry_all_jobs :: :ok
+  def retry_all_jobs() do
```
sorentwo, line 660: This clause can be removed. With some defaults, both clauses are covered by `retry_all_jobs/2`.

lib/oban.ex
```diff
@@ -647,6 +647,43 @@ defmodule Oban do
...
+
+  ## Example
+
+      # Retries all jobs with priority 0
+      Oban.Job |> Ecto.Query.where(priority: 0) |> Oban.retry_all_jobs()
+      :ok
+  """
+  @doc since: "2.8.0"
+  @spec retry_all_jobs(name :: atom(), queryable :: Ecto.Queryable.t()) :: :ok
+  def retry_all_jobs(name \\ __MODULE__, queryable) do
```
sorentwo, line 677: Since the `queryable` needs to have a state condition applied, this can default to `Oban.Job` (aliased as `Job`) instead:

```suggestion
  def retry_all_jobs(name \\ __MODULE__, queryable \\ Job) do
```

lib/oban.ex
```diff
@@ -647,6 +647,43 @@ defmodule Oban do
...
+    name
+    |> config()
+    |> Query.retry_all_jobs(queryable)
+  end
+
+  # @doc since: "2.8.0"
+  # @spec retry_all_jobs(queryable :: Ecto.Queryable.t()) :: :ok
+  # defdelegate retry_all_jobs(queryable), to: Query
+
```
sorentwo, line 686: ✂️ 

```suggestion
```

lib/oban/query.ex
```diff
@@ -84,6 +84,44 @@ defmodule Oban.Query do
...
+          completed_at: nil,
+          cancelled_at: nil,
+          discarded_at: nil
+        ]
+      )
+
+    Repo.update_all(conf, query, [])
+    :ok
+  end
+
```
sorentwo, line 106: This extra clause isn't necessary ✂️ 

```suggestion
```

lib/oban/query.ex
```diff
@@ -84,6 +84,44 @@ defmodule Oban.Query do
...
+  @spec retry_all_jobs(Config.t(), Ecto.Queryable.t()) :: :ok
+  def retry_all_jobs(conf, queryable) do
+    query =
+      update(queryable, [j],
```
sorentwo, line 110: The query should have a `state` condition added to it:

```suggestion
      queryable
      |> where([j], j.state not in ["available", "executing", "scheduled"])
      |> update([j],
```

Interestingly, `retry_job` can use `retry_all_jobs` with a simple `where([j], j.id == ^id)` queryable, too.

sorentwo (changes requested):
Finally back to pick this up. This is a much better approach (and something that would apply to `cancel_all_jobs`. A few changes here if you'd like to pick this back up.

lib/oban.ex
```diff
@@ -647,6 +647,25 @@ defmodule Oban do
...
+  If no queryable is given, all jobs in states where retry is possible will be retried.
+
+  ## Example
+
+      # Retries all jobs with priority 0
+      Oban.Job |> Ecto.Query.where(priority: 0) |> Oban.retry_all_jobs()
+      :ok
+  """
+  @doc since: "2.8.0"
+  @spec retry_all_jobs(name :: atom(), queryable :: Ecto.Queryable.t()) :: :ok
```
sorentwo, line 662: Since we'll get a count, it seems like something we should return. Rather than a plain `:ok`, how about `{:ok, number_of_retries}` here?

lib/oban/query.ex
```diff
@@ -84,6 +84,26 @@ defmodule Oban.Query do
...
+          max_attempts: fragment("GREATEST(?, ? + 1)", j.max_attempts, j.attempt),
+          scheduled_at: ^utc_now(),
+          completed_at: nil,
+          cancelled_at: nil,
+          discarded_at: nil
+        ]
+      )
+
+    Repo.update_all(conf, query, [])
+    :ok
```
sorentwo, line 104: Regarding my count comment above, something like this:

```suggestion
    {count, _} = Repo.update_all(conf, query, [])
    
    {:ok, count}
```

lib/oban.ex
```diff
@@ -647,6 +647,25 @@ defmodule Oban do
     |> Query.retry_job(job_id)
   end
 
+  @doc """
+  Retries all jobs that match on the given queryable. Please note that no matter the used
+  queryable it will never retry available, executing or scheduled jobs.
+  If no queryable is given, all jobs in states where retry is possible will be retried.
```
sorentwo, line 653:
```suggestion
  Retries all jobs that match on the given queryable. Please note that no matter the
  queryable constraints, it will never retry `available`, `executing` or `scheduled` jobs.
  
  If no queryable is given, Oban will retry all jobs in retryable states.
```

lib/oban.ex
```diff
@@ -647,6 +647,25 @@ defmodule Oban do
...
+
+  ## Example
+
+      # Retries all jobs with priority 0
+      Oban.Job |> Ecto.Query.where(priority: 0) |> Oban.retry_all_jobs()
+      {:ok, 5}
```
sorentwo, line 659:
```suggestion
      # Retries all retryable jobs
      Oban.retry_all_jobs()
      {:ok, 9}
      
      # Retries all retryable jobs with priority 0
      Oban.Job |> Ecto.Query.where(priority: 0) |> Oban.retry_all_jobs()
      {:ok, 5}
```

sorentwo (approved):
Nice and simple now 👍

sorentwo:
Thanks for hanging through that @VitorTrin 💛

@v0idpwn What do you think about a matching `cancel_all_jobs/0,1,2` and updating `retry_job` to use the new `retry_all` query?

PR #527, opened 2021-08-04
Add Oban.cancel_all_jobs/0,1,2

v0idpwn:
As suggested in #514


lib/oban.ex
```diff
@@ -695,6 +695,34 @@ defmodule Oban do
...
+
+  However, if it manages to complete successfully then the state will still be `completed`.
+
+  Only jobs with the statuses`executing`, `available`, `scheduled` or `retryable` can be cancelled.
+
+  ## Example
+
+  Cancel all scheduled jobs for a specific worker:
+
+      query = Ecto.Query.where(Oban.Job, worker: MyApp.MyWorker)
```
v0idpwn, line 712: wait, this query doesn't work, right?

lib/oban.ex
```diff
@@ -695,6 +695,34 @@ defmodule Oban do
     Notifier.notify(conf, :signal, %{action: :pkill, job_id: job_id})
   end
 
+  @doc """
+  Cancel many jobs based on a queryable and mark them as `cancelled` to prevent them from running.
+  If any the jobs is currently `executing` it will be killed and otherwise it is ignored.
```
sorentwo, line 700:
```suggestion
  Any currently `executing` jobs are killed while the others are ignored.
```

lib/oban.ex
```diff
@@ -695,6 +695,34 @@ defmodule Oban do
...
+
+  If an executing job happens to fail before it can be cancelled the state is set to `cancelled`.
+
+  However, if it manages to complete successfully then the state will still be `completed`.
+
+  Only jobs with the statuses`executing`, `available`, `scheduled` or `retryable` can be cancelled.
```
sorentwo, line 706:
```suggestion
  If executing jobs happen to fail before cancellation then the state is set to `cancelled`.
  However, any that complete successfully will remain `completed`.

  Only jobs with the statuses `executing`, `available`, `scheduled`, or `retryable` can be cancelled.
```

lib/oban.ex
sorentwo:
```suggestion
      query = Ecto.Query.where(Oban.Job, worker: "MyApp.MyWorker")
```

lib/oban/queue/basic_engine.ex
```diff
@@ -141,16 +141,24 @@ defmodule Oban.Queue.BasicEngine do
...
+
+  @impl true
+  def cancel_all_jobs(%Config{} = conf, queryable) do
     query =
-      Job
-      |> where([j], j.id == ^id)
+      queryable
       |> where([j], j.state not in ["completed", "discarded", "cancelled"])
+      |> update([j], set: [state: "cancelled", cancelled_at: ^utc_now()])
+      |> select([:id])
```
sorentwo, line 157: We can avoid the extra map by selecting the id directly:

```suggestion
      |> select([j], j.id)
```

lib/oban.ex
```diff
@@ -695,6 +695,34 @@ defmodule Oban do
...
+      :ok
+  """
+  def cancel_all_jobs(name \\ __MODULE__, queryable \\ Job) do
+    conf = config(name)
+
+    {:ok, job_ids} = Engine.cancel_all_jobs(conf, queryable)
+
+    Enum.each(job_ids, &Notifier.notify(conf, :signal, %{action: :pkill, job_id: &1}))
+
+    {:ok, Enum.count(job_ids)}
```
sorentwo, line 723: A couple of things here:

1. `Notifier.notify/3` can take a list of messages instead, which lets it make a single database call
2. We should only emit a notification if there is at least one job id
3. `Enum.count/1` will delegate to `length/1` for a list, may as well skip the middleman

```suggestion
    case Engine.cancel_all_jobs(conf, queryable) do
      {:ok, [_ | _] = job_ids} ->
        payload = Enum.map(job_ids, &%{action: :pkill, job_id: &1})

        Notifier.notify(conf, :signal, payload)

        {:ok, length(job_ids)}

      _ ->
        {:ok, 0}
    end
```

sorentwo (changes requested):
Great start, I dig the approach 👍

lib/oban.ex
```diff
@@ -695,6 +695,39 @@ defmodule Oban do
...
+      query = Ecto.Query.where(Oban.Job, worker: "MyApp.MyWorker")
+      Oban.cancel_all_jobs(Oban, query)
+      :ok
+  """
+  def cancel_all_jobs(name \\ __MODULE__, queryable \\ Job) do
+    conf = config(name)
+
+    case Engine.cancel_all_jobs(conf, queryable) do
+      {:ok, [_ | _] = job_ids} ->
+        payload = Enum.map(job_ids, fn id -> %{action: :pkill, job_id: id} end)
```
sorentwo, line 720: Not a fan of the captured map syntax, eh?

sorentwo (approved):
Looks great! There is an `swp` file sitting around, so I'll fix it up and merge locally.

sorentwo:
Merged with 313a753

PR #528, opened 2021-08-06
Add additional telemetry events

dvic:
This PR adds additional telemetry events for 
- `Engine` operations that touch the database
- `Notifier` calls

### DIFF
--- a/lib/oban/notifier.ex
+++ b/lib/oban/notifier.ex
@@ -158,17 +158,29 @@ defmodule Oban.Notifier do
   def notify(conf_or_server \\ Oban, channel, payload)
 
   def notify(%Config{} = conf, channel, payload) when is_channel(channel) do
-    conf.name
-    |> Registry.whereis(Oban.Notifier)
-    |> conf.notifier.notify(channel, normalize_payload(payload))
+    with_span(conf, channel, payload, fn ->
+      conf.name
+      |> Registry.whereis(Oban.Notifier)
+      |> conf.notifier.notify(channel, normalize_payload(payload))
+    end)
   end
 
   def notify(server, channel, payload) when is_channel(channel) do
     conf = Oban.config(server)
 
-    conf.name
-    |> Registry.whereis(Oban.Notifier)
-    |> conf.notifier.notify(channel, normalize_payload(payload))
+    with_span(conf, channel, payload, fn ->
+      conf.name
+      |> Registry.whereis(Oban.Notifier)
+      |> conf.notifier.notify(channel, normalize_payload(payload))
+    end)
+  end
+
+  defp with_span(conf, channel, payload, cb) do
+    tele_meta = %{conf: conf, channel: channel, payload: payload}
+
+    :telemetry.span([:oban, :notifier, :notify], tele_meta, fn ->
+      {cb.(), tele_meta}
+    end)
   end
 
   defp normalize_payload(payload) do
--- a/lib/oban/notifier.ex
+++ b/lib/oban/notifier.ex
@@ -171,11 +171,11 @@ defmodule Oban.Notifier do
     |> notify(channel, payload)
   end
 
-  defp with_span(conf, channel, payload, cb) do
+  defp with_span(conf, channel, payload, fun) do
     tele_meta = %{conf: conf, channel: channel, payload: payload}
 
     :telemetry.span([:oban, :notifier, :notify], tele_meta, fn ->
-      {cb.(), tele_meta}
+      {fun.(), tele_meta}
     end)
   end
 
--- a/lib/oban/notifier.ex
+++ b/lib/oban/notifier.ex
@@ -166,13 +166,9 @@ defmodule Oban.Notifier do
   end
 
   def notify(server, channel, payload) when is_channel(channel) do
-    conf = Oban.config(server)
-
-    with_span(conf, channel, payload, fn ->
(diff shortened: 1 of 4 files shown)


lib/oban/queue/producer.ex
```diff
@@ -221,7 +227,13 @@ defmodule Oban.Queue.Producer do
   end
 
   defp start_jobs(%State{} = state) do
-    {:ok, {meta, jobs}} = Engine.fetch_jobs(state.conf, state.meta, state.running)
+    tele_meta = tele_meta(state)
+
+    {meta, jobs} =
```
sorentwo, line 232: This is already covered by the `dispatch` function, though I admit the `[:oban, :producer, _]` namespace isn't very helpful.

sorentwo (commented):
There are other db interaction points that don't have any telemetry event around them, e.g. `cancel`, `retry`, `put_meta`. What do you think about wrapping each of the functions in the `Engine` module instead?

lib/oban/queue/producer.ex
dvic: Ah I see, so just `[:oban, :engine, :refresh]` then?

dvic: Never mind, I misunderstood you, I'll move the Engine events to the engine module.

dvic:
> There are other db interaction points that don't have any telemetry event around them, e.g. `cancel`, `retry`, `put_meta`. What do you think about wrapping each of the functions in the `Engine` module instead?

Yes that seems like a better idea, I'll do that.

By the way: I just pushed also telemetry events for Notifier, need to still update the docs though (will do that once we settle all the events). Any other candidates that you can think of?

dvic:
Currently I'm missing only the following query:

```sql
UPDATE "public"."oban_producers" AS o0 SET "updated_at" = $1 WHERE (o0."uuid" = $2) RETURNING o0."uuid", o0."name", o0."node", o0."queue", o0."meta", o0."running_ids", o0."started_at", o0."updated_at"
```

any idea where this one is coming from?

lib/oban/notifier.ex
```diff
@@ -158,17 +158,29 @@ defmodule Oban.Notifier do
...
     conf = Oban.config(server)
 
-    conf.name
-    |> Registry.whereis(Oban.Notifier)
-    |> conf.notifier.notify(channel, normalize_payload(payload))
+    with_span(conf, channel, payload, fn ->
+      conf.name
+      |> Registry.whereis(Oban.Notifier)
+      |> conf.notifier.notify(channel, normalize_payload(payload))
+    end)
```
sorentwo, line 175: There was a little duplication in here before, but this amount seems unnecessary. How about this instead?

```elixir
    server
    |> Oban.config()
    |> notify(channel, payload)
```

lib/oban/telemetry.ex
```diff
@@ -69,6 +69,82 @@ defmodule Oban.Telemetry do
   * `:conf` — the config of the Oban supervised producer
   * `:dispatched_count` — the number of jobs fetched and started by the producer
 
+  ### Engine Events
+
+  Oban emits telemetry span events for the following Engine operations:
```
sorentwo, line 74: The `init`, `refresh`, and `put_meta` events may also write to the database. So you'll need to wrap those as well to cover all of the interactions.

lib/oban/notifier.ex
```diff
@@ -158,17 +158,29 @@ defmodule Oban.Notifier do
...
+  end
+
+  defp with_span(conf, channel, payload, cb) do
```
sorentwo, line 178: Very minor, but for the sake of consistency, I'd prefer `fun` rather than `cb` here and in the engine's function.

sorentwo (changes requested):
Nice work. I've left a few nitpicks and pointed out a couple other db interaction spots. After that, this is good to go!

dvic:
> Nice work. I've left a few nitpicks and pointed out a couple other db interaction spots. After that, this is good to go!

Done :)

dvic:
Is there a flaky test? Locally all tests run fine on 1.12 + OTP 24?

sorentwo:
> Is there a flaky test? Locally all tests run fine on 1.12 + OTP 24?

There are some flaky tests due to race conditions and general timing problems. It shows up in CI due to the unpredictable resource contention. I rarely see them locally, and when I do, I fix them.

lib/oban/telemetry.ex
```diff
@@ -69,6 +69,106 @@ defmodule Oban.Telemetry do
...
+  * `[:oban, :engine, :fetch_jobs, :start | :stop | :exception]`
+
+  | event        | measures       | metadata           |
+  | ------------ | -------------- | ------------------ |
+  | `:start`     | `:system_time` | `:conf`, `engine`  |
+  | `:stop`      | `:duration`    | `:conf`, `engine`  |
+  | `:exception` | `:duration`    | `:conf`, `engine`  |
+
+
+  * `[:oban, :engine, :complete_job, :start | :stop | :exception]` — when an engine operation is called
```
sorentwo, line 109: This comment would apply to all of the events.

```suggestion
  * `[:oban, :engine, :complete_job, :start | :stop | :exception]`
```

sorentwo (approved):

sorentwo:
Beautiful work! Thanks 💛

PR #531, opened 2021-08-07
Fix double negation in docs

preciz:
I believe this sentence states the the opposite due to double negation. 

Thank you for reviewing.


sorentwo:
You're absolutely right! Thanks 💛

PR #533, opened 2021-08-17
 Add option to inhibit schema creation in migrations

maxmarcon:
Hi @sorentwo! 👋 I don't know if you remember, but we discussed on Slack the addition of an option to inhibit the schema creation during the migration when the `prefix` option is used. Automatic schema creation can cause problems if the database user in the production database you are deploying to doesn't have permissions to create new schemas.

Here's my stab at it. My personal opinion is that `Oban.Migrations` could use some refactoring, but I tried to keep my changes to a bare minimum. Looking forward to your feedback!

### DIFF
--- a/lib/oban/migrations.ex
+++ b/lib/oban/migrations.ex
@@ -59,6 +59,20 @@ defmodule Oban.Migrations do
     prefix: "private",
     ...
   ```
+
+  In some cases, for example if your "private" schema already exists and your database user in production 
+  doesn't have permissions to create a new schema, trying to create the schema from the migration will result
+  in an error. In such situations, it may be useful to inhibit the creation of the "private" schema:
+
+  ```elixir
+  defmodule MyApp.Repo.Migrations.AddPrefixedObanJobsTable do
+    use Ecto.Migration
+
+    def up, do: Oban.Migrations.up(prefix: "private", create_schema: false)
+
+    def down, do: Oban.Migrations.down(prefix: "private")
+  end
+  ```
   """
 
   use Ecto.Migration
@@ -85,14 +99,19 @@ defmodule Oban.Migrations do
   Run migrations in an alternate prefix:
 
       Oban.Migrations.up(prefix: "payments")
+
+  Run migrations in an alternate prefix but don't try to create the schema:
+
+      Oban.Migrations.up(prefix: "payments", create_schema: false)
   """
   def up(opts \\ []) when is_list(opts) do
     prefix = Keyword.get(opts, :prefix, @default_prefix)
     version = Keyword.get(opts, :version, @current_version)
+    create_schema = Keyword.get(opts, :create_schema, true)
     initial = migrated_version(repo(), prefix)
 
     cond do
-      initial == 0 -> change(prefix, @initial_version..version, :up)
+      initial == 0 -> change(prefix, @initial_version..version, :up, create_schema)
       initial < version -> change(prefix, (initial + 1)..version, :up)
       true -> :ok
     end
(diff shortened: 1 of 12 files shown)


lib/oban/migrations.ex
```diff
@@ -148,13 +167,20 @@ defmodule Oban.Migrations do
...
+        if index == 1 && direction == :up do
+          [prefix, create_schema]
+        else
+          [prefix]
+        end
+
       [__MODULE__, "V#{pad_idx}"]
       |> Module.concat()
-      |> apply(direction, [prefix])
+      |> apply(direction, args)
```
sorentwo, line 183: What if this were `[opts]`, where `opts` included the prefix and whether to create the schema? That would avoid the variable arity and the special casing for the first migration.

lib/oban/migrations.ex
```diff
@@ -85,14 +99,19 @@ defmodule Oban.Migrations do
...
       Oban.Migrations.up(prefix: "payments")
+
+  Run migrations in an alternate prefix but don't try to create the schema:
+
+      Oban.Migrations.up(prefix: "payments", create_schema: false)
   """
   def up(opts \\ []) when is_list(opts) do
     prefix = Keyword.get(opts, :prefix, @default_prefix)
     version = Keyword.get(opts, :version, @current_version)
+    create_schema = Keyword.get(opts, :create_schema, true)
```
sorentwo, line 110: What about using `prefix != "public"` as the default here? That avoids the compound logic and special-casing in the first migration.

sorentwo:
@maxmarcon I'm sure others have encountered this problem and not said anything. Great to see that you didn't forget about this!

lib/oban/migrations.ex
maxmarcon: Yes I agree. That would require modifying each migration to receive a keyword as argument instead of `prefix`, and that's why I didn't do it. I'll do it now.

lib/oban/migrations.ex
maxmarcon: You mean it would avoid the need for this if statement? (https://github.com/sorentwo/oban/pull/533/files#diff-4f2f4029f6bc09bf30159a9909a4edb2d56401527dcff6245b130f24cfc0eb71R7)
I don't think it would, because you could still have `prefix != "public"` and `create_schema = false`. That's why we are making this change, because we want to disable schema creation even when `prefix != "public"`.

sorentwo: I’m suggesting that as the default for `Keyword.get`. Explicitly setting `create_schema` would override it still because it doesn’t work like an OR.

maxmarcon: Yes I got that. But I was wondering what you mean by "That avoids the compound logic and special-casing in the first migration.". The logic of the first migration won't change

sorentwo: As is, the first migration checks both the prefix and whether a create_schema override exists. It doesn’t need to know about the special case of the public schema at all if instead it only checks whether create_schema is set.

maxmarcon: Ah! Now it's clear 👍  thanks. Good idea

lib/oban/migrations.ex
```diff
@@ -148,15 +172,24 @@ defmodule Oban.Migrations do
...
     for index <- range do
       pad_idx = String.pad_leading(to_string(index), 2, "0")
 
+      opts =
+        if direction == :up && index == 1 do
+          opts
+        else
+          Keyword.take(opts, [:prefix])
+        end
+
```
sorentwo, line 185: This is just about as awkward as the previous multi-arity version. The migration format isn't a public API, and we're free to change what gets passed through. What about using `Map.new(opts)` so that the various functions can reliably match on the `prefix`?

maxmarcon: Agreed. I thought you were keen on using Keyword for `opts` :)
But no need to use `Map.new`. I suggest we have `up` and `down` directly pass a map of options to `change`. I pushed the new changes, please take a look.

sorentwo: For public APIs I definitely prefer keywords for opts. For internal abstractions I'm ambivalent, and maps make much more sense for destructuring.

sorentwo (approved):
This is wonderful. Thanks for bearing with all those changes, the result is much cleaner.

maxmarcon:
Awesome! 🚀 You are welcome, glad I could contribute a bit. Love the project. If you need help for something in the future, just ping me

PR #549, opened 2021-09-17
Fix wrong example calculation of dispatch_cooldown in comment

MMore:


sorentwo:
Math! Thanks for the fix 💛

PR #550, opened 2021-09-19
Add customizable prefix to oban-emitted telemetry events

### LINKED ISSUE #545: Support a custom prefix for telemetry events

It's difficult to differentiate between events from multiple oban instances when using systems like Phoenix Live Dashboard. All events share a common path, e.g. `[:oban, :job, :stop]`.

Support a customizable event prefix that is included in all telemetry events. The default will be `[:oban]`, to match the current functionality.

Matching on the `conf.name` metadata. This works for ad-hoc implementations, but not for path-based measurements like Live Dashboard uses.

v0idpwn:
Closes #545.

### DIFF
--- a/lib/oban/config.ex
+++ b/lib/oban/config.ex
@@ -205,9 +205,9 @@ defmodule Oban.Config do
   end
 
   defp validate_opt!({:telemetry_prefix, prefix}) do
-    unless is_list(prefix) and Enum.all?(prefix, &is_atom/1) do
+    unless is_list(prefix) and Enum.all?(prefix, &is_atom/1) and length(prefix) <= 3 do
       raise ArgumentError,
-            "expected :telemetry_prefix to be a list of atoms, got: #{inspect(prefix)}"
+            "expected :telemetry_prefix to be a list of atoms with size <= 3, got: #{inspect(prefix)}"
     end
   end
 
--- a/lib/oban/config.ex
+++ b/lib/oban/config.ex
@@ -205,9 +205,9 @@ defmodule Oban.Config do
   end
 
   defp validate_opt!({:telemetry_prefix, prefix}) do
-    unless is_list(prefix) and Enum.all?(prefix, &is_atom/1) and length(prefix) <= 3 do
+    unless is_list(prefix) and Enum.all?(prefix, &is_atom/1) and length(prefix) in 1..3 do
       raise ArgumentError,
-            "expected :telemetry_prefix to be a list of atoms with size <= 3, got: #{inspect(prefix)}"
+            "expected :telemetry_prefix to be a non-empty list of atoms with less than 3 items, got: #{inspect(prefix)}"
     end
   end
 
--- a/lib/oban/config.ex
+++ b/lib/oban/config.ex
@@ -207,7 +207,7 @@ defmodule Oban.Config do
   defp validate_opt!({:telemetry_prefix, prefix}) do
     unless is_list(prefix) and Enum.all?(prefix, &is_atom/1) and length(prefix) in 1..3 do
       raise ArgumentError,
-            "expected :telemetry_prefix to be a non-empty list of atoms with less than 3 items, got: #{inspect(prefix)}"
+            "expected :telemetry_prefix to be a non-empty list of atoms with up to 3 items, got: #{inspect(prefix)}"
     end
   end
 
--- a/lib/oban/config.ex
+++ b/lib/oban/config.ex
@@ -18,6 +18,7 @@ defmodule Oban.Config do
           queues: [{atom(), Keyword.t()}],
           repo: module(),
           shutdown_grace_period: timeout(),
+          telemetry_prefix: [atom()],
           log: false | Logger.level(),
           get_dynamic_repo: nil | (() -> pid() | atom())
         }
@@ -36,6 +37,7 @@ defmodule Oban.Config do
             queues: [],
             repo: nil,
             shutdown_grace_period: :timer.seconds(15),
+            telemetry_prefix: [:oban],
             log: false,
             get_dynamic_repo: nil
 
@@ -199,6 +201,13 @@ defmodule Oban.Config do
     unless is_pos_integer(period) do
       raise ArgumentError,
             "expected :shutdown_grace_period to be a positive integer, got: #{inspect(period)}"
+    end
+  end
+
+  defp validate_opt!({:telemetry_prefix, prefix}) do
+    unless is_list(prefix) and Enum.all?(prefix, &is_atom/1) do
+      raise ArgumentError,
+            "expected :telemetry_prefix to be a list of atoms, got: #{inspect(prefix)}"
     end
   end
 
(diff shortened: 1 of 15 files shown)


lib/oban/telemetry.ex
```diff
@@ -305,16 +305,20 @@ defmodule Oban.Telemetry do
...
       :ok = Oban.Telemetry.attach_default_logger(:debug)
+
+  Attach a logger with a specific oban prefix:
+
+      :ok = Oban.Telemetry.attach_default_logger(:info, [:my_oban])
   """
   @doc since: "0.4.0"
   @spec attach_default_logger(Logger.level()) :: :ok | {:error, :already_exists}
-  def attach_default_logger(level \\ :info) do
+  def attach_default_logger(level \\ :info, telemetry_prefix \\ [:oban]) do
```
sorentwo, line 315: Using two defaults makes it impossible to set the prefix without setting the level. What if we turn this into a keyword list instead?

```elixir
  def attach_default_logger(level) when is_atom(level) do
    attach_default_logger(level: level)
  end

  def attach_default_logger(opts) when is_list(opts) do
    # ...
```

lib/oban/telemetry.ex
```diff
@@ -370,17 +374,19 @@ defmodule Oban.Telemetry do
...
-    |> Map.merge(converted_measurements(measure))
-    |> log_message("job:#{event}", level)
-  end
-
-  def handle_event([:oban, :circuit, event], _measure, meta, level) do
-    meta
-    |> Map.take([:message, :name])
-    |> log_message("circuit:#{event}", level)
+  def handle_event(event_name, measure, meta, level) do
+    case Enum.slice(event_name, -2..-1) do
```
sorentwo, line 378: I'm concerned about the performance of `Enum.slice` when processing a _lot_ of events. It's a bit manual, but we could specify clauses for prefixes with 1-3 values and put a limit in the `config` validation. Something like this

```elixir
[_, :job, event] -> # ...
[_, _, :job, event] -> # ...
[_, _, _, :job, event] -> # ...
```

That may be overkill, and I'm curious to see a benchmark that proves `Enum.slice` is good enough. WDYT?

test/oban/config_test.exs
```diff
@@ -81,6 +81,13 @@ defmodule Oban.ConfigTest do
       assert_valid(shutdown_grace_period: 10)
     end
 
+    test ":telemetry_prefix is validated as a list of atoms" do
+      refute_valid(telemetry_prefix: :oban)
+      refute_valid(telemetry_prefix: ["oban"])
+
+      assert_valid(telemetry_prefix: [:oban])
```
sorentwo, line 88:
```suggestion
      assert_valid(telemetry_prefix: [:oban])
      assert_valid(telemetry_prefix: [:my_app, :oban])
```

sorentwo (commented):
Nicely done! You've covered just about everything. I have one suggestion for telemetry and an open performance question about `Enum.slice`. Otherwise, this is good to go 👍

lib/oban/telemetry.ex
v0idpwn: That's a good idea! I was afraid of breaking compatibility, but your default solves it :)

lib/oban/telemetry.ex
v0idpwn: I was not sure about which was the best solution there. I'll try to come up with a benchmark

v0idpwn: Pattern matching is around 5-6x faster than relying on Enum.slice. 

I had two options: 
- defining a single clause and using a case
- defining 6 clauses

I went with the second one.

v0idpwn:
I think I addressed all your comments. Let me know if there's something else you'd like to see changed :).

Probably more docs need to be updated as well. I don't know what should be done with the countless references to `[:oban, some, event]` telemetry events there.

test/oban/config_test.exs
```diff
@@ -81,6 +81,15 @@ defmodule Oban.ConfigTest do
...
     end
 
+    test ":telemetry_prefix is validated as a list of atoms" do
+      refute_valid(telemetry_prefix: :oban)
+      refute_valid(telemetry_prefix: ["oban"])
+      refute_valid(telemetry_prefix: [:more, :than, :three, :elements])
+
+      assert_valid(telemetry_prefix: [:oban])
+      assert_valid(telemetry_prefix: [:my_app, :oban])
+    end
```
v0idpwn, line 91: Missed a case: `[]`

lib/oban/config.ex
```diff
@@ -205,9 +205,9 @@ defmodule Oban.Config do
   end
 
   defp validate_opt!({:telemetry_prefix, prefix}) do
-    unless is_list(prefix) and Enum.all?(prefix, &is_atom/1) and length(prefix) <= 3 do
+    unless is_list(prefix) and Enum.all?(prefix, &is_atom/1) and length(prefix) in 1..3 do
       raise ArgumentError,
-            "expected :telemetry_prefix to be a list of atoms with size <= 3, got: #{inspect(prefix)}"
+            "expected :telemetry_prefix to be a non-empty list of atoms with less than 3 items, got: #{inspect(prefix)}"
```
sorentwo, line 210:
```suggestion
            "expected :telemetry_prefix to be a non-empty list of atoms with up to 3 items, got: #{inspect(prefix)}"
```

lib/oban/telemetry.ex
```diff
@@ -302,19 +302,28 @@ defmodule Oban.Telemetry do
...
-  Attach a logger at the `:debug` level:
+  Attach a logger with a specific oban prefix and level:
 
-      :ok = Oban.Telemetry.attach_default_logger(:debug)
+      :ok = Oban.Telemetry.attach_default_logger(telemetry_prefix: [:my_oban], logger_level: :debug)
   """
   @doc since: "0.4.0"
   @spec attach_default_logger(Logger.level()) :: :ok | {:error, :already_exists}
-  def attach_default_logger(level \\ :info) do
+  def attach_default_logger(level_or_opts \\ :info)
```
sorentwo, line 311:
```suggestion
  def attach_default_logger(level_or_opts \\ [logger_level: :info])
```

sorentwo (approved):

lib/oban/telemetry.ex
```diff
@@ -302,19 +302,28 @@ defmodule Oban.Telemetry do
...
-  Attach a logger at the `:debug` level:
+  Attach a logger with a specific oban prefix and level:
 
-      :ok = Oban.Telemetry.attach_default_logger(:debug)
+      :ok = Oban.Telemetry.attach_default_logger(telemetry_prefix: [:my_oban], logger_level: :debug)
   """
   @doc since: "0.4.0"
   @spec attach_default_logger(Logger.level()) :: :ok | {:error, :already_exists}
-  def attach_default_logger(level \\ :info) do
+  def attach_default_logger(level_or_opts \\ [logger_level: :info])
```
sorentwo, line 311: We'll need a different typespec here:

```elixir
  @type attach_option :: {:logger_level, Logger.level()} | {:telemetry_prefix, [atom()]}
```

```elixir
  @spec attach_default_logger(Logger.level() | [attach_option()]) :: :ok | {:error, :already_exists}
```

sorentwo:
Wonderful, thank you! 💛

PR #555, opened 2021-09-23
Downgrade telemetry

leandrocp:
Allow a broader `telemetry` version to avoid blocking the update of Oban along with other libs that still require `~> 0.4` only.

Such change should be safe considering there's no code changes https://github.com/beam-telemetry/telemetry/blob/main/CHANGELOG.md#100 and it seems to be a common pattern as you can see on elixir-ecto/ecto_sql@7209cce and beam-telemetry/telemetry_metrics@ddc41e5


sorentwo:
Thanks. This has caused some trouble recently!

edennis:
Any chance of a version bump with this commit?

sorentwo:
@edennis Yes indeed. Patch release coming today.

PR #558, opened 2021-09-24
Set logger level on insert unique

### LINKED ISSUE #556: pg_try_advisory_xact_lock, log: false still logs
I can't seem to prevent this log, even with log: false.  Is there any workaround?

https://github.com/sorentwo/oban/blob/2054f8f25412f4174aeae7a0f80b06dc7858092c/lib/oban/query.ex#L213

v0idpwn:
Should fix #556.


sorentwo:
Thanks @v0idpwn.  I've replied to the issue directly. Whatever the issue is, this doesn't look like the fix for it.

Issue #553, opened 2021-09-21
`Oban.pause_queue/2` and related functions always return `:ok` even if queues don't exist


vereis:
### Is your feature request related to a problem? Please describe.

The other day, we were working against a clone of one of our production databases in a staging environment. Because we handle a bunch of stuff via `Oban` that we absolutely _did not want to run_, we did an `Oban.pause_queue(queue: :default)` and saw it return `:ok`.

A few days later, we got reports and complaints, because it appears the queue was not in fact paused. Upon some investigation, this was likely due to `Oban.pause_queue(queue: :non_existing_queue_name)` always returning `:ok` making it susceptible to typos.

While we usually have `Oban.Pro` and `Oban.Web`, we didn't have the dashboard available in our staging environment 😢

### Describe the Solution You'd Like

I see you have a function in all related functions that validates `opts` passed in. I'm wondering if we could simply add the following check:

```elixir
defp validate_queue_opts!({:queue, queue}) do
  cond do
    (is_atom(queue) and not is_nil(queue)) or (is_binary(queue) and byte_size(queue) > 0) ->
          raise ArgumentError,
            "expected :queue to be a binary or atom (except `nil`), got: #{inspect(queue)}"

      # Either checking the provided config, or checking Oban.Registry
      is_nil(Registry.whereis(Oban, {:producer, to_string(queue)})) ->
          raise ArgumentError,
            "provided :queue does not exist, got: #{inspect(queue)}"

      true ->
        :ok
  end
end
```

### Describe Alternatives You've Considered

This could be a separate API; or perhaps, instead of returning `:ok` after queue manipulation, you can return a `queue_state()` instead? If so, given that `Oban.check_queue/2` crashes when provided an invalid queue, the same type of crash would happen following calling `Oban.pause_queue/2` (or related function) with an invalid queue name as well.

Perhaps there is a better way of doing this?

### Additional Context

N/A

sorentwo:
There's a subtle but fundamental difference between `check_queue` and the other `_queue` functions. `check_queue` is only run locally, meaning you only get the queue's state on the local node. The other functions are either local or global. The command is executed through pub-sub in a "fire and forget" mode in the default global mode. Thus, affected queues/nodes don't respond to let the caller know if it worked or their state.

In a multi-node cluster it's possible that a couple of nodes are running queue `:a` and others are running queue `:b`, and we must be able to pause `:b` when making the call to nodes running `:a`.

[...]

PR #559, opened 2021-09-25
Validate that a queue exists when checking/controlling queues while `local_only: true` is provided

vereis:
Hi, this should be a ticket that resolves the issue I made [here](https://github.com/sorentwo/oban/issues/553)

Not sure if this is what you had intended 😅 I'm not too familiar with trying to contribute to OSS so I hope this is okay!

Eager to hear any feedback!

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -487,6 +487,10 @@ defmodule Oban do
   * `:queue` - a string or atom specifying the queue to pause, required
   * `:local_only` - whether the queue will be paused only on the local node, default: `false`
 
+  Note: by default, Oban does not verify that the given queue exists unless `:local_only`
+  is set to `true` as even if the queue does not exist locally, it might be running on
+  another node.
+
   ## Example
 
   Pause the default queue:
@@ -503,6 +507,7 @@ defmodule Oban do
   @spec pause_queue(name(), opts :: [queue_option()]) :: :ok
   def pause_queue(name \\ __MODULE__, [_ | _] = opts) do
     validate_queue_opts!(opts, [:queue, :local_only])
+    validate_queue_exists!(name, opts)
 
     conf = config(name)
     data = %{action: :pause, queue: opts[:queue], ident: scope_signal(conf, opts)}
@@ -517,6 +522,10 @@ defmodule Oban do
 
   * `:queue` - a string or atom specifying the queue to resume, required
   * `:local_only` - whether the queue will be resumed only on the local node, default: `false`
+
+  Note: by default, Oban does not verify that the given queue exists unless `:local_only`
+  is set to `true` as even if the queue does not exist locally, it might be running on
+  another node.
 
   ## Example
 
@@ -534,6 +543,7 @@ defmodule Oban do
   @spec resume_queue(name(), opts :: [queue_option()]) :: :ok
   def resume_queue(name \\ __MODULE__, [_ | _] = opts) do
     validate_queue_opts!(opts, [:queue, :local_only])
+    validate_queue_exists!(name, opts)
 
     conf = config(name)
     data = %{action: :resume, queue: opts[:queue], ident: scope_signal(conf, opts)}
@@ -551,6 +561,10 @@ defmodule Oban do
   * `:local_only` — whether the queue will be scaled only on the local node, default: `false`
 
   In addition, all engine-specific queue options are passed along after validation.
+
+  Note: by default, Oban does not verify that the given queue exists unless `:local_only`
+  is set to `true` as even if the queue does not exist locally, it might be running on
+  another node.
 
   ## Example
 
@@ -576,6 +590,7 @@ defmodule Oban do
 
     validate_queue_opts!(opts, [:queue, :local_only])
     validate_engine_meta!(conf, opts)
+    validate_queue_exists!(name, opts)
 
     data =
       opts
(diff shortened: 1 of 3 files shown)


lib/oban.ex
```diff
@@ -840,4 +861,16 @@ defmodule Oban do
...
+
+  defp maybe_validate_queue_exists!(name, opts) do
+    if Keyword.get(opts, :local_only) do
+      validate_queue_exists!(name, opts[:queue])
+    end
+  end
+
+  defp validate_queue_exists!(name, queue) do
+    unless is_pid(Registry.whereis(name, {:producer, to_string(queue)})) do
+      raise ArgumentError, "expected queue does not exist locally"
```
sorentwo, line 873: It would be nice to include the queue name in the message:

```suggestion
      raise ArgumentError, "queue #{inspect(queue)} does not exist locally"
```

test/integration/controlling_test.exs
```diff
@@ -80,6 +80,18 @@ defmodule Oban.Integration.ControllingTest do
...
+    test "trying to stop queues that don't exist" do
+      name = start_supervised_oban!(queues: [])
+
+      # It is only possible to validate if queues exist if `local_only: true` is
+      # provided
+      assert :ok = Oban.pause_queue(name, queue: :alpha)
+
+      assert_raise ArgumentError, fn ->
+        Oban.pause_queue(name, queue: :alpha, local_only: true)
+      end
```
sorentwo, line 93: With `assert_raise/3` we don't need the comment to explain it:

```suggestion
      assert :ok = Oban.pause_queue(name, queue: :alpha)

      assert_raise ArgumentError, "queue :alpha does not exist locally", fn ->
        Oban.pause_queue(name, queue: :alpha, local_only: true)
      end
```

lib/oban.ex
```diff
@@ -840,4 +861,16 @@ defmodule Oban do
...
     with {:error, error} <- conf.engine.init(conf, opts), do: raise(error)
   end
+
+  defp maybe_validate_queue_exists!(name, opts) do
+    if Keyword.get(opts, :local_only) do
+      validate_queue_exists!(name, opts[:queue])
+    end
+  end
+
+  defp validate_queue_exists!(name, queue) do
```
sorentwo, line 871: It's ok to use a single function and omit the `maybe_` prefixed one.

test/integration/controlling_test.exs
```diff
@@ -146,6 +158,30 @@ defmodule Oban.Integration.ControllingTest do
         assert %{paused: true} = Oban.check_queue(name2, queue: :alpha)
       end)
     end
+
+    test "trying to resume queues that don't exist" do
+      name = start_supervised_oban!(queues: [])
+
+      # It is only possible to validate if queues exist if `local_only: true` is
```
sorentwo, line 165: Same change here, to avoid the comment.

test/integration/controlling_test.exs
```diff
@@ -146,6 +158,30 @@ defmodule Oban.Integration.ControllingTest do
...
+
+      assert_raise ArgumentError, fn ->
+        Oban.scale_queue(name, queue: :alpha, local_only: true)
+      end
+    end
+
+    test "trying to pause queues that don't exist" do
+      name = start_supervised_oban!(queues: [])
+
+      # It is only possible to validate if queues exist if `local_only: true` is
```
sorentwo, line 177: And here too.

sorentwo (changes requested):
Thanks for taking a shot at this. The approach looks great. I've left a few nitpicks and suggestions, but it's nearly there.

test/integration/controlling_test.exs
vereis: 🤯 I never knew this was a thing, thanks so much!

lib/oban.ex
vereis: Just gave this a shot, I ended up going for an `if` despite breaking consistency with the `unless ..., do: raise ...` elsewhere because it was a little hard to grok at a glance.

Let me know if you'd prefer this in some other way!

vereis:
@sorentwo just gave the requested changes a shot, thanks for taking a look so quickly 🙏

lib/oban.ex
```diff
@@ -840,4 +861,13 @@ defmodule Oban do
 
     with {:error, error} <- conf.engine.init(conf, opts), do: raise(error)
   end
+
+  defp validate_queue_exists!(name, opts) do
+    queue = opts[:queue]
+    local_only = opts[:local_only]
+
+    if local_only && is_nil(Registry.whereis(name, {:producer, to_string(queue)})) do
```
sorentwo, line 869: These are both booleans, the `and` version should work:

```suggestion
    if local_only and is_nil(Registry.whereis(name, {:producer, to_string(queue)})) do
```

sorentwo (approved):
The changes all look great. Thanks for applying the suggestions!

lib/oban.ex
vereis: Wait, Isn't it optional to pass this in? In that case we should at least default the opt to false right?

That's why I did the &&

sorentwo: Yes, my mistake. This is still hit if `local_only` is `nil`.

sorentwo:
Reverted and fixed locally. Thanks again 💛

PR #582, opened 2021-11-03
Fix error when a job exits with something other than a stacktrace

derekkraan:
We have seen that sometimes Oban is unable to process certain `DOWN` messages. It crashes on a line calling `Exception.format` because the "stacktrace" that we save here is not always a stacktrace after all (see the added test for example).

From https://hexdocs.pm/elixir/Exception.html#format/3, if `kind` is `{:EXIT, pid}`, then it doesn't format the stacktrace.

I'm reasonably certain that this is the correct behaviour but curious what your thoughts are.

### DIFF
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -170,10 +170,16 @@ defmodule Oban.Queue.Executor do
   def emit_event(%__MODULE__{state: :failure} = exec) do
     measurements = %{duration: exec.duration, queue_time: exec.queue_time}
 
+    kind =
+      case exec.kind do
+        {:EXIT, _pid} -> :exit
+        kind when kind in [:exit, :throw, :error] -> kind
+      end
+
     meta =
       Map.merge(exec.meta, %{
         job: exec.job,
-        kind: exec.kind,
+        kind: kind,
         error: exec.error,
         reason: exec.error,
         stacktrace: exec.stacktrace,
--- a/lib/oban/queue/producer.ex
+++ b/lib/oban/queue/producer.ex
@@ -87,7 +87,7 @@ defmodule Oban.Queue.Producer do
 
   # This message is only received when the job's task doesn't exit cleanly. This should be rare,
   # but it can happen when nested processes crash.
-  def handle_info({:DOWN, ref, :process, _pid, reason}, %State{} = state) do
+  def handle_info({:DOWN, ref, :process, pid, reason}, %State{} = state) do
     {_pid, exec} = Map.get(state.running, ref)
 
     {error, stack} =
@@ -100,7 +100,7 @@ defmodule Oban.Queue.Producer do
     # block the producer while awaiting a retry.
     Task.Supervisor.async_nolink(state.foreman, fn ->
       Breaker.with_retry(fn ->
-        %{exec | kind: :error, error: error, stacktrace: stack, state: :failure}
+        %{exec | kind: {:EXIT, pid}, error: error, stacktrace: stack, state: :failure}
         |> Executor.record_finished()
         |> Executor.report_finished()
       end)
--- a/test/integration/executing_test.exs
+++ b/test/integration/executing_test.exs
@@ -71,7 +71,8 @@ defmodule Oban.Integration.ExecutingTest do
 
   defp job do
     gen all queue <- member_of(~w(alpha beta gamma delta)),
-            action <- member_of(~w(OK DISCARD ERROR EXIT FAIL KILL SNOOZE TASK_ERROR)),
+            action <-
+              member_of(~w(OK DISCARD ERROR EXIT FAIL KILL SNOOZE TASK_ERROR TASK_EXIT_ERROR)),
             ref <- integer(),
             max_attempts <- integer(1..20),
             timeout <- frequency([{3, constant(0)}, {1, constant(100)}]),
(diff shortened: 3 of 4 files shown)


derekkraan:
CI is failing on cyclomatic complexity being greater than 10 of a function in `test/support/worker.ex`. I can refactor this function but disabling this check for this function is also an option, what do you think @sorentwo ?

test/support/worker.ex
```diff
@@ -61,6 +61,13 @@ defmodule Oban.Integration.Worker do
         |> Task.await()
 
         :ok
+
+      "TASK_EXIT_ERROR" ->
```
MvanDiemen, line 65: Where does this one come from?

MvanDiemen:
Good find that it doesn't generate a stacktrace

lib/oban/queue/producer.ex
```diff
@@ -100,7 +100,7 @@ defmodule Oban.Queue.Producer do
     # block the producer while awaiting a retry.
     Task.Supervisor.async_nolink(state.foreman, fn ->
       Breaker.with_retry(fn ->
-        %{exec | kind: :error, error: error, stacktrace: stack, state: :failure}
+        %{exec | kind: {:EXIT, pid}, error: error, stacktrace: stack, state: :failure}
```
sorentwo, line 103: The `kind` is exposed through telemetry data, and meant to match the values that come from `telemetry.span`. What if we keep this as `:error` but ensure that `stack` is an empty list?

sorentwo:
> CI is failing on cyclomatic complexity being greater than 10 of a function in `test/support/worker.ex`. I can refactor this function but disabling this check for this function is also an option, what do you think @sorentwo ?

👍 I'd rather disable the check (especially for a test module).

lib/oban/queue/producer.ex
derekkraan: Options:
1. kind: `:error`: `"error" => "** (ErlangError) Erlang error: :timeout"`
2. kind: `:exit` (also a valid `kind` according to `telemetry` docs): `"error" => "** (exit) time out"`
3. kind: `{:EXIT, pid}`: `"error" => "** (EXIT from #PID<0.3249.0>) time out" `

I'm not quite sure what happens in Telemetry when you pass `{:EXIT, pid}`.

derekkraan: I looked into this further, `Exception.kind` has type `:error | :exit | :throw | {:EXIT, pid}`, whilst telemetry kind only includes `:error | :exit | :throw`. I think it's a bit weird. We are using `kind` both in calls to the `Exception` module and in calls to telemetry.

I added a little case statement to convert from `Exception.kind` to the `metadata.kind` that `:telemetry` is expecting. This gives us the best error message, and still conforms to telemetry's spec.

Let me know if you'd still prefer `:exit`.

lib/oban/queue/executor.ex
```diff
@@ -170,10 +170,16 @@ defmodule Oban.Queue.Executor do
   def emit_event(%__MODULE__{state: :failure} = exec) do
     measurements = %{duration: exec.duration, queue_time: exec.queue_time}
 
+    kind =
+      case exec.kind do
```
sorentwo, line 174: Initially, I wasn't sure about this. Now that I see it, I appreciate the explicitness. Nice work 👍

test/integration/executing_test.exs
```diff
@@ -71,7 +71,8 @@ defmodule Oban.Integration.ExecutingTest do
 
   defp job do
     gen all queue <- member_of(~w(alpha beta gamma delta)),
-            action <- member_of(~w(OK DISCARD ERROR EXIT FAIL KILL SNOOZE TASK_ERROR)),
+            action <-
+              member_of(~w(OK DISCARD ERROR EXIT FAIL KILL SNOOZE TASK_ERROR TASK_EXIT_ERROR)),
```
sorentwo, line 75: The other action is `EXIT`, not `EXIT_ERROR` (and this keeps it on a single line still).

```suggestion
            action <- member_of(~w(OK DISCARD ERROR EXIT FAIL KILL SNOOZE TASK_ERROR TASK_EXIT),
```

test/support/worker.ex
```diff
@@ -61,6 +62,13 @@ defmodule Oban.Integration.Worker do
...
```
sorentwo, line 66:
```suggestion
      "TASK_EXIT" ->
```

test/integration/executing_test.exs
```diff
@@ -71,7 +71,7 @@ defmodule Oban.Integration.ExecutingTest do
 
   defp job do
     gen all queue <- member_of(~w(alpha beta gamma delta)),
-            action <- member_of(~w(OK DISCARD ERROR EXIT FAIL KILL SNOOZE TASK_ERROR)),
+            action <- member_of(~w(OK DISCARD ERROR EXIT FAIL KILL SNOOZE TASK_ERROR TASK_EXIT),
```
sorentwo, line 74: Invariably I mess up any suggestion I commit...

```suggestion
            action <- member_of(~w(OK DISCARD ERROR EXIT FAIL KILL SNOOZE TASK_ERROR TASK_EXIT)),
```

derekkraan:
@sorentwo not sure if you intended to close this PR?

sorentwo:
Nope. That was a side effect of changing the primary branch to `main`. Please re-open!

derekkraan:
I think I have to open a new PR. I'll do that tomorrow.

PR #583, opened 2021-11-03
Slightly better naming in README example

matrinox:
Quick change, just thought b_job could be misinterpreted so chose a more suitable name


sorentwo:
Thanks! 💛

Issue #597, opened 2021-12-09
Crontab Reboot with timestamp doesn't create job immediately


bruteforcecat:
### Environment

- Oban Version 2.10.1
- PostgreSQL Version 11.5
- Elixir & Erlang/OTP Versions (`elixir --version`)  1.10.4/OTP 23

### Current Behavior

Set crontab with reboot and timezone will not result in creating job immediately. 

### Expected Behavior

reboot crontab should ignore the "global" timezone and create job immediately.

I crated an minimal demo repo for this issue. https://github.com/bruteforcecat/oban_reboot_not_working

PR #598, opened 2021-12-09
fix cron reboot with timezone not working properly

bruteforcecat:
Hello. Thank you very much for this amazing library.

I recently found an issue where reboot crontab will not create job immediately if timezone is set(The timezone is set for other crontab). Related issue is reported here https://github.com/sorentwo/oban/issues/597

This PR is an attempt to fix this. I tried to keep minimal change but I guess the biggest change is to move timezone from `Cron.State` to `Cron.Expression`. One of the reason I do this is to allow another future PR to allow user set individual crontab expression timezone which provide greater flexibility.

I apologize in advance if the change in this PR is not good. But I am willing to update this PR upon any good feedback.

### DIFF
--- a/lib/oban/cron/expression.ex
+++ b/lib/oban/cron/expression.ex
@@ -64,6 +64,7 @@ defmodule Oban.Cron.Expression do
 
   The parser can handle common expressions that use minutes, hours, days, months and weekdays,
   along with ranges and steps. It also supports common extensions, also called nicknames.
+  The second argument, now is only used for @reboot common expression.
 
   Raises an `ArgumentError` if the expression cannot be parsed.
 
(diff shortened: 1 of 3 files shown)


sorentwo (changes requested):
Thanks for putting so much work into this solution. The single timezone limitation is intentional, as one of the differentiators with Pro's dynamic cron.

I'd prefer not to change the `expression.ex` module so radically. What if the change is limited to `parse!`? Something like this:

```elixir
# Set a default
def parse!(expression, now \\ DateTime.utc_now())

# Ignore _now for the other functions
def parse!("@annually", _now), do: parse!("0 0 1 1 *")

def parse!("@reboot", now) do
  # Use the injected `now` here
end
```

bruteforcecat:
ahh I didn't understand this is Pro feature. I adjusted the PR according to your comment. Please let me know if there is other issue 🙏🏽

sorentwo:
@bruteforcecat Thank you for bearing with me on those changes. Something about passing the timezone into the expression module for parsing didn't sit right with me. It made an already non-deterministic function clause even stranger, and it tainted the entire parse API for a particular edge case.

After thinking about this more I opted for a simpler approach where we track `@reboot` specifically and then discard those entries after the first pass. You can see my solution in 7508d87.

Thanks for reporting the issue and for all of your time working out a good solution with this PR.

PR #600, opened 2021-12-13
adds listen to the documentation

adkron:
The documentation `insert_and_listen/1` doesn't work as it stands alone.
To make it work out of the box, it needs to call `listen/2`. Since the
function is `insert_and_listen`, I thought this was an acceptable solution.

Co-authored-by: Chad Fennel <chad@binarynoggin.com>


sorentwo:
Nice catch! Thanks 💛

PR #613, opened 2022-01-05
Document default value for uniqueness period

crbelaus:
If we don't specify a uniqueness period, Oban will check [60 seconds by default](https://github.com/sorentwo/oban/blob/a0018f11954a898dfe6cd8260a3a8a085f4f0921/lib/oban/job.ex#L261).

The documentation says that "you should always specify a period" but does not mention the default value. This is a small pull request to fix that part of the documentation.


sorentwo:
Thank you! 💛

PR #627, opened 2022-01-31
Fix unique job period query criteria

opsb:


lib/oban/queue/basic_engine.ex
```diff
@@ -382,7 +382,7 @@ defmodule Oban.Queue.BasicEngine do
   defp since_period(query, period) do
     since = DateTime.add(utc_now(), period * -1, :second)
 
-    where(query, [j], j.inserted_at > ^since)
+    where(query, [j], j.inserted_at >= ^since)
```
sorentwo, line 385: This change seems entirely valid, but I'm curious how frequently it comes up outside of a test environment 🤔

sorentwo (commented):
Thanks for tracking this down and opening a PR! All it needs is a matching test (take a look in `uniqueness.exs`).

sorentwo:
@opsb Glad to hear you're sorted. I'm merging this as-is and considering the change covered by existing property tests.

PR #629, opened 2022-02-04
fix: better error message for missing plugin

polvalente:
This PR aims to provide a better error message for when a given plugin is missing.
My use case is that there was an Oban.Pro plugin being used without oban_pro being added as a dependency.

### DIFF
--- a/lib/oban/config.ex
+++ b/lib/oban/config.ex
@@ -159,11 +159,11 @@ defmodule Oban.Config do
   end
 
   defp validate_opt!({:plugins, plugins}) do
-    unless is_list(plugins) and Enum.all?(plugins, &valid_plugin?/1) do
-      raise ArgumentError,
-            "expected :plugins to be a list of modules or {module, keyword} tuples " <>
-              ", got: #{inspect(plugins)}"
+    unless is_list(plugins) do
+      raise ArgumentError, "expected :plugins to be a list, got #{inspect(plugins)}"
     end
+
+    Enum.each(plugins, &validate_plugin/1)
   end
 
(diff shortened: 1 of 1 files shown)


sorentwo (commented):
Great suggestion! This bites people every once in a while. What do you think about changing this from `valid_plugin?` to `validate_plugin!` and calling it with `Enum.each`?

Something like this:

```elixir
  defp validate_opt!({:plugins, plugins}) do
    unless is_list(plugins) do
      raise ArgumentError, "expected :plugins to be a list, got: #{inspect(plugins)}"
    end
    
    Enum.each(plugins, &validate_plugin!/1)
  end
```

Then do the separate "plugin is a loaded atom", "plugin exports a function", "opts is a keyword" checks in there?

polvalente:
> Great suggestion! This bites people every once in a while. What do you think about changing this from `valid_plugin?` to `validate_plugin!` and calling it with `Enum.each`?
> 
> Something like this:
> 
> ```elixir
>   defp validate_opt!({:plugins, plugins}) do
>     unless is_list(plugins) do
>       raise ArgumentError, "expected :plugins to be a list, got: #{inspect(plugins)}"
>     end
>     
>     Enum.each(plugins, &validate_plugin!/1)
>   end
> ```
> 
> Then do the separate "plugin is a loaded atom", "plugin exports a function", "opts is a keyword" checks in there?

Sounds good! I'll make these changes and push then as soon as possible!

sorentwo (approved):
Fantastic! Thanks for the changes.

sorentwo:
Thanks again 💛

PR #637, opened 2022-02-14
Update v2.11 migration docs: add disable_migration_lock setting

mtunski:
As per the docs (https://hexdocs.pm/ecto_sql/Ecto.Migration.html#index/3-adding-dropping-indexes-concurrently), it's necessary to set both `disable_ddl_transaction` **and** `disable_migration_lock` to enable adding/dropping indexes concurrently. The latter was missing in the docs.


sorentwo:
Thanks!

PR #657, opened 2022-03-04
Add note about uniqueness and Oban.insert_all/2

antedeguemon:
Hey folks! 

README already has multiple mentions about `insert_all/2` not supporting uniqueness configuration, but there is no mention of that in the `Unique Jobs` section.

This PR adds one more note about it - this time at the `Unique Jobs` section. 

I hope it saves some debug time from people! 😄


sorentwo:
Thank you!

Issue #632, opened 2022-02-09
Improve testing experience for Oban and its plugins


milmazz:
### Is your feature request related to a problem? Please describe.

This is not an actual problem because you can now find workarounds to test your Oban configuration and some of the plugin configurations. But I'm sure we can improve the experience around these things.

For example, to test your Oban configuration, you can do something like:

```elixir
defmodule MyApp.ObanConfigTest do
  use ExUnit.Case, async: true

 test "check oban configuration" do
    [config, prod] =
      Enum.map(["config/config.exs", "config/prod.exs"], fn path ->
        path
        |> Config.Reader.read!(imports: [], env: :prod, target: :host)
        |> get_in([:core, Oban])
      end)

    assert %Oban.Config{plugins: _plugins} = config |> Keyword.merge(prod) |> Oban.Config.new()
  end
end
```

But, for the previous unit test to make sense, you need to know a bit of the internal implementation of Oban, which is the use of the inner function `Oban.Config.new/1`.

It is important to note that the `Oban.Config.new/1` function doesn't validate the internal configuration for the plugins; it only validates that the given plugin is an atom, is loaded, the plugin in question exports an `init/1` function, and also the given `opts` is a keyword list. 

So, if you want to go further and validate the given options to the `Oban.Plugins.Cron` plugin, for example, you need to know a bit about the internal implementation and use a combination of `Oban.Plugins.Cron.validate!/1` and `Oban.Plugins.Cron.init/1`. Then, finally, you need to assert that the returned `{:ok, state, _}` is what you expect.

I like to validate the cron plugin configuration because it constantly evolves, and it's easy to catch typos in the cron worker module names also validating the cron expressions by doing the following:

```elixir
assert :ok = Oban.Plugins.Cron.validate!(cron_opts)

assert {:ok, _state, {:continue, :start}} = Oban.Plugins.Cron.init(cron_opts)
```

But again, here I'm, relying on a function for internal use. Still, I think it is worth taking the risk of using a non-documented public function in this case; you gain more in your daily routine because it will allow you to catch errors in your Oban Configuration as soon as possible.

### Describe the Solution You'd Like

A public test helper to validate the Oban configuration, including plugins. I think it's also worth normalizing the _plugins_ under a contract or behavior; Oban now enforces defining an `init/1` function for plugins, but I think the behaviour should also include a `validate!/1` function. At least the `validate!/1` should help with the _testing experience_.

[...]

PR #659, opened 2022-03-04
Introduce Plugin behaviour and expand plugin option validation

sorentwo:
This is a WIP to address #632.

- [x] Add `Plugin` behaviour
- [x] Document new behaviour usage and callbacks
- [x] Implement behaviour and validation for Cron
- [x] Implement behaviour and validation for Gossip
- [x] Implement behaviour and validation for Lifeline
- [x] Implement behaviour and validation for Pruner
- [x] Implement behaviour and validation for Reindexer
- [ ] Implement behaviour and validation for Repeater
- [ ] Implement behaviour and validation for Stager
- [ ] Consider Plugin behaviour when verifying plugins in Config

/cc @milmazz

### DIFF
--- a/lib/oban/plugins/gossip.ex
+++ b/lib/oban/plugins/gossip.ex
@@ -35,11 +35,13 @@ defmodule Oban.Plugins.Gossip do
   * `:gossip_count` - the number of queues that had activity broadcasted
   """
 
+  @behaviour Oban.Plugin
+
   use GenServer
 
-  alias Oban.{Config, Notifier}
+  alias Oban.{Notifier, Plugin}
 
-  @type option :: {:conf, Config.t()} | {:name, GenServer.name()} | {:interval, pos_integer()}
+  @type option :: Plugin.option() | {:interval, pos_integer()}
 
   defmodule State do
     @moduledoc false
@@ -47,7 +49,7 @@ defmodule Oban.Plugins.Gossip do
     defstruct [:conf, :name, :timer, interval: :timer.seconds(1)]
   end
 
-  @doc false
+  @impl Plugin
   @spec start_link([option()]) :: GenServer.on_start()
   def start_link(opts) do
     GenServer.start_link(__MODULE__, opts, name: opts[:name])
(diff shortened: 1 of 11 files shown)


lib/oban/plugins/reindexer.ex
```diff
@@ -93,6 +110,24 @@ defmodule Oban.Plugins.Reindexer do
...
+    error in [ArgumentError] -> {:error, error}
+  end
+
+  defp validate_timezone(timezone) do
+    if is_binary(timezone) and match?({:ok, _}, DateTime.now(timezone)) do
+      :ok
+    else
+      {:error, "expected :timezone to be a known timezone, got: #{inspect(timezone)}"}
+    end
+  end
```
milmazz, line 129: I think this validator helper list will grow sooner than later, and some of them will be duplicated in the codebase; for example, the implementation of `validate_timezone` it's the same or very similar to the one found in `validate_opt({:timezone, timezone})`. So, what do you think about wrapping these `validate_*` helpers under a module like `Oban.ConfigValidator` or something similar? Again, this reminds me about `Ecto.Changeset` and their `validate_*` functions.

milmazz (commented):
First, thank you for this effort. I like the way this PR is going. I left a couple of suggestions or comments. HTH

lib/oban/plugins/gossip.ex
```diff
@@ -65,6 +67,11 @@ defmodule Oban.Plugins.Gossip do
     {:ok, state}
   end
 
+  @impl Plugin
+  def validate(_opts) do
+    :ok
```
milmazz, line 72: This should validate the `:interval` option, no?

lib/oban/plugins/lifeline.ex
```diff
@@ -68,17 +69,33 @@ defmodule Oban.Plugins.Lifeline do
...
-    {:ok, schedule_rescue(state)}
+    {:ok, state}
+  end
+
+  @impl Plugin
+  def validate(opts) do
+    Plugin.validate(opts, fn
+      {:conf, _} -> :ok
+      {:name, _} -> :ok
+      {:interval, interval} -> validate_timeout(interval)
```
milmazz, line 95: FYI: this option is missing in the `@moduledoc`

lib/oban/plugins/reindexer.ex
sorentwo: Great point. I'll keep that in mind for a subsequent refactor, as I aim to expose an exception-free validation helper on Config as well.

Issue #664, opened 2022-03-09
Cannot pattern match unique when using all_enqueued


AntoineAugusti:
### Environment

- Oban Version `2.10.1`
- PostgreSQL Version `13.4`
- Elixir & Erlang/OTP Versions `Elixir 1.13.3 (compiled with Erlang/OTP 24)`

### Current Behavior

When inserting a job with an override unique configuration, `all_enqueued` returns a `unique: nil`

```elixir
%{email: "brewster@example.com"}
|> MyApp.Mailer.new(unique: [period: 300, fields: [:queue, :worker])
|> Oban.insert()

# In tests
assert [%Oban.job{unique: unique}] = all_enqueued worker: MyApp.Mailer
refute is_nil(unique)
```

### Expected Behavior

I was expecting to be able to pattern match on all fields on the `Oban.Job` using `all_enqueued`. I know that we can use keywords to get specific workers, queues etc but `unique` seems like a good candidate for pattern matching.

sorentwo:
The issue here is that `unique` is a virtual field. It isn't persisted in the database and therefore can't be checked in tests.

If you'd really like to verify that `unique` options are present, you can check the changeset returned from `Worker.new` like this:

```elixir
assert %{changes: %{unique: _}} = MyApp.Mailer.new(unique: [period: 300, fields: [:queue, :worker])
```

PR #665, opened 2022-03-09
Reproduction: cannot pattern match unique when using all_enqueued 

AntoineAugusti:
Demonstrates the issue described in https://github.com/sorentwo/oban/issues/664


sorentwo:
Thanks for the reproduction. I'm closing this for the reasons outlined in #664.

PR #667, opened 2022-03-11
Fix migrations: allow any prefix name.

azizk:
Using reserved names such as `case` isn't possible,
because it results in a PSQL syntax error.

Using `inspect` will always print a double quoted string,
thereby fixing this issue.

### DIFF
--- a/lib/oban/migrations.ex
+++ b/lib/oban/migrations.ex
@@ -158,13 +158,15 @@ defmodule Oban.Migrations do
 
   @doc false
   def migrated_version(repo, prefix) do
+    escaped_prefix = String.replace(prefix, "'", "\\'")
+
     query = """
     SELECT description
     FROM pg_class
     LEFT JOIN pg_description ON pg_description.objoid = pg_class.oid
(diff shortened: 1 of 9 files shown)


sorentwo:
@azizk Thanks for the PR! I'm not convinced that a schema/prefix _should_ use a reserved word, but who am I to tell people how to design their databases 🤷

This PR touches a lot of code, and I'll need a little time to review it.

azizk:
Well, there are quite a lot of (semi-)reserved words. :slightly_smiling_face:
It think it's good practice anyway to escape stuff that you're interpolating into a plain text query. :+1:

Thanks for reviewing the PR! It's mostly the same modification everywhere. The one place I wasn't really sure about was at `add :state, :"#{quoted_prefix}.oban_job_state"` and whether to use a quoted prefix here or not.

azizk:
Besides, some may legitimately want to use punctuation characters that can appear in web addresses or customer company names. I just updated the commit message.

sorentwo:
@azizk The pure SQL migration change isn't happening, so I made a few tweaks to this to limit duplication and merged it into the main manually. Thanks! 💛

PR #668, opened 2022-03-15
Remove outdated mention of producer circuit breakers

ethangunderson:
Circuit breakers were removed in 4c17c8d1bebdffb61dafde1bd8d62055e6e6da7d but the reference to them is still in the Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,9 +87,6 @@ orphaned due to crashes.
   independently at runtime locally or across _all_ running nodes (even in
   environments like Heroku, without distributed Erlang).
 
-- **Resilient Queues** — Failing queries won't crash the entire supervision tree,
-  instead they trip a circuit breaker and will be retried again in the future.
-
 - **Job Canceling** — Jobs can be canceled in the middle of execution regardless
   of which node they are running on. This stops the job at once and flags it as
   `discarded`.


README.md
```diff
@@ -87,9 +87,6 @@ orphaned due to crashes.
   independently at runtime locally or across _all_ running nodes (even in
   environments like Heroku, without distributed Erlang).
 
-- **Resilient Queues** — Failing queries won't crash the entire supervision tree,
-  instead they trip a circuit breaker and will be retried again in the future.
-
```
sorentwo, line 92: This is still true, but it is done with a fine-grained retry/backoff mechanism rather than circuit breakers. A few examples:

https://github.com/sorentwo/oban/blob/main/lib/oban/queue/executor.ex#L89
https://github.com/sorentwo/oban/blob/main/lib/oban/queue/producer.ex#L108

Pro's `SmartEngine` wraps all of the queue operations in a similar retry/backoff.

Would you mind making the paragraph more accurate, rather than removing it?

sorentwo:
@ethangunderson Updated on main, thanks for the PR/prompt!

ethangunderson:
Ah, thank you for picking this up!

PR #677, opened 2022-03-29
Introduce Testing Guides

sorentwo:
Extract testing hints and suggestions out of the README and into a set of testing guides.

- Move setup from README into a testing introduction
- Add a guide on testing configuration
- Add a guide on testing workers at the unit level
- Add a guide on testing queues at the integration level

/cc @sorenone


guides/testing/testing_workers.md
```diff
@@ -0,0 +1,105 @@
...
+some number of milliseconds:
+
+```elixir
+test "allowing a multiple of the attempt as job timeout" do
+  assert 1000 == MyWorker.timeout(%Oban.Job{attempt: 1})
+  assert 2000 == MyWorker.timeout(%Oban.Job{attempt: 2})
+end
+```
+
+Jobs are just Ecto schemas, and therefore structs. There isn't anything magical
```
milmazz, line 103:
```suggestion
Jobs are Ecto schemas, and therefore structs. There isn't anything magical
```

guides/testing/testing_workers.md
```diff
@@ -0,0 +1,105 @@
...
+* That a custom `new/1,2` callback works properly
+
+If all of the assertions pass, then you'll get the result of `perform/1` for you
+to make additional assertions on.
+
+## Testing Other Callbacks
+
+You may wish to test less-frequently used worker callbacks such as `backoff/1`
+and `timeout/1`, but those callbacks don't have dedicated testing helpers.
+Never fear, it's easy enough to build a job struct and test callbacks directly!
```
milmazz, line 80:
```suggestion
Never fear, it's enough to build a job struct and test callbacks directly!
```

milmazz (approved):
Thanks for taking care of this. Excellent addition 🎉

guides/testing/testing_config.md
```diff
@@ -0,0 +1,80 @@
...
+
+Your app's Oban config is extremely important and it deserves testing!
+
+## Testing Production Config
+
+Within the test environment, apps typically disable queues and plugins for more
+control and to prevent sandbox interference. That means only a subset of the
+full Oban config is verified when your application boots for testing.
+
+The `Oban.Config.validate/1` helper is used internally when the config
```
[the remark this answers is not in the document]
sorentwo: > Is it really beneficial to validate configs in our tests?

In my opinion, absolutely. Test configuration and production configuration may be wildly different.

> As for whether it's configured correctly, it seems not to provide any benefits to test if queue :a is paused or not.

Consider the config for plugins with complex options, like Cron or DynamicQueues. Alternatively, queues can have involved configuration that combines global limits, rate limits, partitioning, etc.

> As for validation, I personally would prefer Oban to refuse to start when the configs are invalid, so my app wouldn't start on production and I can and have to see why in the logs.

It already does this. If the configuration is invalid the supervisor won't boot at all. That is _exactly_ the situation this aims to prevent in production. Not everybody has health checks before the app boots, and multiple people have reported a misconfiguration taking their app down.

guides/testing/testing_config.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Testing Plugin Config
+
+Validation is especially helpful for plugins that have complex configuration,
+e.g. `Cron` or the `Dynamic*` plugins from Oban Pro. As of Oban v2.12.0 all
+plugins implement the `c:Oban.Plugin.validate/1` callback and we can test them
+in isolation as well as through the top level config.
+
+If your plugins are statically defined, then validating them through
+`Oban.Config` is easy and recommended. However, if your config is dynamic, then
```
sorenone, line 48:
```suggestion
`Oban.Config` is easy and recommended. If your config is dynamic, then
```

guides/testing/testing_queues.md
```diff
@@ -0,0 +1,103 @@
+# Testing Queues
+
+Where workers are the primary "unit" of an Oban system, queues are the
+"integration" point between the database and your application. That means to
+test queues and the jobs within them your tests will have to interact with the
```
sorenone, line 5:
```suggestion
test queues and the jobs within them, your tests will have to interact with the
```

guides/testing/testing_queues.md
```diff
@@ -0,0 +1,103 @@
...
+database. To simplify that interaction, reduce boilerplate, and make assertions
+more expressive `Oban.Testing` provides a variety of helpers.
+
+## Asserting Enqueued Jobs
+
+During test runs you don't typically want to execute jobs. Rather, you only need
```
sorenone, line 11:
```suggestion
During test runs you don't typically want to execute jobs. Rather, you need
```

guides/testing/testing_queues.md
```diff
@@ -0,0 +1,103 @@
...
+end
+```
+
+## Asserting Multiple Jobs
+
+Asserting and refuting about a single job isn't always enough. Sometimes you
+need to check for multiple jobs at once, or perform more complex assertions on
+the jobs themselves. In that situation, you can use `all_enqueued` instead.
+
+The first case we'll look at asserts that multiple jobs from the same worker are
```
sorenone, line 50:
```suggestion
The first example we'll look at asserts that multiple jobs from the same worker are
```

guides/testing/testing_workers.md
```diff
@@ -0,0 +1,105 @@
+# Testing Workers
+
+Worker modules are the primary "unit" of an Oban system. You can test a worker's
+callback functions locally, in-process, without touching the database.
```
sorenone, line 4:
```suggestion
Worker modules are the primary "unit" of an Oban system. You can (and should) test a 
worker's callback functions locally, in-process, without touching the database.
```

guides/testing/testing_workers.md
```diff
@@ -0,0 +1,105 @@
...
+
+Most worker callback functions take a single argument: an `Oban.Job` struct. A
+job encapsulates arguments, metadata, and other options. Creating jobs, and
+verifying that they're built correctly, requires some boilerplate—that's where
```
sorenone, line 8:
```suggestion
verifying that they're built correctly, requires some boilerplate...that's where
```

guides/testing/testing_workers.md
```diff
@@ -0,0 +1,105 @@
...
+* That a custom `new/1,2` callback works properly
+
+If all of the assertions pass, then you'll get the result of `perform/1` for you
+to make additional assertions on.
+
+## Testing Other Callbacks
+
+You may wish to test less-frequently used worker callbacks such as `backoff/1`
+and `timeout/1`, but those callbacks don't have dedicated testing helpers.
+Never fear, it's enough to build a job struct and test callbacks directly!
```
sorenone, line 80:
```suggestion
Never fear, it's adequate to build a job struct and test callbacks directly!
```

sorenone (approved):
Well done. Hard fought.

PR #682, opened 2022-04-06
Update URL to the testing guide

harrygr:
The existing URL is stale; use the current URL as appears in the Hex docs.


sorentwo:
Thanks for the PR!

The URL points to an unreleased testing guide: https://github.com/sorentwo/oban/blob/main/guides/testing/testing.md and not the `Oban.Testing` module.

The README points to main, which is 2.12.0-dev and hasn't been published to hex yet. You can find the current docs on hex: https://hexdocs.pm/oban/Oban.html

PR #686, opened 2022-04-13
Fix README formatting

clekstro:
The most recent documentation updates broke links/formatting within the `README`


sorentwo:
Suggesting with fenced code blocks has some issues. Thanks for the fix!

PR #689, opened 2022-04-15
Show workflow for migrating to new Oban version

nathanl:


lib/oban/migrations.ex
```diff
@@ -24,15 +24,34 @@ defmodule Oban.Migrations do
...
+  Open the generated migration in your editor and call the `up` and `down`
+  functions on `Oban.Migrations`, passing a version number:
+
+  ```elixir
+  defmodule MyApp.Repo.Migrations.UpgradeObanToV11 do
+    use Ecto.Migration
+
+    def up, do: Oban.Migrations.up(version: 11)
+
+    def down, do: Oban.Migrations.down(version: 10)
```
sorentwo, line 51: The `down` migration runs every migration down to, and including, the number provided:

```suggestion
    def down, do: Oban.Migrations.down(version: 11)
```

sorentwo (approved):

sorentwo:
Thank you for the clarification. Doc improvements are always welcome 💛

PR #690, opened 2022-04-16
Fix testing guide link

wkirschbaum:
The current testing link points to a non-existing page


sorentwo:
The README is generated from main, which does have the testing guide but hasn't been published to hex yet. If you're using the last published Oban version, then you can find the testing content at [hexdocs.pm/oban](https://hexdocs.pm/oban)

PR #694, opened 2022-04-24
Support flexible use of perform_job with inline testing and with_testing_mode

shanesveller:
_Temporary description:_ 
Tthis branch adds a failing test case for using `perform_job` inside of a `with_testing_mode` block as a minimal reproduction of a test failure I experienced on a new project yesterday. I'm unblocked if I default project-wide to `:manual` testing to preserve the 2.11 behavior, but I thought this was worth surfacing as a relatively intuitive combination of testing techniques.

```
  1) test with_testing_mode/2 with perform_job (Oban.TestingTest)
     test/oban/testing_test.exs:324
     ** (ArgumentError) nil given for `id`. comparison with nil is forbidden as it is unsafe. Instead write a query with is_nil/1, for example: is_nil(s.id)
     code: Testing.with_testing_mode(:manual, fn ->
     stacktrace:
       (ecto 3.7.1) lib/ecto/query/builder/filter.ex:191: Ecto.Query.Builder.Filter.not_nil!/2
       (oban 2.12.0) lib/oban/queue/basic_engine.ex:136: Oban.Queue.BasicEngine.complete_job/2
       (oban 2.12.0) lib/oban/queue/engine.ex:234: anonymous fn/3 in Oban.Queue.Engine.with_span/4
       (telemetry 1.1.0) /home/shane/src/not_me/oban/deps/telemetry/src/telemetry.erl:320: :telemetry.span/3
       (oban 2.12.0) lib/oban/queue/executor.ex:203: Oban.Queue.Executor.ack_event/1
       (oban 2.12.0) lib/oban/queue/executor.ex:197: Oban.Queue.Executor.report_finished/1
       (oban 2.12.0) lib/oban/queue/executor.ex:80: anonymous fn/1 in Oban.Queue.Executor.call/1
       (oban 2.12.0) lib/oban/testing.ex:236: Oban.Testing.perform_job/3
       test/oban/testing_test.exs:326: anonymous fn/0 in Oban.TestingTest."test with_testing_mode/2 with perform_job"/1
       (oban 2.12.0) lib/oban/testing.ex:398: Oban.Testing.with_testing_mode/2
       test/oban/testing_test.exs:325: (test)
```

I'm happy to either stop here if the fix is evident to you now that the test is available, or try to pursue a fix myself. In the case of the latter I will clean up the PR description and commit history afterwards.

### DIFF
--- a/lib/oban/queue/basic_engine.ex
+++ b/lib/oban/queue/basic_engine.ex
@@ -130,6 +130,10 @@ defmodule Oban.Queue.BasicEngine do
   end
 
   @impl Engine
+  def complete_job(%Config{testing: :inline}, %Job{id: nil}) do
+    {0, nil}
+  end
+
   def complete_job(%Config{} = conf, %Job{} = job) do
     Repo.update_all(
       conf,
--- a/lib/oban/testing.ex
+++ b/lib/oban/testing.ex
@@ -122,6 +122,8 @@ defmodule Oban.Testing do
 
   alias Oban.{Config, Job, Queue.Executor, Repo, Worker}
 
+  @type testing_mode :: :inline | :manual
+
   @wait_interval 10
 
   @doc false
@@ -211,11 +213,11 @@ defmodule Oban.Testing do
   @spec perform_job(
           worker :: Worker.t(),
           args :: term(),
-          opts :: [Job.option() | {:repo, module()}]
+          opts :: [Job.option() | {:repo, module()} | {:testing, testing_mode()}]
         ) ::
           Worker.result()
   def perform_job(worker, args, opts) when is_atom(worker) do
-    {conf_opts, opts} = Keyword.split(opts, [:log, :prefix, :repo])
+    {conf_opts, opts} = Keyword.split(opts, [:log, :prefix, :repo, :testing])
 
     opts = Keyword.put_new(opts, :attempt, 1)
 
@@ -228,6 +230,14 @@ defmodule Oban.Testing do
 
     assert_valid_changeset(changeset)
 
+    case Keyword.get(conf_opts, :testing) do
+      mode when mode in [:inline, :manual] ->
+        Process.put(:oban_engine, engine_for_testing_mode(mode))
+
+      _ ->
+        nil
+    end
+
     result =
       conf_opts
       |> Keyword.put(:testing, :inline)
@@ -235,6 +245,8 @@ defmodule Oban.Testing do
       |> Executor.new(create_job(changeset), safe: false)
       |> Executor.call()
       |> Map.fetch!(:result)
+
+    Process.delete(:oban_engine)
 
     assert_valid_result(result)
 
@@ -385,15 +397,9 @@ defmodule Oban.Testing do
       end)
   """
   @doc since: "2.12.0"
-  @spec with_testing_mode(:inline | :manual, (() -> any())) :: any()
+  @spec with_testing_mode(testing_mode(), (() -> any())) :: any()
   def with_testing_mode(mode, fun) when mode in [:manual, :inline] and is_function(fun, 0) do
-    engine =
-      case mode do
-        :manual -> Oban.Queue.BasicEngine
-        :inline -> Oban.Queue.InlineEngine
-      end
-
-    Process.put(:oban_engine, engine)
+    Process.put(:oban_engine, engine_for_testing_mode(mode))
 
     fun.()
   after
@@ -589,4 +595,9 @@ defmodule Oban.Testing do
 
     Changeset.put_change(changeset, field, value)
   end
+
+  defp engine_for_testing_mode(mode)
+  defp engine_for_testing_mode(:manual), do: Oban.Queue.BasicEngine
+  defp engine_for_testing_mode(:inline), do: Oban.Queue.InlineEngine
+  defp engine_for_testing_mode(_unsupported), do: nil
 end
--- a/test/oban/testing_test.exs
+++ b/test/oban/testing_test.exs
@@ -192,6 +192,11 @@ defmodule Oban.TestingTest do
     after
       :telemetry.detach("perform-job-handler")
     end
+
+    test "changing testing mode via opts" do
+      assert :ok = perform_job(Worker, %{ref: 1, action: "OK"}, testing: :inline)
+      assert :ok = perform_job(Worker, %{ref: 1, action: "OK"}, testing: :manual)
+    end
   end
 
   describe "all_enqueued/0,1" do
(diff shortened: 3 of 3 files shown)


sorentwo:
Ah, running perform_job wrapped in manual mode overrides the inline engine that perform_job expects.

You shouldn't have to wrap perform_job anyhow, as it is always "inline." Regardless, it should work in any mode and with any override.

I think the fix here is to set the engine in the process dictionary at the start of perform_job and then undo the change afterwards.

sorentwo:
@shanesveller Are you still interested in implementing the fix yourself?

shanesveller:
I am interested, but I can't currently allocate time for it during my workday so it will probably not be before this weekend.

sorentwo:
No rush here! I only wanted to make sure I didn't step on your toes 🙂

lib/oban/queue/basic_engine.ex
```diff
@@ -130,6 +130,10 @@ defmodule Oban.Queue.BasicEngine do
   end
 
   @impl Engine
+  def complete_job(%Config{testing: :inline}, %Job{id: nil}) do
```
shanesveller, line 133: I'm happy to take any direction on this change, as I feel that I'm potentially papering over what might be a valid/larger concern. I tried to limit the blast radius, though.

lib/oban/testing.ex
```diff
@@ -211,11 +213,11 @@ defmodule Oban.Testing do
   @spec perform_job(
           worker :: Worker.t(),
           args :: term(),
-          opts :: [Job.option() | {:repo, module()}]
+          opts :: [Job.option() | {:repo, module()} | {:testing, testing_mode()}]
```
shanesveller, line 216: Would you want this documented?

Usage at the call site feels a little bit clunky and I wasn't sure if I should provide a naming alias for `:testing` for clarity purposes:

```elixir
      assert :ok = perform_job(SomeWorker, some_args, testing: :manual)
```

lib/oban/testing.ex
```diff
@@ -236,6 +246,8 @@ defmodule Oban.Testing do
       |> Executor.call()
       |> Map.fetch!(:result)
 
+    Process.delete(:oban_engine)
```
shanesveller, line 249: Also deleted in `with_testing_mode` but I didn't recognize a better way to make sure it was reliably cleaned up.

test/oban/testing_test.exs
```diff
@@ -192,6 +192,11 @@ defmodule Oban.TestingTest do
     after
       :telemetry.detach("perform-job-handler")
     end
+
+    test "changing testing mode via opts" do
```
shanesveller, line 196: An expansion test case that matches my original failure mode would be to `perform_job` on a Job that inserts other Jobs while in `inline` mode, but I wanted to get your take on whether this was too corner-case to persist in the test suite. My original misunderstanding might point towards a doc clarification instead?

test/oban/testing_test.exs
```diff
@@ -320,6 +325,20 @@ defmodule Oban.TestingTest do
         refute_received {:ok, 1}
       end)
     end
+
+    test "with perform_job" do
```
shanesveller, line 329: These are mostly guarding against regressions without serving a lot of inherent value to a reader, IMO. Happy to take suggestions.

lib/oban/queue/basic_engine.ex
sorentwo: The `:inline` option maps directly to the `InlineEngine`, so `BasicEngine` shouldn't ever be called with an inline config.

lib/oban/testing.ex
sorentwo: Hmm. There isn't a real job to operate on in the db, so the `perform_job` helper should always be inline and skip any acking/db interaction.

I thought that this would look more like:

```elixir
  engine = Process.get(:oban_engine)

  try do
    Process.put(:oban_engine, :inline)

    # other code
  after
    Process.put(:oban_engine, engine)
  end
```

sorentwo (changes requested):
Thanks for the thorough self-review; much appreciated! The code here looks excellent, but I think it's a _little_ too involved for the solution I had in mind. Let me know what you think about pairing it back and simply wrapping the perform_inline in a `try/after` as suggested.

shanesveller:
🤔  I do see that a much simpler diff like you supplied will pass my supplied tests, which addresses the surface bug. If that feels like the right change I'm happy to revise the PR in that direction.

I have a project where job A inserts zero/one/many of job B based on rows in a given Ecto table, intended for use as part of a cron shedule to instigate other fanout work. That's its sole purpose, so the intended test case for this job is that some number of job Bs were inserted based on other database content. The side effect is the only thing I really want to assert on. I could move that into a pure function and test it outside of the context of Oban, which is not without merit.

My ultimate goal with this PR or at least with raising feedback to you is that I want an opportunity to use a `perform_job` with job A and deliberately **not** fire job B's internal behavior at the time of the `perform_job`, similar to the `:manual` testing mode or <= 2.11 behavior. Ideally this would be possible independent of what my project-wide or test-local testing mode are set to - though it definitely doesn't need to be automatic or particularly first-class, just possible. As one portion of the motive, I wouldn't want to have other job A's tests need to be setting up the ambient requirements for those inserted job Bs, as they may depend on other database rows, setting up Tesla HTTP mocks, etc.

I think my personal taste would be a project-wide default of `:inline`, but with the more minimal version of these changes even using `with_testing_mode` for `:manual` won't prevent job A from recursing into those job Bs.

WDYT?

shanesveller:
The minimal version I tried, relative to commit 4b76efa looked like:

```diff
diff --git a/lib/oban/testing.ex b/lib/oban/testing.ex
index 4b76efa..a4defe0 100644
--- a/lib/oban/testing.ex
+++ b/lib/oban/testing.ex
@@ -228,13 +228,19 @@ defmodule Oban.Testing do
 
     assert_valid_changeset(changeset)
 
+    original_engine = Process.put(:oban_engine, Oban.Queue.InlineEngine)
+
     result =
-      conf_opts
-      |> Keyword.put(:testing, :inline)
-      |> Config.new()
-      |> Executor.new(create_job(changeset), safe: false)
-      |> Executor.call()
-      |> Map.fetch!(:result)
+      try do
+        conf_opts
+        |> Keyword.put(:testing, :inline)
+        |> Config.new()
+        |> Executor.new(create_job(changeset), safe: false)
+        |> Executor.call()
+        |> Map.fetch!(:result)
+      after
+        Process.put(:oban_engine, original_engine)
+      end
 
     assert_valid_result(result)
 
diff --git a/test/oban/testing_test.exs b/test/oban/testing_test.exs
index d3fc712..04fbebe 100644
--- a/test/oban/testing_test.exs
+++ b/test/oban/testing_test.exs
@@ -320,6 +320,16 @@ defmodule Oban.TestingTest do
         refute_received {:ok, 1}
       end)
     end
+
+    test "with perform_job" do
+      Testing.with_testing_mode(:manual, fn ->
+        assert :ok = perform_job(Worker, %{ref: 1, action: "OK"})
+      end)
+
+      Testing.with_testing_mode(:inline, fn ->
+        assert :ok = perform_job(Worker, %{ref: 1, action: "OK"})
+      end)
+    end
   end
 
   defp assert_perform_error(worker, message) when is_binary(message) do
```

sorentwo:
I misunderstood the depth of your goal, possibly due to the failing test cases. Considering the mixed cartesian product of `inline` and `manual` I think a hybrid of your initial approach and the `try/after` safety I suggested is best. To summarize:

* Keep the new `testing` override for `perform_job` and make sure it's documented
* Wrap the engine change in `try/after` to ensure it's restored with an error (an edge case, but worth handling)
* Don't sweat the extra test case(s)

PR #700, opened 2022-04-28
Use query_opts in Repo.transaction options

simonprev:
I don’t know the impact of adding `prefix` in the options but the `telemetry_options` should also be present in the transaction call I think.

Without `telemetry_options`, package like https://github.com/fuelen/ecto_dev_logger will log `begin` and `commit` since the `oban_conf` is not present in the telemetry event.


sorentwo:
This is great! There's no concern about `prefix`, because that option is ignored for `Repo.transaction`. 💛

PR #706, opened 2022-05-08
Allow passing `DateTime` for the `:with_scheduled` option when draining

stefanchrobot:
Follow-up to the discussion on Elixir Slack.

In integration tests it's useful to check the side effects of running background jobs, for example verify that an email was sent. In my particular case, I'm checking for outgoing emails that are scheduled for future. The new `:with_scheduled_up_to` option allows for better control, basically giving the option to move the time forward.

In terms of the implementation, I went with the suggestion from the original discussion on Slack, but maybe there's a better way to handle `:with_scheduled` vs `:with_scheduled_up_to` (Raise when both are set? Use `with_scheduled: <timestamp>` instead?)

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -421,7 +421,7 @@ defmodule Oban do
 
   Drain a queue including jobs scheduled up to a minute:
 
-      Oban.drain_queue(queue: :default, with_scheduled: DateTime.utc_now() |> DateTime.add(60, :second))
+      Oban.drain_queue(queue: :default, with_scheduled: DateTime.add(DateTime.utc_now(), 60, :second))
 
   Drain a queue and assert an error is raised:
 
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -56,7 +56,7 @@ defmodule Oban do
           | {:with_limit, pos_integer()}
           | {:with_recursion, boolean()}
           | {:with_safety, boolean()}
-          | {:with_scheduled, boolean()}
+          | {:with_scheduled, boolean() | DateTime.t()}
 
   @type drain_result :: %{
           discard: non_neg_integer(),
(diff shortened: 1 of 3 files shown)


sorentwo (changes requested):
Thanks, @stefanchrobot !

The code, spec, and doc changes all look spot on. However, looking at the code now, I'm leaning toward having `with_scheduled` accept a timestamp rather than a separate option. That should eliminate confusion about what happens by passing `with_scheduled: false` and `with_scheduled_up_to: time`.

sorentwo:
@stefanchrobot Bump!

lib/oban.ex
```diff
@@ -416,6 +419,10 @@ defmodule Oban do
       Oban.drain_queue(queue: :default, with_scheduled: true)
       %{failure: 0, snoozed: 0, success: 1}
 
+  Drain a queue including jobs scheduled up to a minute:
+
+      Oban.drain_queue(queue: :default, with_scheduled: DateTime.utc_now() |> DateTime.add(60, :second))
```
sorentwo, line 424:
```suggestion
      Oban.drain_queue(queue: :default, with_scheduled: DateTime.add(DateTime.utc_now(), 60, :second))
```

lib/oban/queue/drainer.ex
```diff
@@ -16,22 +16,29 @@ defmodule Oban.Queue.Drainer do
       |> Map.put_new(:with_recursion, false)
       |> Map.put_new(:with_safety, true)
       |> Map.put_new(:with_scheduled, false)
+      |> Map.put_new(:with_scheduled_up_to, nil)
```
sorentwo, line 19:
```suggestion
```

test/integration/draining_test.exs
```diff
@@ -22,16 +22,44 @@ defmodule Oban.Integration.DrainingTest do
...
+
+    test "DateTime range is inclusive" do
+      name = start_supervised_oban!(queues: false)
+
+      scheduled_at = seconds_from_now(3600)
+
+      insert!(%{ref: 1, action: "OK"}, scheduled_at: scheduled_at)
+
+      assert %{success: 1, failure: 0} =
+               Oban.drain_queue(name, queue: :alpha, with_scheduled: scheduled_at)
```
sorentwo, line 61:
```suggestion
      assert %{success: 1} = Oban.drain_queue(name, queue: :alpha, with_scheduled: scheduled_at)
```

sorentwo (approved):
That's much better. Thanks!

PR #708, opened 2022-05-12
Change docs to reference newest notifier

agundy:
Based on the docs [here](https://hexdocs.pm/oban/2.11.3/changelog.html#alternative-pg-process-groups-notifier) I believe this should be updated


sorentwo:
Absolutely right, thanks!

PR #710, opened 2022-05-17
chore: add note about Oban.Peer when upgrading to v2.11

milmazz:


sorentwo:
Edited and merged locally in d73ea6e (you maintained attribution for the commit).

PR #715, opened 2022-05-26
Include minimal worker example

sorenone:
Rather than only showing a complex example with all options, this
minimal example highlights creating a worker with defaults.


sorentwo:
Thanks! 💛❤️

PR #716, opened 2022-05-27
Validate options after all normalizations

michallepicki:
In a project I had a config (for all envs, in config.exs) where a queue had only :global_limit specified (we're using Pro SmartEngine). In test env, when trying to use the new `:inline` mode, even though the changelog says setting the `queues` config to an empty list for tests is no longer needed, I had to do it because this queue didn't have a valid `:limit` option and wasn't passing InlineEngine's validations. This patch should make it actually work.


sorentwo (approved):

sorentwo:
Great catch, thanks! 💛

PR #723, opened 2022-06-10
chore: add release workflow

yordis:
### Context

Publish the package when a new release is published

### Requirements

- It requires a secret `HEX_API_KEY` with the Hex API Key in order to be able to publish it

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,32 @@
+name: Hexpm Release
+
+on:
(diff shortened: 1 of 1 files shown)


sorentwo:
Thanks for the PR! Locally I use an alias that is essentially `git push; git push --tags; mix hex.publish`, and that works well enough for now. This would come in handy if there's ever an additional maintainer that publishes packages. Closing for now.

PR #725, opened 2022-06-16
Fix links to guides in README

dsdshcym:


sorentwo:
Thanks! 💛

dsdshcym:
Ooops, I think this change might break the links in Hexdocs though 😱 
Maybe the fix is not this simple, we may need to make the way oban generates these Hexdocs pages consistent with their structure in the repo?

sorentwo:
Ah. Yeah, the end goal is to minimize the readme and expand all sections into proper guides. When in doubt, favor the hexdocs version of the README version.

dsdshcym:
yeah, that makes sense
I'm not sure if we can make github version work then
I think we'd better revert this PR to fix hexdocs version

sorentwo:
Reverted 👍

PR #727, opened 2022-06-17
Add Core.Repo.aggregate

sax:
We have a use case where we want to constrain queue size and feed our queues incrementally, as we have a large backlog of pending jobs, very slow external services, and small available concurrency on those services. We've estimated that it will take over a year to process all jobs. The prioritization of pending jobs will change on the fly.

We initially used `select: count(job.id)` with some state filters to find unprocessed jobs, but that makes our dialyzer unhappy... the query would return `[integer]`, but the spec on any of the available `Oban.Repo` functions is a map.

This function works as intended and also allows us to run dialyzer.


sorentwo:
Nice addition! Thanks @sax 💛

sorentwo (commented):
This is failing a dialyzer run. Can you fix up the spec?

sax:
Will do! I was so excited that it passed dialyzer in our application, I forgot to run it on oban.

sorentwo:
There we go. Thanks again!

sax:
Yay! Thank you!

PR #730, opened 2022-06-29
Soft replace discard with cancel return value

### LINKED ISSUE #703: Cancel jobs on Oban.Worker.perform/1 result

If you want to interrupt a job nowadays from the worker's perspective, you can only use `:discard` and `{:discard, reason()}` from  [Oban.Worker.result()](https://hexdocs.pm/oban/2.12.0/Oban.Worker.html#t:result/0). But the thing is that the `discarded` won't have the same semantic meaning as `canceled.` 

That became a problem in my use case when we started monitoring our queues so we could come up with dashboards and alerts based on their states. We understand that the `discarded` is highly associated with the exhaustion of retry attempts. That is considered a problem, and we want to alert on it. But we do not want to alert on `cancelled` since that is an intended action.

We found our way around it by relying on `{:discard, reason}` and started interpreting jobs with `{:discard, :cancelled}` error as a cancelled status in our metrics. But the solution is hacky and has some additional costs. We've considered other approaches, but they would all add an extra layer of complexity that we wanted to avoid. 

I want to be able to set a `cancelled` status from the worker by responding: `:cancelled | {:cancelled, reason()}` from perform/1 or process/1 (on Pro workflows).

The only thing I've discussed with my colleagues was updating the job's metadata to include this information and make it easier to figure out when the discard is intentional. But I figured that adding `cancelled` would be better since it's already a defined status in Oban.

### Additional Context

* This was previously discussed on Oban's slack channel: https://elixir-lang.slack.com/archives/CRDGH1XCM/p1651591825203179

sorentwo:
Discard was originally intended to mean "a job exhausted all retries." Later, it was added as a return type and it came to mean either "stop retrying" or "exhausted retries" ambiguously, with no clear way to differentiate. Even later, we introduced `cancel` with a `cancelled` state as way to stop jobs at runtime. At a semantic level, it's better to keep discarding and cancelling distinct.

This introduces a `{:cancel, reason}` return type and soft deprecates the use of `discard` from `perform/1` entirely. The meaning of each action/state is now:

* `cancel`—this job was purposefully stopped from retrying, either from a return value or the cancel command
* `discard`—this job has exhausted all retries

Closes #703

/cc @jc00ke @igorffs @patmaddox

### DIFF
--- a/lib/oban/engines/basic.ex
+++ b/lib/oban/engines/basic.ex
@@ -182,10 +182,24 @@ defmodule Oban.Engines.Basic do
   end
 
   @impl Engine
-  def cancel_job(%Config{} = conf, %Job{id: id}) do
-    query = where(Job, [j], j.id == ^id)
+  def cancel_job(%Config{} = conf, %Job{} = job) do
+    updates = [set: [state: "cancelled", cancelled_at: utc_now()]]
 
-    cancel_all_jobs(conf, query)
+    updates =
+      if is_map(job.unsaved_error) do
+        error = %{attempt: job.attempt, at: utc_now(), error: format_blamed(job.unsaved_error)}
+
+        Keyword.put(updates, :push, errors: error)
+      else
+        updates
+      end
+
+    query =
+      Job
+      |> where(id: ^job.id)
+      |> where([j], j.state not in ["cancelled", "completed", "discarded"])
+
+    Repo.update_all(conf, query, updates)
 
     :ok
   end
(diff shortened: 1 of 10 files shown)


lib/oban/engines/basic.ex
```diff
@@ -182,10 +182,24 @@ defmodule Oban.Engines.Basic do
...
-  def cancel_job(%Config{} = conf, %Job{id: id}) do
-    query = where(Job, [j], j.id == ^id)
+  def cancel_job(%Config{} = conf, %Job{} = job) do
+    updates = [set: [state: "cancelled", cancelled_at: utc_now()]]
+
+    updates =
+      if is_map(job.unsaved_error) do
+        error = %{attempt: job.attempt, at: utc_now(), error: format_blamed(job.unsaved_error)}
+
+        Keyword.put(updates, :push, errors: error)
```
dsdshcym, line 192: When I changed from returning `{:discard, reason}` to `{:cancel, reason}`
The `reason` was not recorded in errors any more

Because I'm using oban_pro SmartEngine, so I think oban_pro 0.12 might be missing this logic

PR #736, opened 2022-07-11
Fix broken link to installation guide

joshnuss:
Small fix. Link in readme didn't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ coordination.
 
 ### Installation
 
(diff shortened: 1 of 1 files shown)


sorentwo:
Changing that link breaks hexdocs, which is what the document is geared toward. This is definitely an issue though, because people keep opening PRs to fix it!

I’ll leave this open for others to see until the docs are sorted.

sorentwo:
Fixed with a link directly to hexdocs 👍

PR #743, opened 2022-07-21
feat: allow to specify query timeouts on Reindexer plugin

milmazz:
In some cases, applying `REINDEX INDEX CONCURRENTLY` on the indexes `oban_jobs_args_index`, and `oban_jobs_meta_index` takes more than the default value (15 seconds). This new option will allow clients to specify other values than the default.

### DIFF
--- a/lib/oban/plugins/reindexer.ex
+++ b/lib/oban/plugins/reindexer.ex
@@ -33,6 +33,8 @@ defmodule Oban.Plugins.Reindexer do
 
     * `:schedule` — a cron expression that controls when to reindex. Defaults to `"@midnight"`.
 
+    * `:timeout` - time in milliseconds to wait for each query call to finish.
+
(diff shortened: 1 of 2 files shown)


lib/oban/plugins/reindexer.ex
```diff
@@ -60,9 +60,9 @@ defmodule Oban.Plugins.Reindexer do
       :name,
       :schedule,
       :timer,
-      :timeout,
       indexes: ~w(oban_jobs_args_index oban_jobs_meta_index),
-      timezone: "Etc/UTC"
+      timezone: "Etc/UTC",
+      timeout: 15_000
```
sorentwo, line 65: Nitpick, this is purely to match other plugins:

```suggestion
      timeout: :timer.seconds(15)
```

sorentwo (approved):
Great work. I'll merge in that tiny suggestion and this is good to go 👍

sorentwo:
💛

Issue #635, opened 2022-02-11
Replacing values for unique jobs is unsafe when the job is executing?


dsdshcym:
> ### Precheck
> 
> - Do a quick search and make sure the bug has not yet been reported
> - For support, favor using the Elixir Forum, Slack, IRC, etc.
> - Be friendly and polite!
> 
> ### Environment
> 
> - Oban Version
> - PostgreSQL Version
> - Elixir & Erlang/OTP Versions (`elixir --version`)

Hi @sorentwo,
I'm planning to use the replacing values feature for unique jobs,
but I think Oban may not handle this edge case well.

### Current Behavior

1. When BasicEngine fetches a job to execute, [it would hold a row-level lock](https://github.com/sorentwo/oban/blob/785a2a787249081d26903a14dc53b888094ad0f1/lib/oban/queue/basic_engine.ex#L109) (SELECT FOR UPDATE)
2. Then BasicEngine [changes the job state to executing, and releases the row-level lock](https://github.com/sorentwo/oban/blob/785a2a787249081d26903a14dc53b888094ad0f1/lib/oban/queue/basic_engine.ex#L116-L129)
3. During the execution, [no lock is held by Oban](https://github.com/sorentwo/oban/blob/785a2a787249081d26903a14dc53b888094ad0f1/lib/oban/queue/producer.ex#L234-L242)
4. If we insert the same job with different `args` during the job execution: `UniqueWorker.new(new_args, replace: [:args])`, [the args would be updated](https://github.com/sorentwo/oban/blob/785a2a787249081d26903a14dc53b888094ad0f1/lib/oban/queue/basic_engine.ex#L293), but the executor is still running with the old args.
5. Then if the executor completes the job successfully, [the job state would change to completed](https://github.com/sorentwo/oban/blob/785a2a787249081d26903a14dc53b888094ad0f1/lib/oban/queue/basic_engine.ex#L134-L138), while the args is now a different value than the one used in executor.

### Expected Behavior

I'd expect Oban to not produce `completed jobs with args that are not used during the execution`
It seems our only option is to disallow `replace during executing`?
Maybe we can change the behaviour to either:
1. Do not replace args or other fields until job leaves executing state (i.e. do not allow replace when executing)
2. Allow specifying granular unique behaviours on conflict (e.g. replace when non-executing, do not replace when executing)

sorentwo:
[...]

PR #744, opened 2022-07-22
bugfix: restrict replace option when inserting unique job

milmazz:
This PR restricts the `replace` option when Oban detects a unique conflict. Passing the `replace` option is only allowed on `available`, `retryable`, or `scheduled` job states.

Fixes: #635


sorentwo (commented):
There isn't any way for users to indicate that they only want to replace for specific states but still consider a job unique in other states. I'm afraid this is a backward incompatible change. This whole bugfix could break the behaviour of existing applications ☹️ 

I've written a fix for this myself in the past before I realized the implications for existing applications. That's why the bug isn't fixed. I believe the only proper solution is to expand the `replace` API and make it either accept a list of fields (as it does here), or a nested list of states -> fields:

```elixir
# The current version
replace: [:args]

# The safer but optional version
replace: [available: [:args], retryable: [:args], scheduled: [:args]]
```

milmazz:
> I'm afraid this is a backward incompatible change. This whole bugfix could break the behaviour of existing applications ☹️

💯 It is backward incompatible. After reading #635, I incorrectly assumed that we were willing to take that risk, and I forgot to do that annotation in the original PR description. Anyway, the idea around the PR is to continue that conversation and see if we agree on the way to fix the _bug_, because I agree that allowing to change some of the job attributes once those reach the _final_ (`completed`, `discarded`, `cancelled`, etc.) states doesn't make too much sense and could be misleading if you do some kind of audit later, at least I haven't found a use case for this `replace` feature yet.

> I've written a fix for this myself in the past before I realized the implications for existing applications. That's why the bug isn't fixed.

It seems I make the same mistake 😆 

> I believe the only proper solution is to expand the replace API and make it either accept a list of fields (as it does here)

I like the _safer_ approach that you propose, it's undoubtedly more explicit, but it allows more control to the clients. What do you think if we add this new _safer_ approach and _soft-deprecate_ the current version? Or if you don't like the idea of _seft-deprecating_ the current practice, I _think_ we should at least document somewhere the possible inconvenience of using it.

sorentwo:
> I like the safer approach that you propose, it's undoubtedly more explicit, but it allows more control to the clients. What do you think if we add this new safer approach and soft-deprecate the current version? Or if you don't like the idea of seft-deprecating the current practice, I think we should at least document somewhere the possible inconvenience of using it.

That's a great approach and precisely what I'd suggest! Keep the old version in place (just like we do for `replace_args`), but remove all documentation about it and silently convert it into the new, safer approach.

The only part I'm unsure of is whether we should support modifying `executing`, `completed`, etc. I can imagine a scenario where you'd want to swap the meta of a `completed` job, but I think that's too niche for us to worry about.

milmazz:
> The only part I'm unsure of is whether we should support modifying `executing`, `completed`, etc.

Me too, but I think the new approach should solve this.

Let's go by parts. I think we both agree on the following:

```elixir
fields = ~w(args max_attempts meta priority queue scheduled_at tags worker)a
```

| `job.state`                           | `replace`                                                      | allowed? | Notes       | 
|---------------------------------|--------------------------------------------------------------|----------|-------------|
| "all_states"                    | `[]`                                                           | ⛔        |  Default           |
| `available`, `scheduled`, `retryable` | `[available: fields, scheduled: fields, retryable: fields]` | ✅        |  |

But we're not sure what to do with the following states because they are a little bit tricky:

```elixir
info_fields = ~w(meta tags)a 
internal_fields = ~w(args max_attempts priority queue scheduled_at worker)a
```

| `job.state`                           | `replace`                                                      | allowed? | Notes       | 
|---------------------------------|--------------------------------------------------------------|----------|-------------|
| `cancelled`, `completed`, `discarded` |  `[cancelled: info_fields, completed: info_fields, discarded: info_fields]`  |  ✅    |             |
| `cancelled`, `completed`, `discarded` |  `[cancelled: internal_fields, completed: internal_fields, discarded: internal_fields]`  |  ❓ | Reject operation or allow + `Logger.warn`  |

I don't see any issues with modifying the `info_fields` on a job in its final state. But, at least I can't imagine a use case where I would want to replace a member of the `internal_fields` once the job finishes. But, _if we assume that the user knows what they're doing in this proposed approach_, we could simply allow them to do whatever they want. An alternative approach is to enable them to replace the value, but we emit a _warning_ about it if they are updating a member of the `internal_fields`.

Now, let us consider _executing_ a particular case. Indeed, attempting to replace specific attributes like `args`, `worker`, `priority`, `queue` could give the wrong impression to the user, _assuming they believe that the executor will take those values on the fly_, when we know that's not the case.

| `job.state`                           | `replace`                                                      | allowed? | Notes       | 
|---------------------------------|--------------------------------------------------------------|----------|-------------|
| `executing` |  `[executing: info_fields]`  |   ✅        |             |
| `executing` |  `[executing: internal_fields]`  | ❓   | Reject operation or allow + `Logger.warn`   |

But again, for simplicity in the implementation, we could assume that the user knows what they're doing because they explicitly included the `executing` state in the `replace` option. So, it seems that it is ok to allow them to replace the value, but we should emit a _warning_ to clarify a possible misunderstanding.

TLDR:

* `available`, `scheduled`, `retryable` we allow the user to replace whatever they want.
* `cancelled`, `completed`, `discarded`, and `executing` we allow the user to replace whatever they want, but we emit a warning in the case they are trying to modify a member of the `internal_fields`.

WDYT?

sorentwo:
Your TLDR solution seems ideal, except that I think we should document a warning rather than emitting one in code. If somebody has chosen to do it purposefully, we shouldn't spam their logs every time the replace option is used.

milmazz:
Ok, no problem. I will try to work on this during this weekend o/

lib/oban/engines/basic.ex
```diff
@@ -324,20 +324,43 @@ defmodule Oban.Engines.Basic do
...
+    match?([{_, _} | _], replace_keys) and
+      replace_keys |> Keyword.keys() |> Enum.all?(&(&1 in all_states))
+  end
+
+  defp maybe_replace(_conf, _query_opts, job, _changeset, []), do: {:ok, job}
+
+  defp maybe_replace(conf, query_opts, job, changeset, replace_keys),
+    do: replace_keys(conf, query_opts, job, changeset, replace_keys)
+
+  defp replace_keys(conf, query_opts, job, changeset, replace_keys) do
```
sorentwo, line 356: Let's avoid the `maybe` variant and rely on a function head:

```suggestion
  defp replace_keys(_conf, _query_opts, job, _changeset, []), do: {:ok, job}
  
  defp replace_keys(conf, query_opts, job, changeset, replace_keys) do
```

test/integration/uniqueness_test.exs
```diff
@@ -158,6 +158,77 @@ defmodule Oban.Integration.UniquenessTest do
...
+               context.name,
+               UniqueWorker.new(%{id: 1},
+                 unique: [keys: [:id]],
+                 scheduled_at: in_two_minutes,
+                 tags: ["b"],
+                 priority: 2,
+                 max_attempts: 3,
+                 replace: [completed: [:scheduled_at, :tags, :priority]]
+               )
+             )
```
sorentwo, line 229: The important thing here is that _no_ fields change. No need to be so exhaustive, all you need to test is one attribute, like `priority`.

lib/oban/engines/basic.ex
```diff
@@ -324,20 +324,43 @@ defmodule Oban.Engines.Basic do
...
+        end
     end
   end
 
+  def valid_replace_option?(replace_keys) do
+    all_states = Oban.Job.states()
+
+    match?([{_, _} | _], replace_keys) and
+      replace_keys |> Keyword.keys() |> Enum.all?(&(&1 in all_states))
+  end
```
sorentwo, line 349: The `match?` got me at first. How about `keyword?/1`

```suggestion
  def state_replace_option?(replace) do
    all_states = Oban.Job.states()

    Keyword.keyword?(replace) and Enum.all?(replace, fn {state, _} -> state in all_states end)
  end
```

lib/oban/engines/basic.ex
```diff
@@ -324,20 +324,43 @@ defmodule Oban.Engines.Basic do
...
         {:ok, job}
 
-      replace_keys ->
-        Repo.update(
-          conf,
-          Ecto.Changeset.change(job, Map.take(changeset.changes, replace_keys)),
-          query_opts
-        )
+      replace ->
+        if valid_replace_option?(replace) do
```
sorentwo, line 333: The legacy options are valid too, just deprecated. How about `state_replace_option?`

sorentwo (commented):
Very elegant start! I've left a few suggestions for clarity 👍

milmazz:
@sorentwo I think this is ready for a second review; please let me know what you think.

README.md
```diff
@@ -483,9 +483,14 @@ or `schedule_in` values will work, but the replace option is always
...
 
 ```elixir
-UrgentWorker.new(args, schedule_in: 1, replace: [:scheduled_at])
+UrgentWorker.new(args, schedule_in: 1, replace: [scheduled: [:scheduled_at]])
 ```
 
+NOTE: If you use this feature to replace a field (e.g. `args`) in the `executing`
+state by doing something like: `UniqueWorker.new(new_args, replace: [executing:
+[:args]])` keep in mind that Oban will update the field or fields, but the Oban
+executor runs with the old values.
```
sorentwo, line 492:
```suggestion
NOTE: If you use this feature to replace a field (e.g. `args`) in the `executing` state by doing
something like: `UniqueWorker.new(new_args, replace: [executing: [:args]])` Oban will update the
`args`, but the job will continue executing with the original value.
```

sorentwo (approved):
Fantastic!

sorentwo:
Thanks @milmazz 💛

PR #754, opened 2022-08-03
Fix specs for insert and insert_all

Schultzer:
This PR fixes the spec issues we have experienced while trying to upgrade Oban from: `2.10` -> `2.13`.


sorentwo:
Dialyzer passed for that one, but it wasn't formatted. I've formatted it and merged it locally. Thanks for the fix!

PR #758, opened 2022-08-17
Fix query that drops invalid indexes

howleysv:
We noticed that the invalid indexes in our repo were not being cleaned up by the reindexer and after looking into it and running the query manually, we discovered the query always silently fails on some name/syntax issues. While debugging we also noticed that the telemetry almost always indicates a failure (by including the `error` field in the event meta), so I cleaned up the return value of `check_leadership_and_reindex()`


sorentwo:
@howleysv Catching silent failures is tricky; thank you for the PR. While investigating the issue, I noticed that the test passed erroneously after your changes. That led me to write a new test and take a different approach to solve the problem.

PR #779, opened 2022-09-28
lifeline: allow override global rescue_time per worker

milmazz:
This PR adds the following two options to the `Oban.Plugins.Lifeline` plugin:

* `timeout` - to be able to modify the query timeouts
* `worker_overrides` - this is mainly to allow OSS developers to change the global `rescue_after` for more specific scenarios based on the worker name.

### DIFF
--- a/lib/oban/plugins/lifeline.ex
+++ b/lib/oban/plugins/lifeline.ex
@@ -30,12 +30,26 @@ defmodule Oban.Plugins.Lifeline do
         plugins: [{Oban.Plugins.Lifeline, rescue_after: :timer.minutes(5)}],
         ...
 
+  If you want a more granular option, you can also override the default period
+  to rescue specific Oban workers by their names:
+
+      config :my_app, Oban,
+        plugins: [{Oban.Plugins.Lifeline, worker_overrides: [{MyApp.MyWorker, :timer.minutes(5)}]],
+        ...
+
   ## Options
 
   * `:interval` — the number of milliseconds between rescue attempts. The default is `60_000ms`.
 
   * `:rescue_after` — the maximum amount of time, in milliseconds, that a job may execute before
   being rescued. 60 minutes by default, and rescuing is performed once a minute.
+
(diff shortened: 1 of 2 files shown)


lib/oban/plugins/lifeline.ex
```diff
@@ -37,6 +37,13 @@ defmodule Oban.Plugins.Lifeline do
   * `:rescue_after` — the maximum amount of time, in milliseconds, that a job may execute before
   being rescued. 60 minutes by default, and rescuing is performed once a minute.
 
+  * :`:worker_overrides` - allows to override the maximum amount of time, in
```
milmazz, line 40: * [ ] TODO: Add an example using this option

lib/oban/plugins/lifeline.ex
```diff
@@ -143,20 +161,41 @@ defmodule Oban.Plugins.Lifeline do
...
+
+          worker_counts =
+            for {worker, rescue_after} <- state.worker_overrides do
+              time = DateTime.add(now, -rescue_after, :millisecond)
+              query = where(base, [j], j.attempted_at < ^time and j.worker == ^worker)
+              run_transitions(query, state)
+            end
+
+          time = DateTime.add(now, -state.rescue_after, :millisecond)
+          query = where(base, [j], j.attempted_at < ^time)
```
milmazz, line 178: * [ ] Probably worth excluding the worker names from this query

sorentwo:
The basic lifeline plugin has a limited set of options intentionally, and worker overrides don't belong here. An optional query timeout is welcome (though unexpected for such a straightforward update query).

milmazz:
> The basic lifeline plugin has a limited set of options intentionally, and worker overrides don't belong here.

Closing this in favor of a custom plugin for my use case, I will leave this here in case another developer finds it useful.

> An optional query timeout is welcome (though unexpected for such a straightforward update query).

Yeah, without additional filters such as the proposed `worker_overrides` it doesn't make sense to provide a `timeout` option for the current simple approach.

PR #782, opened 2022-10-09
Documentation updated to reflect latest Honeybadger API

alexpls:
The Honeybadger package deprecated `notify/3` with its 0.15 release and prints a deprecation notice whenever the function is called ([docs here](https://hexdocs.pm/honeybadger/0.18.1/Honeybadger.html#notify/3)).

This change updates the Honeybadger Telemetry integration example to use the currently recommended API.


sorentwo:
Thank you!

PR #783, opened 2022-10-10
Change backoff example to use Job struct

gosimitz:
The backoff example implies that it takes an `attempt` and `base_amount` but elsewhere in the docs the function takes a `Job` struct. This updates the example.


gosimitz:
I'm also happy to adjust the example if need be!

sorentwo:
The recipe came from a blog post that predated that change. Thanks for the fix!

PR #795, opened 2022-10-21
Handle cancelled jobs in inline engine

pedroperrone:
Currently when a job is cancelled with the inline mode we get the following error:
```elixir
     ** (FunctionClauseError) no function clause matching in Oban.Engines.Inline.complete_job/1

     The following arguments were given to Oban.Engines.Inline.complete_job/1:

         # 1
         %Oban.Queue.Executor{result: {:cancel, :reason}, state: :cancelled, ...} # Rest of executor omitted for readability reasons

     Attempted function clauses (showing 4 out of 4):

         defp complete_job(%{job: job, state: :failure})
         defp complete_job(%{job: job, state: state}) when state === :discard or state === :exhausted
         defp complete_job(%{job: job, state: :success})
         defp complete_job(%{job: job, result: {:snooze, snooze}, state: :snoozed})

     stacktrace:
       (oban 2.13.4) lib/oban/engines/inline.ex:135: Oban.Engines.Inline.complete_job/1
       (oban 2.13.4) lib/oban/engines/inline.ex:29: Oban.Engines.Inline.insert_job/3
       (oban 2.13.4) lib/oban/engine.ex:280: anonymous fn/3 in Oban.Engine.with_span/4
```

With the changes made in this PR, cancelled jobs are correctly handled, receiving the correct state and having `cancelled_at` filled accordingly.

### DIFF
--- a/lib/oban/engines/inline.ex
+++ b/lib/oban/engines/inline.ex
@@ -134,6 +134,12 @@ defmodule Oban.Engines.Inline do
     %Job{job | errors: [error], state: "retryable", scheduled_at: utc_now()}
   end
 
+  defp complete_job(%{job: job, state: :cancelled}) do
+    error = %{attempt: job.attempt, at: utc_now(), error: format_blamed(job.unsaved_error)}
+
+    %Job{job | errors: [error], state: "cancelled", cancelled_at: utc_now()}
+  end
+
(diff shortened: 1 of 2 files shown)


sorentwo (approved):
Perfect. I’m surprised nobody has encountered this before!

PR #807, opened 2022-11-14
Update the @callback of Oban.Peer.leader? to match spec

NickPapacostas:
This was giving me some mocking trouble via Hammox, the spec and implementation expect an Oban.Config or GenServer.name.

### DIFF
--- a/lib/oban/peer.ex
+++ b/lib/oban/peer.ex
@@ -65,7 +65,7 @@ defmodule Oban.Peer do
   @doc """
   Check whether the current peer instance leads the cluster.
   """
-  @callback leader?(pid()) :: boolean()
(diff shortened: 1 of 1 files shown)


sorentwo:
Nice catch! This is a good fix. In general, for testing I suggest using a fake Peer rather than mocking it out.

NickPapacostas:
Thanks! Yea that's what I've done instead, but I'm not sure what the advantage is over directly using the Oban.Peer behavior to generate the mock. 

I might be missing something but I figure as long as we're relying on Oban.Peer.leader? in our code, our tests should check against the behavior directly.

sorentwo:
I meant to merge this before. Thanks!

PR #816, opened 2022-12-14
Improve the documentation around job state changes

chaodhib:
Attempt to improve the documentation around job state changes by providing a bit more information on what happens to the job's `state` depending on the value returned from `c:perform/1`.

### DIFF
--- a/lib/oban/worker.ex
+++ b/lib/oban/worker.ex
@@ -55,12 +55,15 @@ defmodule Oban.Worker do
 
   The value returned from `c:perform/1` can control whether the job is a success or a failure:
 
-  * `:ok` or `{:ok, value}` — the job is successful; for success tuples the `value` is ignored
+  * `:ok` or `{:ok, value}` — the job is successful; for success tuples the `value` is ignored.
+    The job is marked as `completed`.
 
   * `{:cancel, reason}` — cancel executing the job and stop retrying it. An error is recorded
-    using the optional `reason`, though the job is still successful.
+    using the optional `reason`. The job is marked as `cancelled`.
 
-  * `{:error, error}` — the job failed, record the error and schedule a retry if possible
+  * `{:error, error}` — the job failed, record the error. If `max_attempt`
+    has not been reached already, a retry is scheduled and the job is marked as `retryable`.
+    Otherwise, no retry is scheduled and the job is marked as `discarded`.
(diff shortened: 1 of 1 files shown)


lib/oban/worker.ex
```diff
@@ -55,12 +55,15 @@ defmodule Oban.Worker do
 
   The value returned from `c:perform/1` can control whether the job is a success or a failure:
 
-  * `:ok` or `{:ok, value}` — the job is successful; for success tuples the `value` is ignored
+  * `:ok` or `{:ok, value}` — the job is successful; for success tuples the `value` is ignored.
+    The job is marked as `completed`.
```
sorentwo, line 59:
```suggestion
  * `:ok` or `{:ok, value}` — the job is successful and marked as `completed`.  The `value` from
    success tuples is ignored.
```

lib/oban/worker.ex
```diff
@@ -55,12 +55,15 @@ defmodule Oban.Worker do
...
 
   * `{:cancel, reason}` — cancel executing the job and stop retrying it. An error is recorded
-    using the optional `reason`, though the job is still successful.
+    using the optional `reason`. The job is marked as `cancelled`.
```
sorentwo, line 62: It isn't actually optional:

```suggestion
    using the provided `reason`. The job is marked as `cancelled`.
```

lib/oban/worker.ex
```diff
@@ -55,12 +55,15 @@ defmodule Oban.Worker do
...
 
-  * `{:error, error}` — the job failed, record the error and schedule a retry if possible
+  * `{:error, error}` — the job failed, record the error. If `max_attempt`
+    has not been reached already, a retry is scheduled and the job is marked as `retryable`.
+    Otherwise, no retry is scheduled and the job is marked as `discarded`.
```
sorentwo, line 66:
```suggestion
  * `{:error, error}` — the job failed, record the error. If `max_attempts` has not been reached
    already, the job is marked as `retryable` and scheduled to run again. Otherwise, the job is
    marked as `discarded` and won't be retried.
```

sorentwo (approved):

sorentwo:
Thanks 💛

chaodhib:
Thank you ❤️ 💜

PR #817, opened 2022-12-20
support arg value serialization impls

ericdude4:
Not 100% sure about this approach, but I need a way to tell Oban how to serialize argument values before encoding storing in JSONB w/ Jason.

Specifically, we are storing `Decimal.t()` values in our Oban Job args, and they are being deserialized as `%{"coef" => 15, "exp" => 1, "sign" => -1}`. Then, in the Oban worker `perform` callback, we have to cast this string-key map back into a Decimal struct. Our preference would be that Decimal args would just be stored as string representations in the database, so we can simply call `Decimal.new/1` on those strings in the `perform` callback (instead of the hacky map casting approach). 

Creating an implementation for `defimpl Jason.Encoder, for: Decimal` doesn't work for us because Jason is used in many places and layers throughout our production application. It doesn't make sense to create a new encode implementation for Decimal since this will apply to the whole application, instead of just Oban.

Context: https://elixir-lang.slack.com/archives/CRDGH1XCM/p1671488236832899

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -250,6 +250,8 @@ defmodule Oban.Job do
   @doc since: "0.1.0"
   @spec new(args(), [option()]) :: changeset()
   def new(args, opts \\ []) when is_map(args) and is_list(opts) do
+    args = serialize_arguments(args)
+
     params =
       opts
       |> Keyword.put(:args, args)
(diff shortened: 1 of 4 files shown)


sorentwo:
@ericdude4 Thanks for the PR!

The policy for determining what should be in OSS is whether something is _possible_ to implement in your own application. Adding a protocol for serialization handles one side of the encode/decode cycle, but it isn't powerful enough to cover deserialization back to the original type. IMO that logic belongs in your application or in a separate package (e.g. Oban Pro). For that reason, I'll pass on this addition.

However, documentation that describes how to handle custom types like `Decimal` is more than welcome!

ericdude4:
@sorentwo Makes sense! Thank you for the consideration!

PR #824, opened 2023-01-06
Switch from datetime fragment to binary operator

sorentwo:
This is a branch to diagnose why a simple `>=` comparison in the unique query doesn't seem to work without explicit `datetime` casting in a fragment.


sorentwo:
@warmwaffles  (@jc00ke) This comments out the working code for the failing `>=` comparison.

sorentwo:
A quick experiment shows that changing `inserted_at` from `utc_datetime_usec` to `utc_datetime` fixes the issue.

PR #827, opened 2023-01-14
Stabilize the test of retry_job/2 by serializing setup and assertions

sato11:
TLDR; As Producer dispatches or Stager stages, a parallel mutation of jobs happens which randomly makes the test of retry_job/2 fail. Using transaction erases unexpected mutation and therefore stabilizes the test.

Hi there 👋 Test seems [flaky](https://github.com/sorentwo/oban/actions/runs/3911825494/jobs/6685662188) and as I took a look I think I found a reasoning and a possible fix.

Test data is provisioned by independently inserting but as soon as they are committed a parallel mutation can possibly occur:

1. Stager can make "retryable" jobs "available"
1. Producer calls Engine.fetch_jobs/3 which renders "available" jobs "executing"

To prove it, you might want to peep into database like this and run test;

```diff
 @impl Engine
 def retry_job(%Config{} = conf, %Job{id: id}) do
+  print_job_states(conf)
   retry_all_jobs(conf, where(Job, [j], j.id == ^id))
+  Process.sleep(1)
 
   :ok
 end

+defp print_job_states(conf) do
+  Oban.Repo.all(conf, Job |> order_by(asc: :id))
+  |> Enum.map(fn job -> job.state end)
+  |> IO.inspect()
+end
```

and you will find each data is not properly isolated for each iteration:

```
["discarded", "completed", "available", "retryable", "executing"]
["available", "completed", "available", "available", "executing"] <- "retryable" is mutated in 2nd iteration
["executing", "executing", "executing", "executing", "executing"] <- every job starts execution already
["executing", "executing", "executing", "executing", "executing"]
["executing", "executing", "executing", "executing", "executing"]
```

Note that any constant seed does not reproduce the exact same result because of its parallel nature. The demonstration above is only one of possible situations.

This PR proposes that the test use transaction so that insertion is only visible to the current testing process until asserted. Because of the nature of existing assertions, I believe it's alright to only check each job's state functionally, and dispense with the actual orchestration of events.

I can see there are other instances where data isolation of a test is perhaps not guaranteed. On the other hand, since it seems this is where failure can happen most frequently, I'm submitting this hoping my assumption is correct. So will you take a look?

### DIFF
--- a/test/oban/engine_test.exs
+++ b/test/oban/engine_test.exs
@@ -441,23 +441,25 @@ for engine <- [Oban.Engines.Basic, Oban.Engines.Lite] do
       @describetag oban_opts: [queues: [alpha: 5], stage_interval: 10, testing: :disabled]
 
       test "retrying jobs from multiple states", %{name: name} do
-        job_a = insert!(name, %{ref: 1}, state: "discarded", max_attempts: 20, attempt: 20)
-        job_b = insert!(name, %{ref: 2}, state: "completed", max_attempts: 20)
-        job_c = insert!(name, %{ref: 3}, state: "available", max_attempts: 20)
-        job_d = insert!(name, %{ref: 4}, state: "retryable", max_attempts: 20)
-        job_e = insert!(name, %{ref: 5}, state: "executing", max_attempts: 1, attempt: 1)
+        @repo.transaction(fn ->
+          job_a = insert!(name, %{ref: 1}, state: "discarded", max_attempts: 20, attempt: 20)
+          job_b = insert!(name, %{ref: 2}, state: "completed", max_attempts: 20)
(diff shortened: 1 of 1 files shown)


sorentwo (commented):
Each test is already wrapped in a transaction as part of the ecto sandbox. The underlying problem is that any queues are started, and staging is enabled. The proper fix is to disable queues and staging for this test (and the `retry_all_jobs` test below).

```elixir
      @describetag oban_opts: [testing: :disabled]
```

sorentwo:
Separately, thank you for the thorough investigation and write up!

sorentwo:
Fixed with 8454616, thanks for the insight 💛

sato11:
You're welcome and thank you too 🙂

PR #828, opened 2023-01-16
Fix broken example in documentation

RodolfoSilva:
We cannot pin a struct field, instead we should pass an existing variable, otherwise we will get an error like this:

```
invalid argument for unary operator ^, expected an existing variable, got: ^job.id
```


sorentwo (approved):

sorentwo:
Thank you!

PR #839, opened 2023-02-05
Include jobs on fetch_jobs event metadata

joaogularte:
Hey there! 

In Oban Telemetry documentation is described that  `:stop` events for bulk operations have included the `:jobs` key on metadata events, however, this is not the case. Examining at [`Oban.Engine.fetch_jobs/3`](https://github.com/sorentwo/oban/blob/main/lib/oban/engine.ex#L207) callback implementation, we can see that isn't happing as [others bulk operations](https://github.com/sorentwo/oban/blob/main/lib/oban/engine.ex#L191), so this PR has the goal to fix it. 

To validate this change it was added a simple `assert_receive` at the  "inserting and executing jobs" engine integration test, despite I am not fully confident that is the best approach.

### DIFF
--- a/lib/oban/engine.ex
+++ b/lib/oban/engine.ex
@@ -206,7 +206,8 @@ defmodule Oban.Engine do
   @doc false
   def fetch_jobs(%Config{} = conf, %{} = meta, %{} = running) do
     with_span(:fetch_jobs, conf, fn engine ->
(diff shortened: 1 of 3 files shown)


sorentwo:
Excellent addition, thanks! I overlooked that event because each job emitted an `[:oban, :job, :start]`, which had the relevant information.

PR #843, opened 2023-02-11
Pass `pid` in instead of relying on `from`

josevalim:
See current discussion in https://github.com/elixir-ecto/postgrex/pull/643.

PS: I did not test this in any way in form but the idea should be clear. :)


sorentwo (approved):

sorentwo:
TIL gen_statem uses a proxy process to handle timeout calls. Thanks For the preemptive fix 💛

PR #846, opened 2023-02-14
Select only id in stage subquery

alexcastano:
This improves the performance of the stager query a lot. It changes from:

```sql
UPDATE
  "public"."oban_jobs" AS o0
SET
  "state" = $1
FROM (
  SELECT
    so0."id" AS "id",
    so0."state" AS "state",
    so0."queue" AS "queue",
    so0."worker" AS "worker",
    so0."args" AS "args",
    so0."meta" AS "meta",
    so0."tags" AS "tags",
    so0."errors" AS "errors",
    so0."attempt" AS "attempt",
    so0."attempted_by" AS "attempted_by",
    so0."max_attempts" AS "max_attempts",
    so0."priority" AS "priority",
    so0."attempted_at" AS "attempted_at",
    so0."cancelled_at" AS "cancelled_at",
    so0."completed_at" AS "completed_at",
    so0."discarded_at" AS "discarded_at",
    so0."inserted_at" AS "inserted_at",
    so0."scheduled_at" AS "scheduled_at"
  FROM
    "public"."oban_jobs" AS so0
  WHERE
    (so0."state" IN ($4, $5))
    AND (NOT (so0."queue" IS NULL))
    AND (so0."scheduled_at" <= $2)
  ORDER BY
    so0."id"
  LIMIT
    $3 FOR
  UPDATE
    SKIP LOCKED) AS s1
WHERE
  (o0."id" = s1."id") RETURNING s1."id",
  s1."queue",
  s1."state",
  s1."worker"
```

to

```sql
UPDATE
  "public"."oban_jobs" AS o0
SET
  "state" = $1
FROM (
  SELECT
    so0."id" AS "id"
  FROM
    "public"."oban_jobs" AS so0
  WHERE
    (so0."state" IN ($4, $5))
    AND (NOT (so0."queue" IS NULL))
    AND (so0."scheduled_at" <= $2)
  LIMIT
    $3 FOR
  UPDATE
    SKIP LOCKED) AS s1
WHERE
  (o0."id" = s1."id") RETURNING o0."id",
  o0."queue",
  o0."state",
  o0."worker"
```

The reason is that now it can use the "index only scan" for the compose index in the subquery:

`"oban_jobs_state_queue_priority_scheduled_at_id_index" btree (state, queue, priority, scheduled_at, id)`
 
image

Instead of using "index scan" for the ID index.

The main issue was the "unneeded" order by id.

Our system's average execution time changed from more than 2 seconds to less than 100ms.

### DIFF
--- a/lib/oban/stager.ex
+++ b/lib/oban/stager.ex
@@ -123,17 +123,17 @@ defmodule Oban.Stager do
   defp stage_scheduled(state, leader?: true) do
     subquery =
       Job
+      |> select([j], %{id: j.id})
       |> where([j], j.state in ["scheduled", "retryable"])
       |> where([j], not is_nil(j.queue))
       |> where([j], j.scheduled_at <= ^DateTime.utc_now())
-      |> order_by(asc: :id)
       |> limit(^state.limit)
       |> lock("FOR UPDATE SKIP LOCKED")
 
     query =
       Job
       |> join(:inner, [j], x in subquery(subquery), on: j.id == x.id)
-      |> select([_j, x], map(x, [:id, :queue, :state, :worker]))
+      |> select([j, _x], map(j, [:id, :queue, :state, :worker]))
 
     {staged_count, staged} = Repo.update_all(state.conf, query, set: [state: "available"])
 
--- a/lib/oban/stager.ex
+++ b/lib/oban/stager.ex
@@ -126,6 +126,7 @@ defmodule Oban.Stager do
       |> select([j], %{id: j.id})
       |> where([j], j.state in ["scheduled", "retryable"])
       |> where([j], not is_nil(j.queue))
+      |> where([j], j.priority >= 0 and j.priority <= 3)
       |> where([j], j.scheduled_at <= ^DateTime.utc_now())
       |> limit(^state.limit)
       |> lock("FOR UPDATE SKIP LOCKED")
--- a/lib/oban/stager.ex
+++ b/lib/oban/stager.ex
@@ -123,10 +123,10 @@ defmodule Oban.Stager do
   defp stage_scheduled(state, leader?: true) do
     subquery =
       Job
-      |> select([j], %{id: j.id})
+      |> select([:id, :state])
       |> where([j], j.state in ["scheduled", "retryable"])
       |> where([j], not is_nil(j.queue))
-      |> where([j], j.priority >= 0 and j.priority <= 3)
+      |> where([j], j.priority in [0, 1, 2, 3])
       |> where([j], j.scheduled_at <= ^DateTime.utc_now())
       |> limit(^state.limit)
       |> lock("FOR UPDATE SKIP LOCKED")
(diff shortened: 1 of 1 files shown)


sorentwo:
Hi @alexcastano, thanks for investigating stager performance and working to refine the query. I've attempted to recreate your `index_only_scan` performance locally with 200k scheduled jobs, but there hasn't been a meaningful difference. 

Perhaps the planner makes a more optimized query with a larger volume of jobs?

1. How many scheduled jobs do you have when staging takes 2s?
2. Which version of Postgres are you using?

For reference, here's the [public explain with optimizations](https://explain.depesz.com/s/bH9t) and relevant queries.

sorentwo:
I managed to get an index-only scan by including a `priority` condition! https://explain.depesz.com/s/Sy69F

alexcastano:
Hi! I don't understand why PG told me it was using `index only scan`, maybe because I "executed" only the subquery without the outer query. Anyway, I think there is a lot of room for improvement.

The number of "scheduled" jobs is not important for the old query because it was using the id index. It is all the jobs, scheduled or not, you have in the database. We currently have more than 1M rows in the oban_jobs.

Let me apply your suggestion and we'll see the results :)

lib/oban/stager.ex
```diff
@@ -123,17 +123,17 @@ defmodule Oban.Stager do
...
       |> where([j], j.scheduled_at <= ^DateTime.utc_now())
-      |> order_by(asc: :id)
       |> limit(^state.limit)
       |> lock("FOR UPDATE SKIP LOCKED")
 
     query =
       Job
       |> join(:inner, [j], x in subquery(subquery), on: j.id == x.id)
-      |> select([_j, x], map(x, [:id, :queue, :state, :worker]))
+      |> select([j, _x], map(j, [:id, :queue, :state, :worker]))
```
sorentwo, line 137: Flipping this returns the wrong `state` information. The purpose of the join/select is to get the original `scheduled` or `retryable` state exposed for telemetry:

```suggestion
      |> select([j, x], %{id: j.id, queue: j.queue, state: x.state, worker: j.worker})
```

lib/oban/stager.ex
```diff
@@ -123,17 +123,18 @@ defmodule Oban.Stager do
   defp stage_scheduled(state, leader?: true) do
     subquery =
       Job
+      |> select([j], %{id: j.id})
       |> where([j], j.state in ["scheduled", "retryable"])
       |> where([j], not is_nil(j.queue))
+      |> where([j], j.priority >= 0 and j.priority <= 3)
```
sorentwo, line 129: The `>=` and `<=` comparisons won't make use of the index like an `IN` clause:

```suggestion
      |> where([j], j.priority in [0, 1, 2, 3])
```

sorentwo (changes requested):
I've applied all of my suggestions locally and some test improvements that I'll try to push to this branch.

sorentwo:
Modified, squashed, and merged locally as 4e2fde8. Thanks again!

alexcastano:
Hi again!

We deployed the new query in production. It works much faster! However, it doesn't use `index only scan`, but `index scan`. This is the capture from Google Cloud Insights:

image

In addition, when the database is under heavy demand, it sometimes switches back to the slow sequence scan:

image

I think this behaviour may be expected for postgres, but I wanted to share with you because maybe you can make the query `index only scan`, or maybe ,change the index itself to remove the priority field.

Thanks for your work.

sorentwo:
That index is used for multiple queries, one of which fetches jobs and requires the priority field. Additionally, changing indexes is an expensive operation we can't do in a transactional migration without blocking.

PR #855, opened 2023-02-22
Change link caption for `Config Providers`

smaximov:
There's no mention of "Module Config Providers" in `mix release` documentation (`module` in `#module-config-providers` means the anchor refers to the section which is a part of the module docstring).


sorentwo:
Thanks!

PR #857, opened 2023-02-28
Fix documentation links to Oban Pro plugins

ucwaldo:
To make the links less reliant on the module names in Oban Pro, they could also point to `https://getoban.pro/docs/pro/0.13.1/overview.html#plugins`, but they are probably not bound to change any time soon :)


sorentwo:
Fantastic! Thank you for the fixes 💛

PR #861, opened 2023-03-03
Allow `{:snooze, 0}` job return

mtunski:
https://elixir-lang.slack.com/archives/CRDGH1XCM/p1677769842919649


mtunski:
Not sure if this requires some further testing, and if yes - where's the best place to do so.

sorentwo (approved):
This looks perfect. Scheduling relies on `DateTime.add`, which accepts `0, :seconds` perfectly fine. There's no need for additional tests!

sorentwo:
Thanks!

PR #873, opened 2023-03-22
Avoid unnecessary query for "happy path" in `assert_enqueued/2`

smaximov:
`available_jobs/2` is called unconditionally in `assert_enqueued/2` in order to format an error message that is used only if the assertion fails. This results in an extra SQL query when it doesn't, which can be avoided by using `flunk/1` to handle the "unhappy path".


sorentwo (approved):
Coincidentally, I noticed this yesterday and here you have a fix!

PR #882, opened 2023-04-08
Add migrating from postgres notifier to pg notifier

yordis:


lib/oban/notifiers/pg.ex
```diff
@@ -29,6 +29,24 @@ defmodule Oban.Notifiers.PG do
   * For compatibility, message payloads are always serialized to JSON before
     broadcast and deserialized before relay to local processes.
 
+  ## Migrating from `Oban.Notifiers.Postgres`
+
+  When migrating from `Oban.Notifiers.Postgres` you may want to remove
+  `oban_notify` trigger. To do so, you could use the following migration:
```
sorentwo, line 35:
```suggestion
  After switching from `Oban.Notifiers.Postgres`, you may remove the unused `oban_notify` trigger.
  Use the following migration to drop the trigger while retaining the `oban_jobs_notify` function:
```

sorentwo:
Thanks @yordis!

Issue #888, opened 2023-04-17
Helper functions to get available states by "categories"


NetonD:
### Is your feature request related to a problem? Please describe.

Outside from oban envieroment I think that is be nice if we can get "categorized" state types with functions instead we need to do this in application that implements oban.

### Describe the Solution You'd Like

Some simple functions that can return subset of availables states (`Oban.Job.states/0`)

### Describe Alternatives You've Considered

The following subsets can be helpful (the names can be better, this is just a first suggestions): 
- `up_states()`: "scheduled", "available", "executing", "retryable"
- `final_states()`: "completed", "discarded", "cancelled"
- `fail_states()`: "discarded", "cancelled"
- `wating_states()`: "scheduled", "available", "retryable"

This functions can lives on `Oban.Job`.

sorentwo:
Thanks for the suggestion (and the PR). Can you provide a concrete use case for categories of job states?

NetonD:
Sure, I can, the following example is a piece of code that I'm using to list the reports requests that can fail for a bunch of reasons and I need show if the report(represented by a Oban.Job) is running, crashes or if it running/waiting, so I query the jobs and show then in a phoenix template that has this function:

```elixir
  def status(status) do
    cond do
      status in ~w(available retryable scheduled) -> "Waiting"
      status == "completed" -> "Finished"
      status == "executing" -> "Running"
      true -> "Failed"
    end
  end
  ```
  
  But is weird my interface layer knows by hard code the states of Oban, so this can turn into:
  
  ```elixir
  def status(status) do
    case status do
      status when status in Oban.Job.states(:waiting) -> "Waiting"
      status when status in Oban.Job.states(:up) -> "Running"
      :completed -> "Finished"
      _ -> "Failed"
    end
  end
  ```

PR #889, opened 2023-04-17
Fetch states based on "types"

NetonD:
related to issue #888


sorentwo:
Closed because of reasons explained in #888

PR #891, opened 2023-04-20
Add `:interval` to `t:Oban.Plugins.Pruner.option/0`

smaximov:
The `:interval` option is documented but is missing from `t:Oban.Plugins.Pruner.option/0` type spec.


sorentwo (approved):

sorentwo:
Thanks!

PR #894, opened 2023-04-28
Remove reference to Stager

nathanl:
It was removed in 2.14 - https://hexdocs.pm/oban/v2-14.html#remove-repeater-and-stager-plugins


sorentwo (approved):
Stager wasn't removed, it just isn't a plug-in. Close enough!

PR #896, opened 2023-05-01
Fix `validate/1` in the "Writing Plugins" guide

smaximov:
I chose to use `is_integer/1` as a guard because I think it's more idiomatic.


sorentwo (approved):

sorentwo:
Nice catch, thanks!

PR #910, opened 2023-06-16
Fix code sample for expected failures

brandonjoyce:
Tiny code sample fix from the expected failures guide.


sorentwo (approved):

sorentwo:
Thanks for the fix!

PR #911, opened 2023-06-21
Add primary_key explicitly to SQLite3

tverlaan:
If a user has configured Ecto's `:migration_primary_key` to something different than bigserial the schema will not be compatible with the way Oban works.

This change doesn't warrant a new migration since either users have Ecto's default resulting in the same change or they don't have Ecto's default in which case Oban isn't functional.


sorentwo (approved):

sorentwo:
Thanks for the fix!

PR #914, opened 2023-06-29
Add oban instance module

yordis:


yordis:
## Note

I used `Oban.Instance` just to make a different module to simplify things, but maybe it should be under `Oban.__using__/1` instead.

lib/oban/instance.ex
```diff
@@ -0,0 +1,224 @@
...
+              {#{inspect(__MODULE__)}, prefix: "special"}
+            ]
+
+            opts = [strategy: :one_for_one, name: MyApp.Supervisor]
+            Supervisor.start_link(children, opts)
+          end
+      """
+      def child_spec(opts) do
+        unquote(child_opts)
+        |> Keyword.merge(Application.get_env(unquote(otp_app), __MODULE__, []))
```
sorentwo, line 70: I'm unsure how to resolve the combination of application config, child spec options, and module-level options, but I'd like something similar to `Ecto.Repo.init/1` that allows people to break complex config out of `runtime.exs` or `config.exs`.

For example, with a callback like this:

```elixir
  def init(opts) do
    base = [
      engine: Oban.Pro.Queue.SmartEngine,
      repo: MyApp.Repo,
      plugins: plugins()
    ]

    Keyword.merge(base, opts)
  end

  defp plugins do
    [
      {Oban.Plugins.Cron, crontab: crontab()},
      {Oban.Plugins.Gossip, []},
      {Oban.Pro.Plugins.DynamicLifeline, []},
      {Oban.Pro.Plugins.DynamicPruner, pruner_overrides()},
      {Oban.Pro.Plugins.DynamicQueues, queues: queues()},
      {Oban.Pro.Plugins.DynamicScaler, scalers: scalers()},
      {Oban.Web.Plugins.Stats, []}
    ]
  end
```

sorentwo (commented):
Thanks for starting this conversation @yordis. I've left some initial feedback and I can see this evolving into a nice solution for onboarding, config simplification, and managing multiple instances more easily (like your case).

lib/oban/instance.ex
```diff
@@ -0,0 +1,224 @@
...
+  Instead of calling `Oban` functions passing the `Oban` instance, you can use the `Oban.Instance` module directly:
+
+      Oban.Instance.insert(MyApp.MyJob.new(args: %{ "field" => "value" }))
+  """
+  defmacro __using__(opts \\ []) do
+    {opts, child_opts} = Keyword.split(opts, [:otp_app])
+    otp_app = Keyword.fetch!(opts, :otp_app)
+
+    quote do
+      @doc """
```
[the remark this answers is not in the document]
yordis: For most apps I work with, this documentation must be honest.

The anti-pattern is to hide them using `@doc false` because these macros can not tell what functions are available, so people generate the docs and see the public API and the type spec and help Dialyzer-like tools.

You can use the docs to copy-paste things or narrow the app's focus to a particular module since they are juggling too many things at once. I know the typespec and, hopefully, some sample code is crucial. Especially as the application grows and you have less time to Memorize things.

So I would stick to documenting as much as possible, given that they are public APIs.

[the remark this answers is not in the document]
sorentwo: > For most apps I work with, this documentation must be honest.

I'm not suggesting the documentation shouldn't be honest or really `@doc false`, but a thin pointer to the actual function as you've done below.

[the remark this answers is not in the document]
yordis: Just rereading this, what was I thinking using the word "honest" I guess I meant "Hosted"... 🤦🏻 

Regardless, let me work on it!

lib/oban/instance.ex
yordis: My two cents, I would wait until people develop use cases that grant such an `init` callback.

Ecto was done before `runtime.exs`, so it makes sense that it is part of it, but nowadays; (at least personally), I haven't found much need for such a style since people could use `runtime.exs` to read from the environment variables if that is what they want (which for the most part is what the Ecto example shows).

I am not opposed to it, but I rather wait, the worst part is coming up with a naming and how it suppose to work but we already have a reference.

yordis:
@sorentwo any updates over here?

test/oban_test.exs
```diff
@@ -489,6 +490,13 @@ defmodule ObanTest do
     end
   end
 
+  test "config/0 of a given instance" do
+    start_supervised!({MyOban, []})
+    assert config = MyOban.config()
+    assert config.name == MyOban
+    assert config.repo == Oban.Test.Repo
```
sorentwo, line 497:
```suggestion
    start_supervised!({MyOban, []})
    
    assert %{name: MyOban, repo: Oban.Test.Repo} = MyOban.config()
```

lib/oban.ex
```diff
@@ -1100,4 +1103,258 @@ defmodule Oban do
...
+            use Application
+
+            children = [
+              {#{inspect(__MODULE__)}, prefix: "special"}
+            ]
+
+            opts = [strategy: :one_for_one, name: MyApp.Supervisor]
+            Supervisor.start_link(children, opts)
+          end
+      """
```
sorentwo, line 1169: I don't feel enough _applications_ generate documentation for us to include `@doc` or `@spec` for generated functions. Some of the specs such as `insert_all` are already complicated and difficult to verify exhaustively, and I'd rather not duplicate them.

sorentwo (changes requested):
After a long hiatus, I looked over this again. Having everything in the main `Oban` module looks good.

I still believe it's a valuable addition for applications that run multiple instances, but as that's an outlier I'd like to restrict the longer-term maintenance burden and duplication.

yordis:
@sorentwo removed the doc and spec. Let me know what's next

lib/oban.ex
```diff
@@ -14,6 +14,9 @@ defmodule Oban do
   alias Oban.{Config, Engine, Job, Midwife, Notifier, Peer, Registry, Stager}
   alias Oban.Queue.{Drainer, Producer}
 
+  @typedoc """
+  The name of the Oban instance. This is used to identify the instance configuration in the application environment.
+  """
```
sorentwo, line 19:
```suggestion
  The name of an Oban instance. This is used to identify instances in the internal registry for
  configuration lookup.
  """
```

lib/oban.ex
```diff
@@ -1100,4 +1103,127 @@ defmodule Oban do
       raise ArgumentError, "queue #{inspect(queue)} does not exist locally"
     end
   end
+
+  @doc """
+  It creates a facade for the `Oban` functions.
```
sorentwo, line 1108:
```suggestion
  Creates a facade for `Oban` functions and automates fetching configuration from the application
  environment.
```

sorentwo (approved):
This is looking good. I have some documentation suggestions, some of which are complicated enough that they don't work in suggestions. I'm going to merge this and polish the documentation locally.

Thanks!

PR #924, opened 2023-07-20
Do not start a write transaction when there is nothing to write

tverlaan:
The action is a no-op, but SQLite sees this as a write transaction potentially blocking other writers. We should only write when there is something to write.

### DIFF
--- a/lib/oban/engines/lite.ex
+++ b/lib/oban/engines/lite.ex
@@ -128,9 +128,11 @@ defmodule Oban.Engines.Lite do
 
     staged = Repo.all(conf, select_query)
 
-    Repo.update_all(conf, where(Job, [j], j.id in ^Enum.map(staged, & &1.id)),
-      set: [state: "available"]
-    )
+    unless staged == [] do
+      Repo.update_all(conf, where(Job, [j], j.id in ^Enum.map(staged, & &1.id)),
+        set: [state: "available"]
+      )
+    end
 
     {:ok, staged}
   end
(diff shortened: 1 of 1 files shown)


tverlaan:
There was a short discussion here: https://elixir-lang.slack.com/archives/CRDGH1XCM/p1689843949809469

lib/oban/engines/lite.ex
```diff
@@ -128,9 +128,11 @@ defmodule Oban.Engines.Lite do
...
     staged = Repo.all(conf, select_query)
 
-    Repo.update_all(conf, where(Job, [j], j.id in ^Enum.map(staged, & &1.id)),
-      set: [state: "available"]
-    )
+    unless staged == [] do
+      Repo.update_all(conf, where(Job, [j], j.id in ^Enum.map(staged, & &1.id)),
+        set: [state: "available"]
+      )
+    end
```
sorentwo, line 135: Let's invert the conditional from "unless something is empty" to "if something is there":

```suggestion
    if Enum.any?(staged) do
      Repo.update_all(conf, where(Job, [j], j.id in ^Enum.map(staged, & &1.id)),
        set: [state: "available"]
      )
    end
```

lib/oban/engines/lite.ex
```diff
@@ -150,7 +152,9 @@ defmodule Oban.Engines.Lite do
 
     pruned = Repo.all(conf, select_query)
 
-    Repo.delete_all(conf, where(Job, [j], j.id in ^Enum.map(pruned, & &1.id)))
+    unless pruned == [] do
+      Repo.delete_all(conf, where(Job, [j], j.id in ^Enum.map(pruned, & &1.id)))
+    end
```
sorentwo, line 157:
```suggestion
    if Enum.any?(pruned) do
      Repo.delete_all(conf, where(Job, [j], j.id in ^Enum.map(pruned, & &1.id)))
    end
```

sorentwo (approved):
Thanks for the PR. Hopefully this alleviates the contention some 👍

PR #929, opened 2023-07-24
Fix module typo in the migration doc.

mrdotb:
Hello,

I found that the doc example was not working `Oban.Migration` does not exist.


sorentwo (approved):

sorentwo:
Thanks!

PR #934, opened 2023-08-01
Document error behavior for Oban.insert_all/3

egze:
Add more information what will happen if `Oban.insert_all/3` fails.

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -419,6 +419,17 @@ defmodule Oban do
   1. This function always returns a list rather than a tuple of `{count, records}`
   2. This function accepts a list of changesets rather than a list of maps or keyword lists
 
+  > #### 🌟 Error handling and rollbacks {: .info}
+  >
+  > If the insert query encounters an issue, the function will raise an error based on your database adapter. This behavior proves valuable, especially when used in conjunction with `c:Ecto.Repo.transaction/2`, allowing for rollbacks.
+  >
+  > Example:
+  >  ```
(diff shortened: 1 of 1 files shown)


lib/oban.ex
```diff
@@ -419,6 +419,17 @@ defmodule Oban do
...
+  > #### 🌟 Error handling and rollbacks {: .info}
+  >
+  > If the insert query encounters an issue, the function will raise an error based on your database adapter. This behavior proves valuable, especially when used in conjunction with `c:Ecto.Repo.transaction/2`, allowing for rollbacks.
+  >
+  > Example:
+  >  ```
+  >  ** (Postgrex.Error) ERROR 42501 (insufficient_privilege) permission denied for table oban_jobs
+  >  # or
+  >  ** (Ecto.InvalidChangesetError) could not perform insert because changeset is invalid.
+  >  ```
```
sorentwo, line 431: There aren't any other admonition blocks in the Oban module documentation, and this isn't important enough to be the first. I've also edited the paragraph and applied wrapping to match the rest of the module.

```suggestion
  #### Error Handling and Rollbacks
  
  If `insert_all` encounters an issue, the function will raise an error based on your database
  adapter. This behavior is valuable in conjunction with `c:Ecto.Repo.transaction/2` because it
  allows for rollbacks.
  
  For example, an invalid changeset raises:

  `* (Ecto.InvalidChangesetError) could not perform insert because changeset is invalid.`
```

sorentwo:
Thanks!

PR #964, opened 2023-10-12
Typo fix in Backoff.exponential/2 documentation

ipuchkov:
Currently `Oban.Backoff.exponential/2` function documentation refers that function calculates backoff in `millseconds` which is not true, as [Engine](https://github.com/sorentwo/oban/blob/b3b2bddf2592561e3136ba5a65c0e8d08a0f98e3/lib/oban/engines/basic.ex#L199) expects to receive backoff in seconds.

This typo introduces confusion in understanding of how `Job.scheduled_at` timestamp is calculated.

### DIFF
--- a/lib/oban/backoff.ex
+++ b/lib/oban/backoff.ex
@@ -4,7 +4,7 @@ defmodule Oban.Backoff do
   @type jitter_mode :: :inc | :dec | :both
 
   @doc """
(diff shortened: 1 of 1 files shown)


sorentwo:
Thanks for the fix! Accuracy is always welcome, even for a private module 🙂

PR #975, opened 2023-10-21
Allow unique's period option to take units.

JonRowe:
Other parts of oban (e.g. ratelimiting) can take period in the form `{n, unit}` and this is missing from the unique `:period` option, I mentioned this on the Elixir Slack a while back and was encouraged to open an issue about it but I thought I would go one better and create a PR! Totally don't mind if you'd prefer to do this yourself in a different way.

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -221,7 +221,7 @@ defmodule Oban.Job do
 
   Insert a job, ensuring that it is unique within the past minute:
 
-      MyApp.Worker.new(%{id: 1}, unique: [period: 60])
+      MyApp.Worker.new(%{id: 1}, unique: [period: {1, :minute}])
 
   Insert a unique job where the period is compared to the `scheduled_at` timestamp rather than
   `inserted_at`:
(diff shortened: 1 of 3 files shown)


lib/oban/job.ex
```diff
@@ -366,12 +366,34 @@ defmodule Oban.Job do
...
+
   @doc false
   @spec valid_unique_opt?({:fields | :period | :states, [atom()] | integer()}) :: boolean()
   def valid_unique_opt?({:fields, [_ | _] = fields}), do: fields -- [:meta | @unique_fields] == []
   def valid_unique_opt?({:keys, []}), do: true
   def valid_unique_opt?({:keys, [_ | _] = keys}), do: Enum.all?(keys, &is_atom/1)
   def valid_unique_opt?({:period, :infinity}), do: true
+
+  def valid_unique_opt?({:period, {period, unit}}),
+    do: is_integer(period) and period > 0 and unit in @valid_period_units
```
sorentwo, line 395: Nitpick, but we don't use this style anywhere else:

```suggestion
  def valid_unique_opt?({:period, {period, unit}}) do
    is_integer(period) and period > 0 and unit in @valid_period_units
  end
```

sorentwo (approved):
Excellent work! And here I only hoped you'd open an issue 🙂

PR #980, opened 2023-10-30
bugfix(cron): handle invalid range expressions

milmazz:
This PR should fix those cases where the cron expressions include invalid ranges or at least the cases where the left side is greater than the right one.

### DIFF
--- a/lib/oban/cron/expression.ex
+++ b/lib/oban/cron/expression.ex
@@ -164,7 +164,7 @@ defmodule Oban.Cron.Expression do
           rmin..rmax
         else
           raise ArgumentError,
-                "left side (#{rmin}) of the range must be lower or equal than right side (#{rmax})"
+                "left side (#{rmin}) of a range must be less than or equal to the right side (#{rmax})"
         end
     end
   end
--- a/lib/oban/cron/expression.ex
+++ b/lib/oban/cron/expression.ex
@@ -110,12 +110,23 @@ defmodule Oban.Cron.Expression do
 
   defp parse_part(part, range) do
     cond do
-      part == "*" -> range
-      part =~ ~r/^\d+$/ -> parse_literal(part)
-      part =~ ~r/^\*\/[1-9]\d?$/ -> parse_step(part, range)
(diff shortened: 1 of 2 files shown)


lib/oban/cron/expression.ex
```diff
@@ -110,12 +110,23 @@ defmodule Oban.Cron.Expression do
 
   defp parse_part(part, range) do
     cond do
-      part == "*" -> range
-      part =~ ~r/^\d+$/ -> parse_literal(part)
-      part =~ ~r/^\*\/[1-9]\d?$/ -> parse_step(part, range)
-      part =~ ~r/^\d+(\-\d+)?\/[1-9]\d?$/ -> parse_range_step(part, range)
-      part =~ ~r/^\d+\-\d+$/ -> parse_range(part, range)
-      true -> raise ArgumentError, "unrecognized cron expression: #{part}"
+      part == "*" ->
```
sorentwo, line 113: The extra space here makes it more readable. Are there any other changes to the conditions? My visual regex diffing can't spot anything 🙂

lib/oban/cron/expression.ex
```diff
@@ -146,7 +157,15 @@ defmodule Oban.Cron.Expression do
...
       [rmin, rmax] ->
-        String.to_integer(rmin)..String.to_integer(rmax)
+        rmin = String.to_integer(rmin)
+        rmax = String.to_integer(rmax)
+
+        if rmin <= rmax do
+          rmin..rmax
+        else
+          raise ArgumentError,
+                "left side (#{rmin}) of the range must be lower or equal than right side (#{rmax})"
```
sorentwo, line 167:
```suggestion
                "left side (#{rmin}) of a range must be less than or equal to the right side (#{rmax})"
```

sorentwo (approved):
Great fix!

lib/oban/cron/expression.ex
milmazz: I wasn't sure about pushing that change because it was done automatically by my editor, but after seeing the result, it was more readable, as you also noticed; I'm glad you accepted the change.

> Are there any other changes to the conditions? 

No, the 🥩 of the fix was in the `parse_range/2` helper.

Issue #836, opened 2023-02-01
Support for MySQL


pepicrft:
### Is your feature request related to a problem? Please describe.
I'm trying to use Oban in a Phoenix app that's setup to use MySQL the database solution. However, it doesn't work because the only two adapters provided are for SQLite and Postgres

### Describe the Solution You'd Like
It'd be great if MySQL was supported to. I'd be happy to contribute this feature myself with some guidance from the maintainers of the repository. Is there any technical limitation that would prevent this feature's existence?

### Describe Alternatives You've Considered
I considered using Postgres. However, I'd like to continue using MySQL because I'm using [PlanetScale](https://planetscale.com/) to run and scale the database in production.

sorentwo:
@pepicrft SQLite was _just_ added in the previous release, and that release isn't a week old yet. After the dust settles and we have reports of SQLite usage in the wild we can explore supporting MySQL.

pepicrft:
Thanks a lot @sorentwo for your answer. That makes a lot of sense. I didn't know that the support of SQLite has just been released.

sorentwo:
After some deliberation, we won't be adding MySQL support to Oban core. Adding MySQL through a separate package that defines an engine, peer, notifier, and migrations is possible if anybody is up to the challenge 🙂

jclem:
@sorentwo Makes sense! Keeping the core of Oban focused is a good thing. Is there anything from your deliberations you can share that may help folks looking to implement an Oban engine etc. for MySQL in the future?

sorentwo:
> Is there anything from your deliberations you can share that may help folks looking to implement an Oban engine etc. for MySQL in the future?

Implementing a new engine, peer, and migrations should be straightforward if you follow the `Lite` engine's pattern. My deliberation was about time core competency, time commitments, and supporting Oban's existing users/customers.

hubertlepicki:
To reanimate this question, @sorentwo , the "straightforward" part would be making an engine for *Oban* and not *Oban Pro*, right? I think supporting all the Oban Pro features would require more work and I don't think it has abstractions to do it cleanly now, would have to modify oban pro itself which is obviously not open source

sorentwo:
@hubertlepicki That's correct. Making a "basic" engine for MySQL, like the existing ones for Postgres and SQLite should be straightforward. That would not work with most existing Pro features, much like SQLite doesn't either.

PR #986, opened 2023-11-12
feat(migrations): allow a 3rd party migrator to be configured

hamptokr:
Thanks for the wonderful library!

I'm working on an external package that will provide a MySQL adapter for Oban (#836) and I need to be able to override the migrator used. This change will allow a `migrator` key in the Repo's config to allow for that:

```elixir
config :oban_mysql, ObanMySQL.Test.MySQLRepo,
  priv: "test/support/mysql",
  url: System.get_env("DATABASE_URL") || "mysql://root@localhost/oban_mysql_test",
  migrator: ObanMySQL.Migrations.MySQL,
  pool: Ecto.Adapters.SQL.Sandbox
```

If you have a different name suggestion or have a different approach in mind, please let me know! 🙂 

The very very WIP repo is over here: https://github.com/hamptokr/oban_mysql
With this change I was able to get the migrations running and the migration tests passing. I'm just starting to port over the engine and its test suite.

### DIFF
--- a/lib/oban/migration.ex
+++ b/lib/oban/migration.ex
@@ -194,7 +194,7 @@ defmodule Oban.Migration do
     case repo().__adapter__() do
       Ecto.Adapters.Postgres -> Oban.Migrations.Postgres
       Ecto.Adapters.SQLite3 -> Oban.Migrations.SQLite
-      _ -> repo().config()[:migrator]
+      _ -> Map.fetch!(repo().config(), :migrator)
     end
   end
 end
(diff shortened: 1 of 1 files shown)


lib/oban/migration.ex
```diff
@@ -194,6 +194,7 @@ defmodule Oban.Migration do
     case repo().__adapter__() do
       Ecto.Adapters.Postgres -> Oban.Migrations.Postgres
       Ecto.Adapters.SQLite3 -> Oban.Migrations.SQLite
+      _ -> repo().config()[:migrator]
```
sorentwo, line 197: Having thought about this for a few days, I don't see a cleaner alternative.

* Config doesn't help because Oban isn't running at this point.
* Protocols don't work because adapters are modules, not structs.
* Sticking key on the `config/0` map is much better than some global configuration

We don't want this to return `nil` silently, so at a minimum, this should use `Map.fetch!/2`:

```suggestion
      _ -> Map.fetch!(repo().config(), :migrator)
```

PR #990, opened 2023-11-20
fix sqlite link in installation.md

ruslandoga:
👋

This PR fixes the reference to point to `ecto_sqlite3` instead of `sqlite` which is undefined: https://github.com/sorentwo/oban/blob/b092e278c1091319fa48d6f9d9b445ffff57edb5/guides/installation.md?plain=1#L111


sorentwo:
Thanks!

PR #1009, opened 2024-01-05
Support Oban.insert_all from Stream

iamvery:
As discussed in email, here is a WIP for streaming support for `Oban.insert_all`. I came about the changes thus far by TDD. Thus far they seem to be merely ceremonial and without any real improvement.

Specifically, while the interface now support accepting a stream, the library still ultimately converts the stream to an Enum before performing the bulk insert, so I'm skeptical there would be any memory savings as-is based on the original problem we emailed about.

Do you agree?

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -42,7 +42,7 @@ defmodule Oban do
   @type option ::
           {:dispatch_cooldown, pos_integer()}
           | {:engine, module()}
-          | {:get_dynamic_repo, nil | (-> pid() | atom())}
+          | {:get_dynamic_repo, nil | (() -> pid() | atom())}
           | {:log, false | Logger.level()}
           | {:name, name()}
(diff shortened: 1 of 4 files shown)


sorentwo:
> Specifically, while the interface now support accepting a stream, the library still ultimately converts the stream to an Enum before performing the bulk insert, so I'm skeptical there would be any memory savings as-is based on the original problem we emailed about.

There wouldn't be any memory savings for the Lite and Basic, but there would be for the Smart engine because it supports auto-batching bulk inserts. Regardless, supporting streams makes inserting more flexible.

sorentwo:
Thanks for taking an initial shot at this. The various clauses and typespecs for `insert_all` are complicated and will benefit from development alongside `oban_pro`, so I'm going to tackle this one myself. I appreciate the proof of concept and your enthusiasm 🙂

PR #1011, opened 2024-01-09
Add missing discard cases to expected test message

kelostrada:
I noticed these were missing in `perform_job` result check


sorentwo:
Returning `:discard` or `{:discard, reason}` from a job is deprecated, undocumented, and not encouraged, so it isn’t listed as a possibility in the error message.

PR #1012, opened 2024-01-09
Format code

kelostrada:
Code seemed not to be formatted correctly


sorentwo:
Formatting is enforced in CI using the latest configured version. Perhaps you checked with an older elixir version that formatted differently? Either way, formatting is a-ok here!

PR #1017, opened 2024-01-16
Update upgrade guide to reference the new down migration in version 12

thejohncotton:
Noticed this when implementing an upgrade, looks like there is indeed a down migration for version 12 - I believe this is just a typo.


sorentwo:
Thank you!

PR #1029, opened 2024-02-06
Update installation instructions for 2.17

bamorim:
Seems like 2.17 got a new schema version, so I updated installation instructions <3

The rationale is that for brand new installations we probably want to "opt-in" for the migration. If that's not the case the instructions might need more complex instructions.


sorentwo:
Major oversight on our part, thanks!

PR #1034, opened 2024-02-13
Fix typo in splitting queues recipe

bhicks:
I noticed a minor typo in the _Splitting Queues_ recipe.

Thanks for all of the work on Oban!


sorentwo:
Thanks!

PR #1044, opened 2024-02-22
Improve docs on retrying job in case of an uncaught exception, exit or raise

drobnyd:


sorentwo (approved):
This is a welcome improvement. Once you take it out of draft mode I'll merge it in 👍

PR #1049, opened 2024-02-28
Include cron indicator and expression in meta

### LINKED ISSUE #1048: Crontab information in Oban jobs (or job-related Telemetry events)

I’m working on integrating Oban with the [Sentry SDK for Elixir](https://github.com/getsentry/sentry-elixir), in particular support for [Sentry crons](https://docs.sentry.io/product/crons/). To me, the obvious path forward is using Oban's Telemetry events to report started and finished jobs to Sentry.

Sentry's Crons require information about the cron jobs to work effectively. That includes the crontab, schedule, and whatnot. Currently, there doesn't seem to be data about that in the job or in the Telemetry event. I know that the Oban config includes 

```elixir
plugins: [
  {Oban.Plugins.Cron, [crontab: [...]]},
```

but getting from there to the actual crontab of the specific job is not ideal, and I’m not sure if it would be reliable.

Parsing the config, but as I said I’m not quite sure this would work and would be reliable; any advice appreciated 🙃 

Thanks all the great work peeps! 💟

sorentwo:
When the Cron plugin inserts jobs the original, unnormalized cron expression is now stored in a job's meta under the `cron_expr` key. For parity and backward compatibility with Pro's `DynamicCron`, meta also has `cron: true` injected.

Closes #1048

/cc @whatyouhide

### DIFF
--- a/lib/oban/plugins/cron.ex
+++ b/lib/oban/plugins/cron.ex
@@ -15,17 +15,19 @@ defmodule Oban.Plugins.Cron do
 
   Schedule various jobs using `{expr, worker}` and `{expr, worker, opts}` syntaxes:
 
-      config :my_app, Oban,
-        plugins: [
-          {Oban.Plugins.Cron,
-           crontab: [
-             {"* * * * *", MyApp.MinuteWorker},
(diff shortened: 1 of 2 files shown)


whatyouhide (approved):
Looks great! 😍

PR #1065, opened 2024-04-10
Include app name in `:pg` group namespace

warmwaffles:
We are running 4 separate elixir applications in the same cluster. Each application talks to each other using `:pg`. But each application has its own `Oban` instance running. We were running into issues where other nodes were showing up in the dashboard.

I tried setting just `name: MyApp.Oban` and on the dashboard router helper `oban_name: MyApp.Oban`. This did not work. Locally I made the following changes in my `deps/` directory, recompiled and this seems to do the trick.

If you have a better solution, feel free to close this. I just wanted to bring this up as potential solution.

### DIFF
--- a/lib/oban/notifiers/pg.ex
+++ b/lib/oban/notifiers/pg.ex
@@ -50,7 +50,7 @@ defmodule Oban.Notifiers.PG do
   @impl Notifier
   def notify(server, channel, payload) do
     with %{conf: conf} <- get_state(server) do
-      pids = :pg.get_members(__MODULE__, conf.prefix)
+      pids = :pg.get_members(__MODULE__, {conf.name, conf.prefix})
 
       for pid <- pids, message <- payload_to_messages(channel, payload) do
         send(pid, message)
@@ -65,7 +65,7 @@ defmodule Oban.Notifiers.PG do
     put_state(state)
 
     :pg.start_link(__MODULE__)
-    :pg.join(__MODULE__, state.conf.prefix, self())
+    :pg.join(__MODULE__, {state.conf.name, state.conf.prefix}, self())
 
     {:ok, state}
   end


warmwaffles:
I'm unsure if there are tests for this in the test directory.

lib/oban/notifiers/pg.ex
```diff
@@ -50,7 +50,7 @@ defmodule Oban.Notifiers.PG do
   @impl Notifier
   def notify(server, channel, payload) do
     with %{conf: conf} <- get_state(server) do
-      pids = :pg.get_members(__MODULE__, conf.prefix)
+      pids = :pg.get_members(__MODULE__, {conf.name, conf.prefix})
```
sorentwo, line 53: This would be a breaking change for any app that relies on connecting different apps that share an `oban_jobs` table. Instead, I think the solution is to pass a `namespace` option through and have it fall back to `conf.prefix`.

In application config it would look like this:

```elixir
notifier: {Oban.Notifiers.PG, namespace: {Oban, "public"}}
```

WDYT?

warmwaffles: Funny enough @mbaleczny suggested something like that as well.

warmwaffles: @sorentwo I don't see how this would effect two apps connected to the same database using the same table. If both have the same `name:` then they will still communicate as normal but if the `name:` is different, then they won't communicate.

Either way, I do agree that this should probably be a configuration done with the `:notifier`. I don't know where all of that would need to change as my understanding is that `:notifier` is expected to be a module only.

sorentwo: > I don't see how this would effect two apps connected to the same database using the same table. If both have the same name: then they will still communicate as normal but if the name: is different, then they won't communicate.

That's the issue—if people are relying on the fact that two instances with different names (`Oban.A` and `Oban.B`) can communicate, then this would break their app.

> I don't know where all of that would need to change as my understanding is that :notifier is expected to be a module only.

I can make the changes. It will look roughly like https://github.com/sorentwo/oban_notifiers_phoenix 🙂

warmwaffles: Cool, I'll close this PR then.

sorentwo:
@warmwaffles Done! See the linked commit.

PR #1075, opened 2024-04-26
fix: only use pids from `:"$ancestors"` list

### LINKED ISSUE #1074: Regression:  `Check process ancestry tree for test mode override` commit makes my tests fail
Hello,

This recent [commit](https://github.com/sorentwo/oban/commit/49ae042a4950a6106c82c86e34b21c4c979831c8) made my tests fail as it seems that the `:"$ancestors"` list can contain the name of a Supervisor which is not a Pid and the call to `Process.info` fails.

barnabasJ:
Fixes #1074


sorentwo:
Thanks for the PR, but this is already fixed with https://github.com/sorentwo/oban/commit/0abfe77e31b88c65e5fc31e58bd01a324dda4de3

Issue #1077, opened 2024-05-01
Node name not set from DYNO var with mix release on Heroku


eprothro:
### Environment

- Oban 2.17.8, Oban Pro 1.4.6
- PostgreSQL 15.6
- Elixir 1.16.2 / 26.2.4

### Current Behavior

This application deploys on Heroku. When switching from executing via `mix phx.server` to mix releases (via https://github.com/elixir-buildpack/heroku-buildpack with the `release=true` option), the node name in the Oban web ui changed from `worker.1` (e.g. the `DYNO` ENV value) to `app_name@hostname`. 

### Expected Behavior

It would be nice if Oban picked up the `DYNO` value at runtime.

### Workaround

Set `node` explicitly in `config/runtime.exs`. 

```
#config/runtime.exs

config :my_app, Oban,
  node: System.get_env("DYNO") || :inet.gethostname() |> elem(1) |> List.to_string(),
   ...
```

### Notes

Unlike, #170, we're not setting a hostname explicitly, the hostname that comes through is a uid set by Heroku on the VM.

sorentwo:
That's because the node is considered "alive" by default when it's part of a release, and the first condition uses the `node()` name if the node is alive. The behaviour was established at a time when releases weren't as ubiquitous and `runtime.exs` didn't exist. Changing the order to look for `DYNO` first feels backwards now.

I believe the best approach is to document setting the `node` option in `runtime.exs`. Would you open a PR?

sorentwo:
Closed with #1078

sorentwo:
Thanks!

PR #1078, opened 2024-05-02
add documentation for heroku node name

eprothro:
addresses #1077

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -400,11 +400,20 @@ defmodule Oban do
 
   ## Node Name
 
-  When the `node` value hasn't been configured it is generated based on the environment:
+  When the `node` value has not been configured it is generated based on the environment:
 
-  * In a distributed system the node name is used
-  * In a Heroku environment the system environment's `DYNO` value is used
-  * Otherwise, the system hostname is used
+  1. If the local node is alive (e.g. in a distributed system, or when running from a mix release)
+     the node name is used
+  2. In a Heroku environment the system environment's `DYNO` value is used
+  3. Otherwise, the system hostname is used
+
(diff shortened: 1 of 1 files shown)


lib/oban.ex
```diff
@@ -400,11 +400,20 @@ defmodule Oban do
...
+     the node name is used
+  2. In a Heroku environment the system environment's `DYNO` value is used
+  3. Otherwise, the system hostname is used
+
+  Note: When running a mix release on a Heroku node, the node is alive even if not part of a
+  distributed system. In order to use the `DYNO` value, configure the node value using runtime
+  configuration via `config/runtime.exs:
+
+        config :my_app, Oban,
+          node: System.get_env("DYNO") || :inet.gethostname() |> elem(1) |> List.to_string()
```
sorentwo, line 415: The fallback portion after the `||` isn't explained in the prelude, and I don't think it's necessary to explain how to set the `node`.

```suggestion
        config :my_app, Oban,
          node: System.get_env("DYNO", "nonode@nohost")
```

lib/oban.ex
```diff
@@ -400,11 +400,20 @@ defmodule Oban do
...
 
-  When the `node` value hasn't been configured it is generated based on the environment:
+  When the `node` value has not been configured it is generated based on the environment:
+
+  1. If the local node is alive (e.g. in a distributed system, or when running from a mix release)
+     the node name is used
+  2. In a Heroku environment the system environment's `DYNO` value is used
+  3. Otherwise, the system hostname is used
+
+  Note: When running a mix release on a Heroku node, the node is alive even if not part of a
```
sorentwo, line 410:
```suggestion
  When running a mix release on a Heroku node, the node is alive even if not part of a
```

sorentwo (approved):

PR #1084, opened 2024-05-16
Add support for only: [{job_status}] configuration in Oban.Plugins.Pruner

punchcafe:
Hello!

We've recently adopted Oban (it's great so far, so thanks!) and are looking at ways to manage our dead letter queue for discarded jobs. 

This PR would add the ability to specify which finished status to prune in the Basic Pruner, which would make dead lettering fairly trivial for projects which don't expect massive amounts of discards.

### DIFF
--- a/lib/oban/engines/basic.ex
+++ b/lib/oban/engines/basic.ex
@@ -150,10 +150,12 @@ defmodule Oban.Engines.Basic do
     limit = Keyword.fetch!(opts, :limit)
     time = DateTime.add(DateTime.utc_now(), -max_age)
 
+    prune_states = Keyword.get(opts, :only, ~w(completed cancelled discarded))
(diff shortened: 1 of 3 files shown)


sorentwo:
Thanks for the PR, but the basic pruner is purposefully bare bones. As noted in the [Oban.Plugins.Pruner](https://hexdocs.pm/oban/Oban.Plugins.Pruner.html) docs:

> This plugin limited to a fixed interval and a single max_age check for all jobs. To prune on a cron-style schedule, retain jobs by a limit or age, or provide overrides for specific queues, workers, and job states; see Oban Pro's [DynamicPruner](https://getoban.pro/docs/pro/Oban.Pro.Plugins.DynamicPruner.html).

PR #1086, opened 2024-05-17
Request to rename [:oban, :notifier, :switch] emitted log attribute `status` to `connectivity_status`

punchcafe:
Hello Again!

This is a small change request to make the default telemetry logger integrate nicer with Datadog.

The logs emitted in response to [:oban, :notifier, :switch] events have a status field, which is a reserved field for JSON logs with Datadog. The result was that all :switch events were being treated as critical logs and raising alerts. I imagine a lot of Oban users use Datadog so I thought it may be worth making this small tweak to avoid conflict.

Thanks,
Luca

### DIFF
--- a/lib/oban/telemetry.ex
+++ b/lib/oban/telemetry.ex
@@ -431,14 +431,14 @@ defmodule Oban.Telemetry do
         :isolated ->
           %{
             event: "notifier:switch",
-            status: status,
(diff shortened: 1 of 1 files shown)


sorentwo:
@punchcafe Thanks! This is a welcome change; whatever we can do to ease integration with external tools is wonderful.

PR #1103, opened 2024-06-13
fix: make `t:Oban.Job.replace_by_state_option/0` a keyword item instead of a keyword

fahchen:
According to the usage in the documentation, `replace_by_state_option` should be a keyword item. I believe this may be a mistake.

https://github.com/sorentwo/oban/blob/5668f23fc3dc99e09ac3cced5e8a0f22cffd8552/README.md?plain=1#L546-L553

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -63,7 +63,7 @@ defmodule Oban.Job do
           | :worker
         ]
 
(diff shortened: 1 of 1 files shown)


sorentwo:
You're correct, it was a mistake. Thanks!

PR #1121, opened 2024-07-25
Update testing_queues.md to be consistent with installation guide

esse:
look in: https://hexdocs.pm/oban/installation.html

Since by default jobs will be execute in testing mode - however it will be done in inline fashion, hence testing docs are wrong in that manner. This PR correct this.


sorentwo:
Thank you for the fix!

PR #1122, opened 2024-07-26
Fix: query!/4 should dispatch :query!

preciz:
Thank you for reviewing.


sorentwo:
Well spotted; thanks!

PR #1128, opened 2024-07-30
Sonar: fix pruning stale nodes

feld:


feld:
> The oldest supported version of Elixir that Oban supports is v1.13+. `Map.reject` was added in Elixir v1.13; there shouldn't be any issue.

My deployment was still building against 1.12, very sorry!

PR #1133, opened 2024-08-08
Update README.md

simonmcconnell:
Add clarity to unique conflicts when using Pro's Smart Engine.  This is described in the insert_all docs but was missing here.


sorentwo:
Thanks!

PR #1143, opened 2024-08-24
Add ErrorTracker to the guides

odarriba:
This change proposal adds [ErrorTracker](https://github.com/elixir-error-tracker/error-tracker) to the guide that lists integrations that use `Oban.Telemetry` to integrate out-of-the-box with Oban.

In the case of the ErrorTracker, it uses Telemetry events to catch and store context and exceptions of Oban jobs executed on the system.

### DIFF
--- a/guides/preparing_for_production.md
+++ b/guides/preparing_for_production.md
@@ -138,3 +138,7 @@ tools built on `Oban.Telemetry`:
 * [AppSignal](https://docs.appsignal.com/elixir/integrations/oban.html)—The AppSignal for Elixir
   package instruments jobs performed by Oban workers, and collects metrics about your jobs'
   performance.
+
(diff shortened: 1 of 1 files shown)


odarriba:
I was not sure if I should have add it to the Error tracking section, as it does not require any special module to handle the exceptions. 

Should I add a sentence there explaining that it can be done with the ErrorTracker library, or even linking to the file that we use as a Telemetry handler as an example?

Thanks for your awesome work on Oban 🙌

sorentwo:
> Should I add a sentence there explaining that it can be done with the ErrorTracker library, or even linking to the file that we use as a Telemetry handler as an example?

Yes please. Alongside Sentry and AppSignal would be excellent.

odarriba:
Updated! ❤️

sorentwo:
Thanks! 💛

PR #1145, opened 2024-08-27
Use proper fields for finding prunable jobs

ademenev:
Using "scheduled_at" is not correct. Depending on job state, one of "cancelled_at", "discarded_at" or "completed_at" should be used.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "editor.formatOnSave": true
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "editor.formatOnSave": true
-}
--- a/lib/oban/engines/basic.ex
+++ b/lib/oban/engines/basic.ex
@@ -153,7 +153,7 @@ defmodule Oban.Engines.Basic do
     subquery =
       queryable
       |> select([:id, :queue, :state])
-      |> where([j], j.state == "completed" and j.completed_at < ^time)
+      |> where([j], j.state == "completed" and j.scheduled_at < ^time)
       |> or_where([j], j.state == "cancelled" and j.cancelled_at < ^time)
       |> or_where([j], j.state == "discarded" and j.discarded_at < ^time)
       |> where([j], not is_nil(j.queue))
--- a/lib/oban/engines/basic.ex
+++ b/lib/oban/engines/basic.ex
@@ -153,9 +153,10 @@ defmodule Oban.Engines.Basic do
     subquery =
       queryable
       |> select([:id, :queue, :state])
-      |> where([j], j.state in ~w(completed cancelled discarded))
+      |> where([j], j.state == "completed" and j.completed_at < ^time)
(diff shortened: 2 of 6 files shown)


lib/oban/engines/basic.ex
```diff
@@ -153,9 +153,10 @@ defmodule Oban.Engines.Basic do
     subquery =
       queryable
       |> select([:id, :queue, :state])
-      |> where([j], j.state in ~w(completed cancelled discarded))
+      |> where([j], j.state == "completed" and j.completed_at < ^time)
```
sorentwo, line 156: We switched to the `scheduled_at` timestamp to make use of the compound index and speed up pruning. There are likely to be _many_ more `completed` jobs than `cancelled` or `discarded`, and a `completed` job has to execute to end up in that state. So, let's use `scheduled_at` for the `completed` state to make use of that index.

```suggestion
      |> where([j], j.state == "completed" and j.scheduled_at < ^time)
```

lib/oban/engines/lite.ex
```diff
@@ -146,8 +146,9 @@ defmodule Oban.Engines.Lite do
     select_query =
       queryable
       |> select([j], map(j, [:id, :queue, :state]))
-      |> where([j], j.state in ~w(completed cancelled discarded))
-      |> where([j], j.scheduled_at < ^time)
+      |> where([j], j.state == "completed" and j.completed_at < ^time)
```
sorentwo, line 149: Chances are a `lite` database doesn't have nearly as many jobs, but we should still use `scheduled_at` for consistency.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
sorentwo, line 1: Please remove this file ✂️

lib/oban/engines/basic.ex
ademenev: Yes, that makes perfect sense performance-wise

If you do not do some tricky stuff, completed, cancelled and discarded rows never change. Probably they  could all use a single timestamp field instead of 3, with a compound index on state and that field

ademenev: Isn't this redundant? queue column is not nullable

```elixir
where([j], not is_nil(j.queue))
```

sorentwo: The `is_nil` check is pointless with the current index structure, but it was required to force index usage with the old compound table where `queue` was before `state`. The index change was optional (way back when), and some people never chose that option.

We can probably remove it at this point, but it doesn't change the query plan.

sorentwo:
Thanks for submitting this improvement!

ademenev:
Would you consider this?

> If you do not do some tricky stuff, completed, cancelled and discarded rows never change. Probably they could all use a single timestamp field instead of 3, with a compound index on state and that field

May be a bit of work and require a heavy migration though

sorentwo:
> Would you consider this?

In theory, yes. It would require a few migrations for zero downtime and coordinated changes between packages. That's a lot and I'm not convinced it's worth it.

PR #1148, opened 2024-09-04
Add information on leadership changes to `oban.peer.election` event

notslang:
This is an attempt at including whether Oban leadership changed in the `oban.peer.election`

I added the key `was_leader`, which indicates if the current node was the leader before the election happened.

If `was_leader=false` and `leader=true` in a telemetry event, then we know that the current node just became the leader. If `was_leader=true` and `leader=false` then we know that the current node lost leadership.

Marking this as a draft because I haven't tested it out yet in my application. I think I can add some tests to `test/oban/peers/postgres_test.exs` and `test/oban/peers/global_test.exs` too. That seems to be where this telemetry event gets tested.

Also, lmk if you'd prefer a different name.

### DIFF
--- a/lib/oban/peers/global.ex
+++ b/lib/oban/peers/global.ex
@@ -90,13 +90,13 @@ defmodule Oban.Peers.Global do
 
   @impl GenServer
   def handle_info(:election, %State{} = state) do
-    meta = %{conf: state.conf, leader: state.leader?, peer: __MODULE__}
+    meta = %{conf: state.conf, leader: state.leader?, peer: __MODULE__, was_leader?: nil}
 
     locked? =
       :telemetry.span([:oban, :peer, :election], meta, fn ->
         locked? = :global.set_lock(key(state), nodes(), 0)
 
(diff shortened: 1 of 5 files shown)


sorentwo:
> Marking this as a draft because I haven't tested it out yet in my application. I think I can add some tests to test/oban/peers/postgres_test.exs and test/oban/peers/global_test.exs too. That seems to be where this telemetry event gets tested.

That's the right place.

> Also, lmk if you'd prefer a different name.

The `was_leader` name works for me. Most of our other predicates have a trailing question mark, so perhaps `was_leader?` instead?

sorentwo:
@notslang Do you have time to add the tests or would you like us to pick up this change?

notslang:
I think this is good to go now. Renamed `was_leader` -> `was_leader?` and tested it out in my application. Looking for events where `leader=true` and `was_leader?=false` seems to reliably detect nodes becoming leaders.

sorentwo:
Excellent, thanks!

PR #1169, opened 2024-10-30
Document notifier tradeoffs

nathanl:
Capturing insights from [this ElixirForum conversation](https://elixirforum.com/t/does-oban-cancel-job-2-stop-a-running-job-if-called-from-a-rolled-back-transaction/65010) and from an Oban support email.

### DIFF
--- a/lib/oban/notifier.ex
+++ b/lib/oban/notifier.ex
@@ -13,10 +13,7 @@ defmodule Oban.Notifier do
   an implementation of the `Oban.Notifier` behaviour. 
 
   * `Oban.Notifiers.Postgres` — A Postgres notifier that uses `LISTEN/NOTIFY` to broadcast
-    messages. This is the default, and the only one which respects database transactions.
-    For example, if you call `cancel_job` within a transaction which then gets rolled back, the
-    job will not be cancelled. However, this notifier is the least scalable because each notification
-    requires a separate query.
+    messages.
 
   * `Oban.Notifiers.PG` — A process groups notifier that relies on Distributed Erlang to broadcast
     messages.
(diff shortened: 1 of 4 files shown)


lib/oban/notifier.ex
```diff
@@ -13,7 +13,10 @@ defmodule Oban.Notifier do
   an implementation of the `Oban.Notifier` behaviour. 
 
   * `Oban.Notifiers.Postgres` — A Postgres notifier that uses `LISTEN/NOTIFY` to broadcast
-    messages. This is the default.
+    messages. This is the default, and the only one which respects database transactions.
+    For example, if you call `cancel_job` within a transaction which then gets rolled back, the
+    job will not be cancelled. However, this notifier is the least scalable because each notification
+    requires a separate query.
```
sorentwo, line 19:
```suggestion
    messages.
```

lib/oban/notifiers/postgres.ex
```diff
@@ -22,6 +22,11 @@ if Code.ensure_loaded?(Postgrex) do
 
     The notifications system is built on PostgreSQL's `LISTEN/NOTIFY` functionality. Notifications
     are only delivered **after a transaction completes** and are de-duplicated before publishing.
+    This means that notifications sent during a transaction will not be sent if the transaction is
+    rolled back, providing consistency; this is the only notifer which provides that guarantee.
+    However, it is not as scalable as other notifiers because because each notification requires a
+    separate query.
```
sorentwo, line 28:
```suggestion
    However, it is not as scalable as other notifiers because because each notification requires a
    separate query and notifications can't exceed 8kb.
```

sorentwo:
Thanks for this @nathanl. I removed the differentiation from the list of notifiers because the detail is in each of the modules, but otherwise left the additions intact.

PR #1170, opened 2024-10-30
Note that tradeoffs exist here

nathanl:
Followup to #1169 - I get that we don't want to rehearse the differences here, but think it's useful to let people know they should compare the options.


sorentwo:
Well stated. Thanks for the follow up!

PR #1175, opened 2024-11-07
Move "Periodic Jobs" section out of README into guide

whatyouhide:
This is the start of some work on https://github.com/oban-bg/oban/issues/850. We'll want to split the README out into guides, and the cron plugin seemed like a good candidate to start off—it's isolated and its functionality is self-contained.


guides/periodic_jobs.md
```diff
@@ -0,0 +1,119 @@
...
+  * `*` — Wildcard, matches any value (0, 1, 2, …)
+  * `0` — Literal, matches only itself (only 0)
+  * `*/15` — Step, matches any value that is a multiple (0, 15, 30, 45)
+  * `0-5` — Range, matches any value within the range (0, 1, 2, 3, 4, 5)
+  * `0-9/2` — Step values can be used in conjunction with ranges (0, 2, 4, 6, 8)
+
+Each part may have multiple rules, where rules are separated by a comma. The allowed values for
+each field are as follows:
+
+| Field      | Allowed Values                               |
```
whatyouhide, line 57: Turned some stuff into tables here, I think it looks great.

lib/oban/plugins/cron.ex
```diff
@@ -1,20 +1,19 @@
 defmodule Oban.Plugins.Cron do
   @moduledoc """
-  Periodically enqueue jobs through CRON based scheduling.
-
-  This plugin registers workers a cron-like schedule and enqueues jobs automatically. Periodic
-  jobs are declared as a list of `{cron, worker}` or `{cron, worker, options}` tuples.
+  Periodically enqueue jobs through [cron](https://en.wikipedia.org/wiki/Cron)-based scheduling.
```
whatyouhide, line 3: Moved some stuff around here, I like the flow now, thoughts?

guides/periodic_jobs.md
```diff
@@ -0,0 +1,119 @@
...
+
+  config :my_app, Oban,
+    repo: MyApp.Repo,
+    plugins: [
+      {Oban.Plugins.Cron,
+       crontab: [
+         {"* * * * *", MyApp.SlowWorker, args: custom_args, unique: unique_opts}
+       ]}
+    ]
+  ```
```
sorentwo, line 116: This is no longer true. As of Oban v2.18 cron doesn't use uniqueness at all, and for many versions it has avoided that problem by only scheduling at the top of the minute. Furthermore, only the leader schedules jobs.

guides/periodic_jobs.md
```diff
@@ -0,0 +1,119 @@
...
+
+  * `MyApp.MinuteWorker` — Inserted once every minute
+  * `MyApp.HourlyWorker` — Inserted at the first minute of every hour with custom arguments
+  * `MyApp.DailyWorker` — Inserted at midnight every day with no retries
+  * `MyApp.MondayWorker` — Inserted at noon every Monday in the "scheduled" queue
+  * `MyApp.AnotherDailyWorker` — Inserted at midnight every day with no retries
+
+The crontab format respects all [standard rules][cron] and has one minute resolution. Jobs are
+considered unique for most of each minute, which prevents duplicate jobs with multiple nodes and
+across node restarts.
```
sorentwo, line 32: Oops, this is outdated. The bit about "jobs are considered unique for most of each minute" isn't true. My comment below explains it a little more, but essentially the combination of leadership and only scheduling at the top of the minute prevents this.

sorentwo (commented):
A fantastic start! Everything you've done here is spot on and the direction we're hoping to go. I've left a few comments about outdated information that was ported from the README. This is ready to go once that's corrected/updated 🎉

whatyouhide:
@sorentwo updated ✅

lib/oban/plugins/cron.ex
sorentwo: Agreed, the new flow is an improvement!

guides/periodic_jobs.md
```diff
@@ -0,0 +1,99 @@
...
+
+The crontab would insert jobs as follows:
+
+  * `MyApp.MinuteWorker` — Inserted once every minute
+  * `MyApp.HourlyWorker` — Inserted at the first minute of every hour with custom arguments
+  * `MyApp.DailyWorker` — Inserted at midnight every day with no retries
+  * `MyApp.MondayWorker` — Inserted at noon every Monday in the "scheduled" queue
+  * `MyApp.AnotherDailyWorker` — Inserted at midnight every day with no retries
+
+The crontab format respects all [standard rules][cron] and has one minute resolution. Jobs always run at the top of the minute. All jobs get scheduled by the leader.
```
sorentwo, line 30:
```suggestion
The crontab format respects all [standard rules][cron] and has one minute resolution. Jobs are
always inserted at the top by the leader node, which prevents duplicate jobs regardless of
cluster size, restarts, or crashes.
```

sorentwo (approved):

sorentwo:
An excellent start, thank you!

PR #1180, opened 2024-11-12
Split everything left out of the README

whatyouhide:
Oooph, this is the final one! I’m pretty happy with this one, excited to hear your thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ without worrying about jobs being lost or orphaned due to crashes.
 > [!TIP] Oban Web+Pro
 >
 > A web dashboard for managing Oban, along with an official set of extensions, plugins, and
-> workers that expand what Oban is capable are available as licensed packages:
+> workers that expand what Oban is capable of are available as licensed packages:
 > 
 > * [🧭 Oban Web](https://oban.pro#oban-web)
 > * [🌟 Oban Pro](https://oban.pro#oban-pro)
@@ -176,7 +176,7 @@ and configuring Oban in your application.
        queues: [mailers: 20]
      ```
 
-  1. Define a worker to process jobs in the `mailers` queue (see `Oban.Worker`):
+  2. Define a worker to process jobs in the `mailers` queue (see `Oban.Worker`):
 
      ```elixir
      defmodule MyApp.MailerWorker do
@@ -190,7 +190,7 @@ and configuring Oban in your application.
      end
      ```
 
-  1. Enqueue a job (see [the documentation](https://hexdocs.pm/oban/Oban.Job.html#enqueueing-jobs)):
+  3. Enqueue a job (see [the documentation](https://hexdocs.pm/oban/Oban.Job.html#enqueueing-jobs)):
 
      ```elixir
      %{email: %{to: "foo@example.com", body: "Hello from Oban!"}}
@@ -198,7 +198,7 @@ and configuring Oban in your application.
      |> Oban.insert()
      ```
 
-  1. The magic happens! Oban executes the job when there is available bandwidth in the
+  4. The magic happens! Oban executes the job when there is available bandwidth in the
      `mailer` queue.
 
 <!-- MDOC -->
--- a/README.md
+++ b/README.md
@@ -33,22 +33,11 @@
 
 - [Features](#features)
 - [Oban Web+Pro](#oban-webpro)
-- Usage
-  - [Engines](#engines)
-  - [Requirements](#requirements)
-  - [Learning](#learning)
-  - [Installation](#installation)
-  - [Configuring Queues](#configuring-queues)
-  - [Defining Workers](#defining-workers)
-  - [Enqueueing Jobs](#enqueueing-jobs)
-  - [Scheduling Jobs](#scheduling-jobs)
-  - [Prioritizing Jobs](#prioritizing-jobs)
-  - [Unique Jobs](#unique-jobs)
-  - [Pruning Historic Jobs](#pruning-historic-jobs)
-  - [Periodic Jobs](https://hexdocs.pm/oban/periodic_jobs.html)
-  - [Error Handling](https://hexdocs.pm/oban/error_handling.html)
-  - [Instrumentation and Logging](https://hexdocs.pm/oban/instrumentation.html)
-  - [Instance and Database Isolation](https://hexdocs.pm/oban/isolation.html)
+- [Learning](#learning)
+- [Requirements](#requirements)
+- [Engines](#engines)
+- [Installation](#installation)
+- [Quick Getting Started](#quick-getting-started)
 - [Community](#community)
 - [Contributing](#contributing)
 
@@ -135,7 +124,8 @@ without worrying about jobs being lost or orphaned due to crashes.
 
 ## Oban Web+Pro
 
-> [!TIP]
+> [!TIP] Oban Web+Pro
+>
 > A web dashboard for managing Oban, along with an official set of extensions, plugins, and
 > workers that expand what Oban is capable are available as licensed packages:
 > 
@@ -158,7 +148,7 @@ suitability for production.
 
 ## Requirements
 
-Oban requires: 
+Oban requires:
 
 * Elixir 1.14+
 * Erlang 24+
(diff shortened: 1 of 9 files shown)


guides/operational_maintenance.md
```diff
@@ -0,0 +1,28 @@
+# Operational Maintenance
```
whatyouhide, line 1: Love this name. Thoughts?

lib/oban/worker.ex
```diff
@@ -254,6 +269,78 @@ defmodule Oban.Worker do
...
+
+  Occasionally you may need to insert a job for a worker that exists in another
+  application. In that case you can use `Oban.Job.new/2` to build the changeset
+  manually:
+
+      %{id: 1, user_id: 2}
+      |> Oban.Job.new(queue: :default, worker: OtherApp.Worker)
+      |> Oban.insert()
+
+  ## Scheduling Jobs
```
whatyouhide, line 283: I think scheduling jobs and prioritizing jobs belong here; thoughts?

README.md
```diff
@@ -163,417 +152,42 @@ Oban requires Elixir 1.14+, Erlang 23+, and PostgreSQL 12.0+ or SQLite3 3.37.0+.
...
-MyApp.MyWorker.new(args, queue: dynamic_queue)
-```
-
-Successful jobs should return `:ok` or an `{:ok, value}` tuple. The value returned from
-`perform/1` is used to control whether the job is treated as a success, a failure, cancelled or
-deferred for retrying later.
-
-See the `Oban.Worker` docs for more details on failure conditions and `Oban.Telemetry` for details
-on job reporting.
+## Quick Getting Started
```
whatyouhide, line 155: Now, we have this section which kind of "shows off" Oban, but then you have real docs and guides for getting into more detail. The README got real slim with this.

guides/operational_maintenance.md
sorenone: Love it!

guides/job_uniqueness.md
```diff
@@ -0,0 +1,131 @@
...
+> UniqueWorker.new(new_args, replace: [executing: [:args]])
+> ```
+>
+> then Oban will update `:args`, but the job will continue executing with the original value.
+
+## Strong Guarantees
+
+Oban **guarantees** uniqueness of jobs through transactional locks and database queries.
+Uniqueness *does not* rely on unique constraints in the database. This makes uniqueness entirely
+configurable by application code, without the need for database migrations.
```
sorenone, line 129:
```suggestion
## Unique Guarantees

Oban **strives** for uniqueness of jobs through transactional locks and database queries.
Uniqueness *does not* rely on unique constraints in the database, which leaves it prone to race
conditions in some circumstances.
```

lib/oban/plugins/pruner.ex
```diff
@@ -1,9 +1,9 @@
 defmodule Oban.Plugins.Pruner do
   @moduledoc """
-  Periodically delete `completed`, `cancelled` and `discarded` jobs based on their age.
+  Periodically delete `completed`, `cancelled`, and `discarded` jobs based on their age.
 
   Pruning is critical for maintaining table size and continued responsive job processing. It
-  is recommended for all production applications.
+  is recommended for all production applications. See also the [*Operational Maintenance* guide](operational_maintenance.html).
```
sorenone, line 6:
```suggestion
  is recommended for all production applications. See also the
  [*Operational Maintenance* guide](operational_maintenance.html).
```

lib/oban/worker.ex
```diff
@@ -254,6 +269,78 @@ defmodule Oban.Worker do
           if MyApp.something?(job), do: :ok, else: {:snooze, 60}
         end
       end
+
+  ## Workers in A Different Application
+
+  Occasionally you may need to insert a job for a worker that exists in another
```
sorenone, line 275:
```suggestion
  Occasionally, you may need to insert a job for a worker that exists in another
```

lib/oban/worker.ex
sorenone: Scheduling might not make as much sense to be here. Maybe it does? Do you feel it gets lost here? Prioritizing belongs here, because it's an option you define on the worker.

lib/oban/worker.ex
```diff
@@ -254,6 +269,78 @@ defmodule Oban.Worker do
...
+  Local mode is less efficient and will only happen if you're running in an
+  environment where neither `Postgres` nor `PG` notifications work. That situation
+  should be rare and limited to the following conditions:
+
+    1. Running with a connection pooler, like [`pg_bouncer`][pg_bouncer],
+       in transaction mode.
+    2. Running without clustering, that is, without *distributed Erlang*.
+
+  If **both** of those criteria apply and PubSub notifications won't work, then
+  staging will switch to polling in `local` mode.
```
sorenone, line 320: Do you think there a better location for this section? Maybe a guide about distribution/pubsub/notifiers? You're in the head-space for it more than we are right now.

sorenone (commented):
Great work! The slimmed down Readme and quick Getting Started has been much improved. Please double-check the suggestions and formatting to ensure I haven't thwarted your efforts. ;)

lib/oban/worker.ex
whatyouhide: Yeah maybe scheduling is not the same... But I don't know, a page on its own? Maybe it makes sense. Let's go with that for now

README.md
```diff
@@ -138,7 +126,8 @@ orphaned due to crashes.
 
 ## Oban Web+Pro
 
-> [!TIP]
+> [!TIP] Oban Web+Pro
+>
 > A web dashboard for managing Oban, along with an official set of extensions, plugins, and
 > workers that expand what Oban is capable are available as licensed packages:
```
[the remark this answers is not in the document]
whatyouhide: Done.

README.md
```diff
@@ -163,417 +152,42 @@ Oban requires Elixir 1.14+, Erlang 23+, and PostgreSQL 12.0+ or SQLite3 3.37.0+.
...
-
-Scheduling is _always_ in UTC. You'll have to shift timestamps in other zones to
-UTC before scheduling:
-
-```elixir
-%{id: 1}
-|> MyApp.Business.new(scheduled_at: DateTime.shift_zone!(datetime, "Etc/UTC"))
-|> Oban.insert()
-```
+  1. Define a worker to process jobs in the `mailers` queue (see `Oban.Worker`):
```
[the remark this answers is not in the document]
whatyouhide: Sure, no probs. Markdown does that automagially but I can also use real numbers 🙃 Done.

guides/configuration.md
```diff
@@ -0,0 +1,50 @@
...
+
+  * Only jobs in the configured queues will execute. Jobs in any other queue will
+  stay in the database untouched.
+
+  * Pay attention to the number of concurrent jobs making expensive system calls (such as calls to
+  resource-intensive tools like [FFMpeg][ffmpeg] or [ImageMagick][imagemagick]). The BEAM ensures
+  that the system stays responsive under load, but those guarantees don't apply when using ports
+  or shelling out commands.
+
+[ffmpeg]: https://www.ffmpeg.org
```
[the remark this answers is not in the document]
whatyouhide: Woah, missed this. Good catch.

guides/job_uniqueness.md
```diff
@@ -0,0 +1,131 @@
...
+use Oban.Worker, unique: true
+```
+
+Here's a more complex example which uses multiple criteria:
+
+```elixir
+use Oban.Worker,
+  unique: [
+    # Jobs should be unique for 2 minutes...
+    period: 120,
```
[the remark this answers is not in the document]
whatyouhide: Oooph, nice. Didn't know this was a thing.

lib/oban/worker.ex
whatyouhide: Done, moved to its own guide.

lib/oban/worker.ex
whatyouhide: Yes, we can split this out. The nice thing about splitting it out is that we'll have a home for this stuff (and a URL for it too).

whatyouhide:
@sorentwo second pass done ✅

mix.exs
```diff
@@ -65,15 +65,21 @@ defmodule Oban.MixProject do
...
+      "guides/scheduling_jobs.md",
+      "guides/job_uniqueness.md",
       "guides/isolation.md",
       "guides/troubleshooting.md",
       "guides/release_configuration.md",
+      "guides/operational_maintenance.md",
       "guides/writing_plugins.md",
       "guides/periodic_jobs.md",
+      "guides/clustering.md",
+      "guides/sqlite3.md",
```
sorenone, line 82: We smacked-down the sqlite3 guide earlier today with the mysql addition. 

```suggestion
      "guides/clustering.md"
```

PR #1184, opened 2024-11-18
fix: allow overriding oban functions with optional params

btkostner:
When defining a function with optional parameters, it really becomes two different function heads. Because of this, if you want to make those functions overridable, both arities need to be overridable. It's easier to explain in code:

```elixir
defmodule MyOban do
  use Oban

  def insert(changeset, opts \\ []) do
    # Some code here
  end
end
```

This would result in a compile error like so:

```
warning: this clause for insert/1 cannot match because a previous clause at line 2 always matches
   │
 4 │         def insert(_changeset, _opts \\ []) do
```

If you wanted to get over this, you could exclude the optional argument, but then you'd have to specify it at every use like so:

```elixir
defmodule MyOban do
  use Oban

  def insert(changeset, opts) do
    # Some code here
  end
end

MyOban.insert(changeset, [])
```

I created two tests to reproduce this issue though I'm not a huge fan of defining modules in the test function like so. It should show warning messages if you remove the fix in `oban.ex`.


sorentwo:
Thanks for the PR! Merged the first commit locally and added the base of a `use` test in the compiled exercises module.

PR #1194, opened 2024-11-26
Correct elapsed type for queue shutdown events

wcpines:


lib/oban/telemetry.ex
```diff
@@ -312,7 +312,8 @@ defmodule Oban.Telemetry do
 
   #### Queue Shutdown Events
 
-  * `ellapsed` — the amount of time the queue waited for shutdown, in milliseconds
+  * `elapsed` — the amount of time the queue waited for shutdown, in milliseconds
+  * `ellapsed` — alias for elapsed
```
wcpines, line 316: What is the best way to preserve backwards compatibility?

sorentwo: There's no need to maintain compatibility for the log message. We can just fix the spelling.

sorentwo, line 316:
```suggestion
  * `elapsed` — the amount of time the queue waited for shutdown, in milliseconds
```

lib/oban/telemetry.ex
```diff
@@ -458,7 +459,8 @@ defmodule Oban.Telemetry do
   def handle_event([:oban, :queue, :shutdown], measure, %{orphaned: [_ | _]} = meta, opts) do
     log(opts, fn ->
       %{
-        ellapsed: measure.ellapsed,
+        elapsed: measure.elapsed,
+        ellapsed: measure.elapsed,
```
sorentwo, line 463:
```suggestion
        elapsed: measure.elapsed,
```

sorentwo (approved):

sorentwo:
Thank you!

PR #1208, opened 2024-12-26
Igniter task for installing Oban

sorentwo:
This adds the skeleton of a task with a few basic options.

- [x] Add the oban dependency
- [x] Configure the otp app
- [x] Configure testing
- [x] Add Oban to the supervision tree
- [x] Import oban as into formatter deps
- [x] Generate ecto migration
- [x] Automatically change the engine based on the repo adapter
- [x] Document using `mix igniter.install oban` as a single step install
- [x] Test the task (https://hexdocs.pm/igniter/Igniter.Test.html)


guides/introduction/installation.md
```diff
@@ -1,5 +1,23 @@
 # Installation
 
+Oban may be installed and configured via [igniter](https://hex.pm/packages/igniter) with a single command:
+
+```bash
+mix igniter.install oban
+```
```
sorentwo, line 7: This may need a tip on how to install `igniter` itself.

test/mix/tasks/oban.install_test.exs
```diff
@@ -0,0 +1,66 @@
...
+      |> assert_has_patch("lib/oban/application.ex", """
+        | {Oban, Application.fetch_env!(:oban, Oban)},
+        | MyWeb.Endpoint
+      """)
+    end
+
+    test "installing selects correct config with alternate repo" do
+      [app_name: :oban]
+      |> test_project()
+      |> Map.replace!(:args, %{options: [repo: "Oban.Test.LiteRepo"]})
```
sorentwo, line 55: This is definitely invasive, but I didn't see any official API for setting options.

lib/mix/tasks/oban.install.ex
```diff
@@ -0,0 +1,166 @@
...
+          """
+
+          {:error, Igniter.add_issue(igniter, issue)}
+      end
+    end
+
+    defp extract_repo(igniter, _app_name, module) do
+      repo = Igniter.Project.Module.parse(module)
+
+      if Code.ensure_loaded?(repo) do
```
zachdaniel, line 122: One of the things that can make installers much more composable is not using the compiled information about a module. For example, in `ash_postgres` we generate a repo for you if it doesn’t exist. So if someone says:

`mix igniter.install ash_postgres,oban`, this would fail because the module isn’t compiled.

to address that, you can use `Igniter.Project.Module.find_module` to look up the module. This gives you back a zipper. Then you can go to the `use Ecto.Repo` and look for the `adapter` option and use `Igniter.Code.Common.expand_literal` to get a proper module name out.


Along these lines, something I’ve considered doing is adding something like `Igniter.ensure_all_compiled` which will compile (but *not save*) everything that’s been done so far. It’s a big lift though so probably wouldn’t happen soon. If we did that then you could just drop it at the beginning of this task and then introspect modules to your hearts desire.

Anyway, the suggestion is to use the source code and not the compiled information to allow for composability with tools that generate the repo. Not strictly necessary, and could also be done later. I will probably PR some additional checks that make this installer work against `AshPostgres` and `AshSqlite` later if you use the repo source code, as those modify the `use` statement :)

sorentwo: Good note. Ideally this would compose with modules that aren't compiled yet.

The initial approach was to use `Ecto.list_repos/1` to find the defaults and `Module.module_exists/2` to verify that the module was available. Each failed to find the expected modules, which could be down to the fact they're in `test/support/repo` and there are multiple modules in the same file. I'll give the `find_module` and parse approach a try.

zachdaniel: I'm building a big installer for new Ash projects, and would love to have a single checkbox for "Background Jobs" that installs both Oban and the Ash integration for it in one go❤️.

zachdaniel: I'll probably reach out to workshop the copy I use for the tooltip for that option, and how it links to Oban etc.

sorentwo: Done!

zachdaniel (approved):

zachdaniel:
Looks great!

PR #1213, opened 2025-01-02
Proposal: Add Oban.json_library/0

thiamsantos:
Elixir 1.18 introduced [official JSON support](https://hexdocs.pm/elixir/changelog.html#json-support). 

This PR adds support to configure the json library used by Oban, in order to allow the use of the official JSON module. 

By default Oban will keep using Jason.

The code is [adapted from Phoenix](https://github.com/phoenixframework/phoenix/blob/v1.7.18/lib/phoenix.ex#L43).

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -1441,6 +1441,19 @@ defmodule Oban do
     {:ok, length(deleted_jobs)}
   end
 
+  @doc """
+  Returns the configured JSON encoding library for Oban.
+
+  To customize the JSON library, including the following
+  in your `config/config.exs`:
+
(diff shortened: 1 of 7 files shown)


sorentwo:
@thiamsantos Thanks for the PR and nudging us towards using the new JSON module. When Phoenix added the `json` library config it was to help migrate from the `Poison` to `Jason` library. Now, we're migrating to an official module that's included in the language itself and won't need to support alternatives once the minimum supported Elixir version is v1.18.

Our alternative strategy was to add a wrapper that delegates to the best option at compile time.

thiamsantos:
@sorentwo thank you :heart:

PR #1216, opened 2025-01-06
Update scaling.md

javiercr:
Fix `Oban.Plugins.Reindexer` links.

The docs were incorrectly linking to https://github.com/sorentwo/oban_notifiers_phoenix


sorentwo:
Oops, thanks for the fix!

PR #1218, opened 2025-01-06
improvement: support ash in `mix igniter.install oban`

zachdaniel:
chore: small changes to use more convenient igniter functions


sorentwo:
Merged locally after some modifications (acf9cce). The reverse `with` and nested `else` didn't sit well with me 🙂

zachdaniel:
I like your version better too :)

PR #1223, opened 2025-01-10
Use query_type: :text for cancel_all_jobs in Dolphin engine

beerlington:
Using literal SQL allows us to use `query_type: text`, which is required in environments such as Planetscale where the binary protocol is incompatible with prepared statements.

This is an attempt at fixing https://github.com/oban-bg/oban/issues/1222.


sorentwo:
Fixed with c36c7fe

PR #1229, opened 2025-01-23
Remove half-sentence from changelog

PragTob:
Seems like a half-sentence was left there trying to describe what happened previously. We should either fully formulate it or delete it - since I don't know what it was previously I opted to remove it but yeah depends on you what is better :)

Also, thanks for oban! :green_heart: 

IMG_20220601_080610


sorentwo:
Thanks! Well spotted. The "Previously" bit wasn't part of the original commit message either, so I'm stumped 🤷

PR #1230, opened 2025-01-27
Make oban installer idempotent

zachdaniel:
by using `on_exists: :skip` when generating the migration, it is safe for other igniter installers (like the one I'm working on) to add `{:oban, "~> 2.0"}` to their `installs` list in their igniters. If oban is present, the installer does nothing, but if not the user gets a full oban installation.

Also removes an unnecessary `add_dep` call which prompts users to change the dep in cases where they have something like `~> 2.0` instead of exactly `~> 2.18`.

### DIFF
--- a/lib/mix/tasks/oban.install.ex
+++ b/lib/mix/tasks/oban.install.ex
@@ -90,12 +90,11 @@ if Code.ensure_loaded?(Igniter) do
           """
 
           igniter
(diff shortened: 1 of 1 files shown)


zachdaniel:
If you happen to remember, please let me know here if/when this is released 🙇

PR #1233, opened 2025-01-31
Fix :get_dynamic_repo with releases

davydog187:
Lave Ducia here, long-time user, ~~first-time~~ second-time contributor!

It seems that the recommendation for a [separate Oban pool](https://hexdocs.pm/oban/scaling.html#pooling) does not work with OTP releases, as anonymous functions cannot be used in configuration.

This PR allows an MFA tuple to be passed instead, and updates the documentation to recommend this instead.

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -50,7 +50,7 @@ defmodule Oban do
   @type option ::
           {:dispatch_cooldown, pos_integer()}
           | {:engine, module()}
-          | {:get_dynamic_repo, nil | (-> pid() | atom())}
+          | {:get_dynamic_repo, nil | (-> pid() | atom()) | {module(), atom(), list()}}
           | {:log, false | Logger.level()}
           | {:name, name()}
           | {:node, oban_node()}
--- a/lib/oban/config.ex
+++ b/lib/oban/config.ex
@@ -13,7 +13,7 @@ defmodule Oban.Config do
   @type t :: %__MODULE__{
           dispatch_cooldown: pos_integer(),
           engine: module(),
-          get_dynamic_repo: nil | (-> pid() | atom()),
+          get_dynamic_repo: nil | (-> pid() | atom()) | {module(), atom(), list()},
           insert_trigger: boolean(),
           log: false | Logger.level(),
           name: Oban.name(),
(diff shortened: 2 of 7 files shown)


lib/oban/config.ex
```diff
@@ -140,7 +140,8 @@ defmodule Oban.Config do
     Validation.validate_schema(opts,
       dispatch_cooldown: :pos_integer,
       engine: {:behaviour, Oban.Engine},
-      get_dynamic_repo: {:or, [:falsy, {:function, 0}]},
+      get_dynamic_repo:
+        {:or, [:falsy, {:function, 0}, {:custom, &validate_mfa(:get_dynamic_repo, &1)}]},
```
davydog187, line 144: If theres a better way to do this I'm all ears

lib/oban.ex
```diff
@@ -50,7 +50,7 @@ defmodule Oban do
   @type option ::
           {:dispatch_cooldown, pos_integer()}
           | {:engine, module()}
-          | {:get_dynamic_repo, nil | (-> pid() | atom())}
+          | {:get_dynamic_repo, nil | (-> pid() | atom()) | {module(), atom(), list()}}
```
sorentwo, line 53:
```suggestion
          | {:get_dynamic_repo, nil | (-> pid() | atom()) | mfa()}
```

lib/oban/config.ex
```diff
@@ -13,7 +13,7 @@ defmodule Oban.Config do
   @type t :: %__MODULE__{
           dispatch_cooldown: pos_integer(),
           engine: module(),
-          get_dynamic_repo: nil | (-> pid() | atom()),
+          get_dynamic_repo: nil | (-> pid() | atom()) | {module(), atom(), list()},
```
sorentwo, line 16:
```suggestion
          get_dynamic_repo: nil | (-> pid() | atom()) | mfa(),
```

lib/oban/config.ex
sorentwo: How about adding `mfa` to the validator?

lib/oban/config.ex
davydog187: unfortunately, `mfa()` is aliased to `{module(), atom(), arity()}` which isn't what we want here

davydog187:
@sorentwo ready for another look

lib/oban/config.ex
sorentwo: Indeed it is 😞

lib/oban/validation.ex
```diff
@@ -167,6 +167,16 @@ defmodule Oban.Validation do
     end
   end
 
+  defp validate_type(:mfa, key, {module, func, args})
+       when is_atom(module) and is_atom(func) and is_list(args) do
+    if function_exported?(module, func, length(args)) do
+      :ok
+    else
+      {:error,
+       "missing function #{inspect(Exception.format_mfa(module, func, length(args)))} for #{key}"}
```
sorentwo, line 176: Exception.format_mfa/3 returns a string, so using `inspect/1` escapes the quotes.

```suggestion
      {:error, "missing function #{Exception.format_mfa(module, func, length(args))} for #{key}"}
```

sorentwo:
Thank you Lave!

PR #1251, opened 2025-02-20
Add callback for new/1 to Worker

nallwhy:
If you try to override new in `Worker`, it is not possible because `new/1` is not included in `defoverridable` due to the absence of a corresponding callback.

To resolve this issue, we add a `new/1` callback to `Worker`.

```elixir
defmodule MyWorker do
  use Oban.Woker, ...

  @impl Oban.Worker
  def new(args, opts \\ [])  when is_map(args) and is_list(opts) do
    ...
  end
end
```

compile warnings

```
warning: this clause for new/1 cannot match because a previous clause at line 3 always matches
    │
  3 │   Defines a behavior and macro to guide the creation of worker modules.
    │   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    │
    └─ deps/oban/lib/oban/worker.ex:3
```

### DIFF
--- a/lib/oban/worker.ex
+++ b/lib/oban/worker.ex
@@ -363,6 +363,7 @@ defmodule Oban.Worker do
 
   See `Oban.Job.new/2` for the available options.
   """
+  @callback new(args :: Job.args()) :: Job.changeset()
   @callback new(args :: Job.args(), opts :: [Job.option()]) :: Job.changeset()
 
   @doc """


nallwhy:
@sorentwo To remove the warning, I must do it this way.

```elixir
defmodule MyWorker do
  use Oban.Woker, ...

  # @impl Oban.Worker <-- new/1 is not a callback of Oban.Worker
  def new(args) do
    new(args, [])
  end

  @impl Oban.Worker
  def new(args, opts)  when is_map(args) and is_list(opts) do
    ...
  end
end
```

sorentwo:
> It works as you said, but I can't remove the warning because new/1 is not overridable.

You shouldn't be writing `new/1`, just `new/2`. The default implementation of `new/1` will still work anyhow:

```elixir
defmodule MyApp.Worker do
  use Oban.Worker

  @impl Oban.Worker
  def new(args, opts) do
    IO.inspect({args, opts})

    super(args, opts)
  end

  @impl Oban.Worker
  def perform(_job), do: :ok
end

MyApp.Worker.new(%{foo: :bar})
# => {%{foo: :bar}, []}
```

nallwhy:
@sorentwo Ah, I see. I misunderstood. Thanks for letting me know!

Issue #1253, opened 2025-02-24
Proposal: Anti-patterns guide


gugahoa:
Hi all, wanted to propose that we write a guide on common anti patterns when using Oban.

Here's a few I have already seen on my team:
- Using recorded jobs to hold domain significant data, to be used by other jobs (e.g. using oban jobs to fetch a given entity external id every time it's needed instead of saving on a proper table)
- Using Workflows that each job depends on the previous one instead of using Chained jobs

gugahoa:
there's a third one I still dont know how to word it: disconnected jobs that operate on a specific entity on a external service.

For example, one job creates a draft order on Shopify, another mark it as completed. The first depends on the first, thus it should be a Chained job.

gugahoa:
* Using Oban jobs in place of Tasks – e.g. max_attempts: 1 is a code smell

gugahoa:
Asked around the Elixir's Brazil Telegram community, here's a few anti-patterns/smells that were shared:
- Wanting to save Oban jobs forever, to hold state
- Using the `default` queue blindly

gugahoa:
Here's how I thought we could structure the guide:
```markdown
# Oban Anti-Patterns Guide
## Introduction
## Anti-Pattern 1: Example anti pattern title
### Problem
### Example
### Refactoring
### Additional Remarks
```

So far the list I gathered contains:
1. Using Recorded Jobs to Hold Domain-Significant Data
2. Using Workflows Where Chained Jobs Would Suffice
3. Disconnected Jobs for Connected Operations
4. Using Oban Jobs Instead of Tasks
5. Saving Oban Jobs Forever for State
6. Using the Default Queue Blindly

sorentwo:
Thanks for starting this discussion @gugahoa. Some of the items you've gathered already are definite anti-patterns, but a few of them are in the "it depends" category.

> Using Workflows Where Chained Jobs Would Suffice

This applies specifically to "infinite" workflows used for flow control, where jobs are perpetually appended. A linear workflow that has a defined end is perfectly appropriate.

> Using Oban Jobs Instead of Tasks

A job with `max_attempts: 1` is perfectly fine. The [deciding factor](https://oban.pro/articles/oban-starts-where-tasks-end) should be whether you _need_ that task to succeed, or want observability into it, or have it run on a separate node, or schedule it, etc.

The structure looks good as well. It follows @sorenone's scaling guide and has a good flow.

sorentwo:
Closing in favor of #1257

PR #1257, opened 2025-02-25
Anti-patterns Guide

gugahoa:
Related to: #1253 

Full disclaimer: I used AI to write the first draft. For the example I either pulled directly from my work's codebase and anonymized it, got from Oban's doc or let the AI cook up an example.

I added one more section than from the issue, as it's related to another thing I noticed from my team: not differentiating between discard and cancel returns.

### DIFF
--- a/guides/recipes/anti-patterns.md
+++ b/guides/recipes/anti-patterns.md
@@ -1,0 +1,843 @@
+# Oban Anti-Patterns Guide
+
+## Introduction
+
+[Oban](https://github.com/oban-bg/oban) is a robust job processing system for Elixir, providing applications with reliability, persistence, and concurrency for background job processing. Despite its excellent design, even experienced developers can fall into patterns that undermine Oban's effectiveness or create maintenance challenges.
+
+This guide identifies common anti-patterns when working with Oban and provides concrete refactoring strategies. By recognizing and avoiding these anti-patterns, you'll create more maintainable, performant, and reliable background job systems that better leverage Oban's architecture.
+
+## Anti-Pattern 1: Using Recorded Jobs as Domain Data Storage
+
+### Problem
+
+Oban Pro's `recorded: true` option preserves job results for workflow coordination. However, some developers misuse this feature as a primary storage mechanism for domain data that should reside in dedicated database tables.
+
+**Why this is problematic:**
+
+- **Data lifecycle mismatch**: Jobs are designed to be ephemeral while domain data often requires long-term persistence
+- **Data loss risk**: Recorded data is lost when jobs are pruned (which is a necessary maintenance operation)
+- **Query limitations**: Retrieving data from job records is inefficient compared to properly designed schemas
+- **Architectural coupling**: Creates an inappropriate dependency between background processing and core domain data
+
+### Example
+
+```elixir
+defmodule MyApp.Workers.ExternalEntityImporter do
+  use Oban.Pro.Workers.Workflow, recorded: true, queue: :imports
+
+  @impl true
+  def process(%Job{args: %{"external_id" => external_id}}) do
+    # Fetch and transform external data
+    {:ok, response} = MyApp.ExternalAPI.fetch_entity(external_id)
+
+    # Store essential domain data in the job record instead of a proper table
+    {:ok, %{id: response.id, name: response.name, details: response.details}}
+  end
+end
+
+defmodule MyApp.Workers.ExternalEntityProcessor do
+  use Oban.Pro.Workers.Workflow, queue: :processing
+
+  @impl true
+  def process(%Job{args: args} = job) do
+    # Retrieve domain data from the recorded job rather than from a proper DB table
+    with {:ok, entity} <- fetch_entity_data(job) do
+      # Process the entity using data from the recorded job
+      MyApp.Processor.process_entity(entity)
+    end
+  end
+
+  defp fetch_entity_data(job) do
+    job
+    |> Oban.Pro.Workers.Workflow.all_jobs(names: ["import_entity"])
+    |> List.first()
+    |> fetch_recorded()
+  end
+end
+```
+
+### Refactoring
+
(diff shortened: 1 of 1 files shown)


guides/recipes/anti-patterns.md
```diff
@@ -0,0 +1,843 @@
...
+      |> where([j], j.worker == "MyApp.Workers.InvoiceProcessor")
+      |> where([j], j.state == "completed")
+      |> where([j], fragment("?->>'year' = ?", j.args, ^to_string(year)))
+      |> where([j], fragment("?->>'month' = ?", j.args, ^to_string(month)))
+      |> MyApp.Repo.all()
+
+    # Process jobs for reporting
+    process_jobs_for_report(jobs)
+  end
+end
```
gugahoa, line 480: This example seems bad

guides/recipes/anti-patterns.md
```diff
@@ -0,0 +1,843 @@
...
+  end
+
+  def get_by_period(job_type, year, month) do
+    JobResult
+    |> where([r], r.job_type == ^job_type)
+    |> where([r], r.year == ^year)
+    |> where([r], r.month == ^month)
+    |> Repo.all()
+  end
+end
```
gugahoa, line 549: Since the example is weak, the refactor also seems weak

guides/recipes/anti-patterns.md
```diff
@@ -0,0 +1,843 @@
...
+|----------------|---------------|
+| **Duration** | Separate quick jobs from long-running ones |
+| **Resource usage** | Limit concurrency for resource-intensive operations |
+| **Priority** | Use both queues and priority values for fine-grained control |
+| **External dependencies** | Create dedicated queues for rate-limited external services |
+| **Business importance** | Ensure critical operations get processed first |
+
+The Smart Engine will dynamically adjust concurrency based on job execution patterns, but it works best when you start with thoughtfully designed queue structure.
+
+## Anti-Pattern 7: Insufficient Error Handling in Jobs
```
gugahoa, line 694: This is a new one, not in the issue. It arose from a discussion within my team where it was not clear the distinction on when to use `{:discard, any()}` and when to use `{:cancel, any()}`

My take on this is that `:discard` should be used either when attempts are exhausted or we faced a unrecoverable error. And `:cancel` should be used when although not successful, it's not an error – kinda like a `no-op` return

guides/recipes/anti-patterns.md
```diff
@@ -0,0 +1,843 @@
...
+          {:error, %{status: 400, reason: reason}} ->
+            # Client error - don't retry
+            Logger.warn("User sync failed with client error: #{reason}", args: args)
+            {:discard, :client_error}
+
+          {:error, %{status: status}} when status >= 500 ->
+            # Server error - worth retrying with backoff
+            backoff = :math.pow(2, attempt) |> round()
+            Logger.warn("User sync failed with server error: #{status}", args: args)
+            {:snooze, backoff}
```
gugahoa, line 752: I think this is a bad example on `:snooze`. I use it when some resource the job needs is not ready yet, but it should be.

gugahoa: Although that may lead to another anti pattern? Infinitely snoozing jobs

guides/recipes/anti-patterns.md
```diff
@@ -0,0 +1,843 @@
...
+          {:error, :timeout} ->
+            # Network issue - worth retrying
+            Logger.warn("User sync timed out", args: args)
+            {:error, :timeout}
+        end
+
+      {:error, :not_found} ->
+        # Don't retry when the user doesn't exist
+        Logger.info("Cannot sync user #{user_id} - not found", args: args)
+        {:discard, :user_not_found}
```
gugahoa, line 763: Maybe a good example of when to use `cancel`? As the job became a no-op

guides/recipes/anti-patterns.md
```diff
@@ -0,0 +1,843 @@
...
+
+Additional recommendations:
+
+- ✅ Log context-rich information for all errors
+- ✅ Use exponential backoff for retries
+- ✅ Set appropriate `max_attempts` values based on error types
+- ✅ Consider uniqueness constraints for idempotency
+- ✅ Use structured error returns rather than exceptions
+
+## Best Practices for Oban Usage
```
gugahoa, line 792: This whole section was the AI being too eager, I left it here in case you find it valuable. Though it should be in another guide and not in the anti-patterns one

guides/recipes/anti-patterns.md
sorentwo: Noted on Slack:

> Regarding the `cancel` vs `discard` discussion, the discard tuple is deprecated and undocumented now. Discard should always be because of exhausted retries, and cancel is always a manual action.

guides/recipes/anti-patterns.md
sorentwo: It's better to use the backoff provided by Oban itself, either through the worker (`backoff/1`) or the `Oban.Backoff` module. This would lead to infinitely snoozing jobs, which is bad. Also, I wouldn't include arbitrary logger warnings in example code.

guides/recipes/anti-patterns.md
sorentwo: Definitely use `{:cancel, :user_not_found}` rather than `{:discard, _}`. And don't include logging.

guides/recipes/anti-patterns.md
sorentwo: Using a job result temporarily is fine, which is why there's a `recorded` option in the first place. If the result is important in the long term then it should be written to a proper database entity, and certinly not a `JobResult`.

sorentwo:
@gugahoa Are you able to revisit this? Any other thoughts here?

sorentwo:
Closing this due to lack of activity. We can return to it later if there's renewed interest.

PR #1264, opened 2025-03-10
chore: add docs about args and fields relantionship

yordis:


yordis:
@sorentwo what about this one?

sorentwo:
Tweaked the language and merged manually with 2dc0bf1

PR #1269, opened 2025-03-12
chore: Move preferred_cli_env into def cli

halfdan:
Fixes another Elixir 1.19 deprecation: https://hexdocs.pm/elixir/main/changelog.html

`def cli` was introduced in 1.15 so it's backward compatible down to that version (which also matches the minimum version requirement currently listed in the README -> https://github.com/oban-bg/oban?tab=readme-ov-file#requirements)


sorentwo:
Thanks!

PR #1273, opened 2025-03-19
Correct staleness logic

kivra-emifal:
We wanted to use `Oban.Notifier.status/0` in a healthcheck to see if we had a broken notifier connection. When testing this locally we found that the local node never got stale and was never removed from the list of nodes in Sonar.

After some thinking about it we think the logic is incorrect and something like this is correct. Tested it locally by running the node, closing the database and seeing that it was pruned after the threshold time had past.

### DIFF
--- a/lib/oban/sonar.ex
+++ b/lib/oban/sonar.ex
@@ -115,8 +115,9 @@ defmodule Oban.Sonar do
   end
 
   defp prune_stale_nodes(state) do
-    stale = System.monotonic_time(:millisecond) + state.interval * state.stale_mult
-    nodes = Map.reject(state.nodes, fn {_, recorded} -> recorded > stale end)
+    now = System.monotonic_time(:millisecond)
+    threshold = state.interval * state.stale_mult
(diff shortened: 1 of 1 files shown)


lib/oban/sonar.ex
```diff
@@ -115,8 +115,9 @@ defmodule Oban.Sonar do
   end
 
   defp prune_stale_nodes(state) do
-    stale = System.monotonic_time(:millisecond) + state.interval * state.stale_mult
-    nodes = Map.reject(state.nodes, fn {_, recorded} -> recorded > stale end)
+    now = System.monotonic_time(:millisecond)
+    threshold = state.interval * state.stale_mult
+    nodes = Map.reject(state.nodes, fn {_, recorded} -> now - recored > threshold end)
```
sorentwo, line 120:
```suggestion
    nodes = Map.reject(state.nodes, fn {_, recorded} -> now - recorded > threshold end)
```

sorentwo:
Thanks for the PR. It needed a test change as well, so I've recreated the fix and updated the test in a separate commit.

kivra-emifal:
No problems! Thanks for quick handling of it!

PR #1277, opened 2025-03-25
Disallow :keys when :fields don't contain :args or :meta

Geekfish:
# Context

According to the docs on [Unique Jobs](https://hexdocs.pm/oban/2.19.4/unique_jobs.html):

> `:keys` — A specific subset of the `:args` or `:meta` to consider when comparing against historic jobs. This allows a job with multiple key/value pairs in its arguments to be compared using only a subset of them.

However users are currently allowed to insert jobs with a  `:keys` configuration, even when neither `:args` nor `:meta` are passed, e.g.

```elixir
use Oban.Worker,
  unique: [
    keys: [:my_identifier],
    fields: [:worker]
  ]
  ```
  
  the above setup is allowed without any indication that `:keys` will be ignored.
  
  # Solution
  
  This MR adds validation for the above case when the `Job` changeset is validated.

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -651,7 +651,7 @@ defmodule Oban.Job do
       not (is_list(keys) and Enum.all?(keys, &is_atom/1)) ->
         {:error, "expected :keys to be a list of atoms"}
 
-      is_nil(keys) and Enum.any?(@keyable_fields, &(&1 in Keyword.get(unique, :fields)))) ->
+      not (is_list(keys) and Enum.any?(@keyable_fields, &(&1 in Keyword.get(unique, :fields)))) ->
         {:error,
          "using :keys expects :fields to contain at least one of #{inspect(@keyable_fields)}"}
 
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -201,7 +201,7 @@ defmodule Oban.Job do
 
   @unique_fields ~w(args meta queue worker)a
   @unique_timestamps ~w(inserted_at scheduled_at)a
-  @key_compatible_fields ~w(args meta)a
+  @keyable_fields ~w(args meta)a
 
   @unique_defaults %{
     fields: ~w(args queue worker)a,
@@ -651,10 +651,9 @@ defmodule Oban.Job do
       not (is_list(keys) and Enum.all?(keys, &is_atom/1)) ->
         {:error, "expected :keys to be a list of atoms"}
 
-      not (is_list(keys) and
-               Enum.any?(@key_compatible_fields, &(&1 in Keyword.get(unique, :fields)))) ->
+      is_nil(keys) and Enum.any?(@keyable_fields, &(&1 in Keyword.get(unique, :fields)))) ->
         {:error,
-         "using :keys expects :fields to contain at least one of #{inspect(@key_compatible_fields)}"}
+         "using :keys expects :fields to contain at least one of #{inspect(@keyable_fields)}"}
 
       true ->
         :ok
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/oban/job.ex [hunk 1]
```diff
@@ -643,4 +642,18 @@ defmodule Oban.Job do
         {:error, "unknown option, #{inspect(option)}"}
     end)
   end
+
+  defp validate_keys(keys, unique) do
```
Geekfish, line 646: Moved this to a separate function after CI complaining about cyclomatic complexity.

test/oban/job_test.exs
```diff
@@ -129,6 +129,38 @@ defmodule Oban.JobTest do
       assert Job.new(%{}, worker: Fake, unique: [states: [:random]]).errors[:unique]
       assert Job.new(%{}, worker: Fake, unique: [timestamp: :updated_at]).errors[:unique]
     end
+
+    test "empty keys don't require args or meta in fields" do
+      assert Job.new(%{}, worker: Fake, unique: [keys: [], fields: [:worker]]).valid?
+    end
```
Geekfish, line 135: Hey @sorenone ! Sorry for not checking this fully locally first, CI should now be green and accept this case.

lib/oban/job.ex
```diff
@@ -201,6 +201,7 @@ defmodule Oban.Job do
 
   @unique_fields ~w(args meta queue worker)a
   @unique_timestamps ~w(inserted_at scheduled_at)a
+  @key_compatible_fields ~w(args meta)a
```
sorenone, line 204: Shortening the attribute name for conciseness below. Also, so that @sorentwo gives me a smooch.

```suggestion
  @keyable_fields ~w(args meta)a
```

lib/oban/job.ex
```diff
@@ -643,4 +642,22 @@ defmodule Oban.Job do
...
+  defp validate_keys(keys, unique) do
+    cond do
+      keys == [] ->
+        :ok
+
+      not (is_list(keys) and Enum.all?(keys, &is_atom/1)) ->
+        {:error, "expected :keys to be a list of atoms"}
+
+      not (is_list(keys) and
+               Enum.any?(@key_compatible_fields, &(&1 in Keyword.get(unique, :fields)))) ->
```
sorenone, line 655: Changed for readability and conciseness. 

```suggestion
      is_nil(keys) and Enum.any?(@keyable_fields, &(&1 in Keyword.get(unique, :fields)))) ->
```

lib/oban/job.ex
```diff
@@ -643,4 +642,22 @@ defmodule Oban.Job do
...
+        {:error,
+         "using :keys expects :fields to contain at least one of #{inspect(@key_compatible_fields)}"}
```
sorenone, line 657:
```suggestion
         "using :keys expects :fields to contain at least one of #{inspect(@keyable_fields)}"}
```

test/oban/job_test.exs
sorenone: @Geekfish No issue! Thank you!

sorenone (approved):

lib/oban/job.ex
```diff
@@ -643,4 +642,21 @@ defmodule Oban.Job do
...
+
+  defp validate_keys(keys, unique) do
+    cond do
+      keys == [] ->
+        :ok
+
+      not (is_list(keys) and Enum.all?(keys, &is_atom/1)) ->
+        {:error, "expected :keys to be a list of atoms"}
+
+      is_nil(keys) and Enum.any?(@keyable_fields, &(&1 in Keyword.get(unique, :fields)))) ->
```
sorenone, line 654:
```suggestion
      not (is_list(keys) and Enum.any?(@keyable_fields, &(&1 in Keyword.get(unique, :fields)))) ->
```

sorenone:
@Geekfish Thank you!

PR #1279, opened 2025-03-27
Add index state discarded_at / state cancelled_at

Graborg:
Hi! After upgrading to the latest release we noticed that cancelled / discarded jobs are [now being properly pruned](https://github.com/oban-bg/oban/commit/861dfa55e4e3e3cf65a3882baccca64c7fbcbc3d#diff-dae8a75e38cc8e3b8a29cf54657560da701f191a872e9d47c6c4fbf4c036cabcR157-R158) 🙌 , but we noticed that the pruner query got a lot slower. We tracked this to be due to two nonexisting indexes. Let me know if anything can be improved!

### DIFF
--- a/lib/oban/migrations/postgres/v13.ex
+++ b/lib/oban/migrations/postgres/v13.ex
@@ -1,0 +1,10 @@
+defmodule Oban.Migrations.Postgres.V13 do
+  @moduledoc false
+
+  use Ecto.Migration
+
+  def change do
+    create index(:oban_jobs, [:state, :cancelled_at])
+    create index(:oban_jobs, [:state, :discarded_at])
+  end
+end


sorentwo:
> After upgrading to the latest release we noticed that cancelled / discarded jobs are [now being properly pruned](https://github.com/oban-bg/oban/commit/861dfa55e4e3e3cf65a3882baccca64c7fbcbc3d#diff-dae8a75e38cc8e3b8a29cf54657560da701f191a872e9d47c6c4fbf4c036cabcR157-R158) 🙌 , but we noticed that the pruner query got a lot slower.

That's precisely why the previous query only used `scheduled_at`, because it's part of the compound index 🙂 

> We tracked this to be due to two nonexisting indexes. Let me know if anything can be improved!

We're hesitant to add new migrations, and especially single column/single use indexes like these. However, they may be important for performance on some systems. What do you think about adding some documentation on adding these indexes as for performance?

kivra-mikgra:
Thanks for the quick reply! Yeah, that makes sense.
Of course, I'll add some documentation!

Graborg:
Thanks for the quick reply! Yeah, that makes sense.
Of course, I'll add some documentation!

Graborg:
I added a migration guide, although I'm uncertain what the specific version will be, but I guess that can be changed once this is merged and a release version is determined. Is this similar what you had in mind?

guides/upgrading/v2.20.md
```diff
@@ -0,0 +1,46 @@
...
+> removing jobs that have been cancelled or discarded. 
+
+## Bump Your Deps
+
+Update Oban (and optionally Pro) to the latest versions:
+
+```elixir
+[
+  {:oban, "~> 2.20"},
+  {:oban_pro, "~> 1.2", repo: "oban"}
```
sorentwo, line 17: This is rather outdated, and incompatible with current versions of Oban.

```suggestion
```

guides/upgrading/v2.20.md
```diff
@@ -0,0 +1,46 @@
...
+]
+```
+
+## Run Oban.Migrations for v13 (Optional)
+
+The v13 migration adds indexes for [state, cancelled_at] and [state, discarded_at] columns. This is done
+to improve the performance of the Oban.Plugins.Pruner, which prunes completed, cancelled and discarded jobs.
+See (https://github.com/oban-bg/oban/commit/861dfa55e4e3e3cf65a3882baccca64c7fbcbc3d#diff-dae8a75e38cc8e3b8a29cf54657560da701f191a872e9d47c6c4fbf4c036cabcR157-R158). This will significantly improve the performance of the pruner when pruning large oban_jobs tables, with little to no impact on smaller tables.
+
```
sorentwo, line 26: No need to be quite so specific, or link to the commit:

```suggestion
The v13 migration adds compound indexes for `cancelled_at` and `discarded_at` columns. This is
done to improve `Oban.Plugins.Pruner` performance for cancelled and discarded jobs.
```

guides/upgrading/v2.20.md
```diff
@@ -0,0 +1,46 @@
+# Upgrading to v2.20
+
+This Oban release includes an optional, but recommended migration.
+
+> #### Improve query time for pruning cancelled & discarded jobs
+>
+> The v13 migration needs to be run for the pruner to use indexed queries when 
+> removing jobs that have been cancelled or discarded. 
+
```
sorentwo, line 9: This is covered below, and I don't believe it warrants an admonition block.

```suggestion
```

sorentwo:
There are some flickering tests that are unrelated to the changes in this PR. Nothing that blocks merging this 🙂

kivra-mikgra:
I see, thanks!

PR #1286, opened 2025-04-13
Fix Oban Pro link in Table of Contents

henriqueberlesi:
Looks like the :star2: added by commit 62b87db23221adc713fb3ec8c5732166e5bd1e28 broke the link in Table of Contents to Oban Pro section (tested in Firefox and GitHub app).


sorentwo:
Thanks for the fix!

PR #1293, opened 2025-04-28
Cron: support timezone option for individual cron jobs

azizk:
Hi!

At my job we would appreciate the ability to override/define the timezone for individual cron jobs. :sunglasses:


sorentwo:
Thanks for taking the time to make a PR, but we won't be adding this feature. It is intentionally missing because it's part of [DynamicCron](https://oban.pro/docs/pro/1.6.0-rc.3/Oban.Pro.Plugins.DynamicCron.html#module-overriding-the-timezone) in Pro.

azizk:
Ah yes, I knew about `DynamicCron` but missed the part about timezone overrides. Thanks!

PR #1304, opened 2025-05-22
Add crash_reason to logs metadata and move log serialization to report_cb/1 callback

martosaur:
## Problem

Currently, almost [every](https://github.com/getsentry/sentry-elixir/blob/cd718888f6caddc6d8d05334d46e954b7dc45260/lib/sentry/integrations/oban/error_reporter.ex#L8-L18) [logging](https://github.com/appsignal/appsignal-elixir/blob/4e4b0c99f53e8cdba4f214c5313bff74652b231c/lib/appsignal/oban.ex#L29) [library](https://github.com/elixir-error-tracker/error-tracker/blob/7c0259882f8155f4cb67000702c78ebfd6723655/lib/error_tracker/integrations/oban.ex#L54-L58) tracks Oban errors by reporting them off of telemetry. Which leads to the situation when Oban errors are either tracked, but not logged (if Oban default logger is disabled or set to lower level) or potentially tracked twice (once by the logging library off of a telemetry event and second time by the logger handler off of a log event).

I can see 3 reasons why logging libraries do it this way:
1. Telemetry event is structured, while Oban default logger serializes its report to JSON by default
2. Telemetry event clearly indicates that an event is an unhnandled error. It's harder to infer this for log events, where handlers usually look for `crash_reason` key in metadata, which Oban default logger doesn't provide.
3. Telemetry event error is structured: it contains `kind`, `reason` and `stacktrace` - the holy trinity of error reporting that libraries know how to handle. Log event only contains serialized error under `error` key.

## Proposed Solution

This PR aims to improve on all three points with a very modest set of changes:

### 1. Make Oban log events always a report by moving optional JSON serialization to `report_cb/1` callback

Oban can make structured log reports always accessible to logger handlers by moving optional JSON serialization to [`report_cb/1`](https://www.erlang.org/doc/apps/kernel/logger.html#t:report_cb/0) callback. This also opens the door to nicer human-readable Oban log message if the team wants to implement them. See how `Task` module [formats own reports for example](https://github.com/elixir-lang/elixir/blob/7b20c281d521aa7aa2ad2baa1e9ae6c579d79d0c/lib/elixir/lib/task/supervised.ex#L145).

As a side note, serializing own log message to JSON is fairly unusual, in my experience. Normally it is the job for Logger Formatters (like [`LoggerJSON`](https://hex.pm/packages/logger_json)) and isn't trivial as they need to ensure that they can serialize arbitrary terms and often scrub sensitive data. I'm not sure why Oban does this, but I would love to know if possible! Depending on the use case for this, this change might not be a viable solution.

### 2-3. Put `crash_reason` to log event metadata

This one is fairly straightforward, whenever we have an error on our hands, put it under `crash_reason` metadata case as per Elixir convention. This will allow logger handlers recognize log event as an unhandled exception and treat it the way they prefer.

## Changes for users

* Libraries that prefer to track Oban errors off of telemetry events will be able to continue to do so
* Libraries that prefer to incorporate Oban logic into their logger handler and effectively give them the same treatment as e.g. GenServer, will be able to do so
* Oban users with default handler and console formatter won't see any change
* Oban users that have handlers and/or formatters relying on pre-serialized log event will likely find these changes breaking.

### DIFF
--- a/lib/oban/telemetry.ex
+++ b/lib/oban/telemetry.ex
@@ -468,30 +468,35 @@ defmodule Oban.Telemetry do
     log(opts, fn ->
       details = Map.take(meta.job, ~w(attempt args id max_attempts meta queue tags worker)a)
 
-      extra =
+      {extra, metadata} =
         case event do
           :start ->
-            %{event: "job:start", system_time: measure.system_time}
+            {%{event: "job:start", system_time: measure.system_time}, []}
 
           :stop ->
-            %{
+            extra = %{
               duration: convert(measure.duration),
               event: "job:stop",
               queue_time: convert(measure.queue_time),
               state: meta.state
             }
 
+            {extra, []}
+
           :exception ->
-            %{
+            extra = %{
               error: Exception.format_banner(meta.kind, meta.reason, meta.stacktrace),
               event: "job:exception",
               duration: convert(measure.duration),
               queue_time: convert(measure.queue_time),
               state: meta.state
             }
+
+            {extra, crash_reason: to_crash_reason(meta.kind, meta.reason, meta.stacktrace)}
         end
 
-      Map.merge(details, extra)
+      report = Map.merge(details, extra)
+      {report, metadata}
     end)
   end
 
@@ -550,21 +555,24 @@ defmodule Oban.Telemetry do
 
   def handle_event([:oban, :plugin, :exception], measure, meta, opts) do
     log(opts, fn ->
-      error =
+      {error, crash_reason} =
         case meta do
           %{kind: kind, reason: reason, stacktrace: stacktrace} ->
-            Exception.format_banner(kind, reason, stacktrace)
+            {Exception.format_banner(kind, reason, stacktrace),
+             to_crash_reason(kind, reason, stacktrace)}
 
           %{error: error} ->
-            Exception.format_banner(:error, error)
+            {Exception.format_banner(:error, error), {error, []}}
         end
 
-      %{
+      message = %{
         duration: convert(measure.duration),
         error: error,
         event: "plugin:exception",
         plugin: inspect(meta.plugin)
       }
+
+      {message, crash_reason: crash_reason}
     end)
   end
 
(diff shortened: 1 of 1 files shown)


martosaur:
Ah this makes sense. Alright, I removed the `report_cb` change from the PR.

How do you feel about also adding `:oban` to the log event domain? Oban does insert `source: "oban"` into the report, but it's hard to access if the report is json. I'm mostly worried about the scenario that once `crash_reason` change is rolled out, logging libraries might end up logging Oban errors both from log events and telemetry handlers and they would either want to deprecate their integration or ignore Oban events. The latter is a quicker fix, but they'll need a way to easily recognize Oban logs.

sorentwo:
> How do you feel about also adding :oban to the log event domain? Oban does insert source: "oban" into the report, but it's hard to access if the report is json.

It would be an excellent change for all of the log output where `source: "oban"` is used.

>  I'm mostly worried about the scenario that once crash_reason change is rolled out, logging libraries might end up logging Oban errors both from log events and telemetry handlers and they would either want to deprecate their integration or ignore Oban events.

That's an excellent point. I believe most people, and libraries (including Sentry and AppSignal) use telemetry specifically to handle error reporting for Oban. We can't ship something that would potentially cause duplicate reports for people. These changes may have to hold for Oban 3.0 as well.

martosaur:
Fair enough. I'll leave this PR at your mercy then. 

One last thing I can think of in defense of this change is that Elixir itself will change how [log events look like](https://github.com/elixir-lang/elixir/pull/14380) in Elixir 1.19 in a lot of cases and that wasn't considered a breaking change. There is also a `plug_cowboy` where `crash_reason` was added in a [patch release](https://github.com/elixir-plug/plug_cowboy/blob/master/CHANGELOG.md#v202-2019-03-18). But it's totally understandable why Oban might not want to do this!

sorentwo:
A typical application using Oban sees _many_ errors and exceptions for job failures, especially compared to plug/cowboy connection errors. In addition, companies frequently monitor Oban error rates for alerts and dashboards, and we'd rather not risk impacting those systems with a non-major release.

These changes are now listed in the [3.0 changes checklist](https://github.com/oban-bg/oban/issues/849) and won't be forgotten. Thank you!

grzuy:
> How do you feel about also adding `:oban` to the log event domain? Oban does insert `source: "oban"` into the report, but it's hard to access if the report is json

For the record, this was separately added in https://github.com/oban-bg/oban/pull/1364.

PR #1332, opened 2025-08-14
Allow for non-double errors in perform handling

krainboltgreene:
A few libraries return tuples that have `:error` as a first element but also don't conform to `{atom(), any()}` instead opting for more data, for example the notable Ecto.Mutli returns: `{:error, :successful_event, Ecto.Changeset.t(), list(Ecto.Changeset.t())}`

Right now every oban user who has a job that ends in a Multi that errors out might not be seeing failures that they would normally see if they had used the function version of `Ecto.Repo.transact()`.

I realize that this is probably not enough to get merged and I am willing to spend time on any change requests or additional efforts (like testing) once I approval to continue.

### DIFF
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -165,9 +165,13 @@ defmodule Oban.Queue.Executor do
         %{exec | result: result, state: :snoozed, snooze: seconds}
 
       returned ->
-        log_warning(exec, returned)
-
-        %{exec | state: :success, result: returned}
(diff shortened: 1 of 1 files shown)


sorentwo:
This is a welcome and understandable change. However, it's changing how the worker behavior operates and will need to wait for Oban 3.0.

This is now tracked in the [3.0 changes issue](https://github.com/oban-bg/oban/issues/849).

krainboltgreene:
Understood, thanks for the feedback. Ping me if you want me to reapproach this for 3.0.

PR #1337, opened 2025-08-22
Update documentation for job telemetry

elfenlaid:
The changes elaborate on what counts as success in the context of job telemetry. 

Maybe my wording here is off, but essentially it would be nice to point out that not only jobs that return `ok` count as successful ones. I think that currently it might be difficult to figure out from the first glance purely by reading documentation.

It's also possible to circumvent this whole description of success by replacing `succeeded` with `finished` and providing the list of values corresponding to events 😅

### DIFF
--- a/lib/oban/telemetry.ex
+++ b/lib/oban/telemetry.ex
@@ -25,9 +25,9 @@ defmodule Oban.Telemetry do
   * `[:oban, :job, :stop]` — after a job succeeds and the success is recorded in the database
   * `[:oban, :job, :exception]` — after a job fails and the failure is recorded in the database
 
-  Jobs that complete with `:ok`, `:cancelled`, `:snoozed`, and `:discard` count as 
-  suceeded and emit the `:stop` event. Jobs that complete with an `:error` or raised an exception
-  emit the `:exception` event.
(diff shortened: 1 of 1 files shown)


lib/oban/telemetry.ex
```diff
@@ -25,6 +25,10 @@ defmodule Oban.Telemetry do
   * `[:oban, :job, :stop]` — after a job succeeds and the success is recorded in the database
   * `[:oban, :job, :exception]` — after a job fails and the failure is recorded in the database
 
+  Jobs that complete with `:ok`, `:cancelled`, `:snoozed`, and `:discard` count as 
+  suceeded and emit the `:stop` event. Jobs that complete with an `:error` or raised an exception
+  emit the `:exception` event.
```
sorentwo, line 30: The `:discard` state shouldn't be included in this list. Returning a `discard` tuple is long deprecated and not documented as a viable return anymore, so the only way a `discard` should happen is on the final error.

```suggestion
  Jobs that complete with `:ok`, `:cancelled`, or `:snoozed`count as succeeded and emit the `:stop`
  event. Jobs that finish with an `:error`, exception, or crash will emit the `:exception`
  event.
```

sorentwo:
Thank you!

elfenlaid:
Keep up the awesomeness! ❤️

PR #1352, opened 2025-10-10
fix: measurements from the Executor process, not parent

martinsvalin:
When Producer starts jobs, it calls `Executor.new` in its own process, before handing it off to `Task.Supervisor.async_nolink` to call `Executor.call`. That means we currently write in the Producer pid, and take measurements from that process.

This was the simplest approach I could think of to have the executor read its own measurements. Did not think it merits a test, but that judgement call is yours to make. 
Thank you so much for Oban! It's a joy to work with and to read.

### DIFF
--- a/lib/oban/queue/executor.ex
+++ b/lib/oban/queue/executor.ex
@@ -36,7 +36,6 @@ defmodule Oban.Queue.Executor do
     :error,
     :job,
     :meta,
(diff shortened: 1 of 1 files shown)


sorentwo:
Looks perfect! Thanks for the fix. The existing unit test passes, so this is golden.

PR #1356, opened 2025-10-17
Update backoff timing details in worker documentation

chaodhib:
Based on our conversation in Slack and the following spreadsheet:
image

the spreadsheet shows that:
* the time between the 19th & 20th attempt is 576733.3 seconds (so 6.6 days)
* the time between the first and 20th attempt is 320 hours and 29 minutes (so around 13.3 days). This seems to be consistent with the discarded jobs I have found.


sorentwo:
Thanks!

PR #1364, opened 2025-10-29
Add domain to log metadata

marpo60:
Hello,

While I was reading https://github.com/oban-bg/oban/pull/1304# I thought that it my be a good addition to add [domain](https://hexdocs.pm/logger/Logger.html#module-metadata) to the log metadata.

This would allow logger handlers to filter messages that come from Oban.

Although this information is present in the `source` attribute, it’s not part of the metadata, it’s part of the attributes. And when `encode = true`, it’s generated as a string, which makes it less convenient to work with.

This as far as I know would not require a major version bump.

Similar additions
* [bandit](https://github.com/mtrudel/bandit/blob/6a1976c069e215e4c7d1c250d039288e5b88d436/CHANGELOG.md?plain=1#L177)
* [plug_cowboy](https://github.com/elixir-plug/plug_cowboy/commit/3760ab479b02279d39ddf2b9cf7791f07470277c)

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -82,6 +82,7 @@
         {Credo.Check.Warning.IExPry, []},
         {Credo.Check.Warning.IoInspect, []},
         {Credo.Check.Warning.LazyLogging, false},
+        {Credo.Check.Warning.MissedMetadataKeyInLoggerConfig, false},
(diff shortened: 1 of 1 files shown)


sorentwo:
Great addition, thanks! You're right, it's backward compatible and can be added before 3.0 changes.

PR #1368, opened 2025-11-04
README.md: Update explanatory link for queues

xtian:
The current link's hash does not point anywhere in the `Oban` module docs


sorentwo:
Thanks! That was left over from the docs refactor.

PR #1381, opened 2025-12-09
Validate no duplicate option is passed to Oban config

rschef:
We ran into an issue recently that an option was passed twice to the Oban configuration by mistake and it ended up causing issues:

```
config :my_app, Oban,
  peer: Oban.Peers.Database,
  ...,
  # This was the desired value, but the first one was read
  peer: false
```
This MR validates that no duplicate option is passed.

### DIFF
--- a/lib/oban/config.ex
+++ b/lib/oban/config.ex
@@ -135,7 +135,7 @@ defmodule Oban.Config do
   """
   @spec validate([Oban.option()]) :: :ok | {:error, String.t()}
   def validate(opts) when is_list(opts) do
-    with :ok <- validate_no_duplicate_option(opts) do
+    with :ok <- validate_unique_opts(opts) do
       opts = normalize(opts)
 
       Validation.validate_schema(opts,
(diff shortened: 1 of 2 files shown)


lib/oban/config.ex
```diff
@@ -135,26 +135,28 @@ defmodule Oban.Config do
...
-      peer: {:or, [:falsy, {:behaviour, Oban.Peer}]},
-      plugins: {:custom, &validate_plugins/1},
-      prefix: {:or, [:falsy, :string]},
-      queues: {:custom, &validate_queues(opts, &1)},
-      repo: {:module, [config: 0]},
-      shutdown_grace_period: :non_neg_integer,
-      stage_interval: :timeout,
-      testing: {:enum, @testing_modes}
-    )
+    with :ok <- validate_no_duplicate_option(opts) do
```
sorentwo:
```suggestion
    with :ok <- validate_unique_opts(opts) do
```

rschef:
Thanks, I applied the suggestions!

sorentwo:
Thanks!

PR #1407, opened 2026-02-01
Explain that pause in queue definition is local only

apreifsteck:
## What
Noted that queue pauses defined in configuration will only affect queues on the local node.
## Why
I'd figure I'd enshrine the answer @sorentwo gave me in this [forum thread](https://elixirforum.com/t/do-pause-semantics-in-oban-queue-configuration-apply-locally-or-globally/74115/2?u=apreifsteck) so that anyone looking for an answer to a similar question could find it right in the docs.


sorentwo:
Thanks!

PR #1411, opened 2026-02-04
fix oban installer missing callback

jkbbwr:


sorentwo:
I'm unable to reproduce the issue. The callback appears to be defined automatically.

PR #1413, opened 2026-02-10
Fix Sonar and Midwife listener loss after Notifier crash

QuinnWilton:
Note that this issue was found while running a static analysis tool I'm working on. It seems like a legitimate finding, but it isn't something I observed while operating `oban`.

Both Sonar and Midwife register as listeners on the Notifier during init, but under the one_for_one supervisor strategy, a Notifier crash only restarts the Notifier — not its siblings. The new Notifier starts with an empty listener map, silently breaking notification delivery.

For Sonar, this means pings are sent but never received back, causing it to degrade to :isolated status. For Midwife, signal notifications (queue start/stop) are never delivered.

Sonar now re-registers on every ping cycle by calling Notifier.listen before Notifier.notify. The call is idempotent when already registered and self-healing when registration is lost. Both calls are wrapped in a try/catch to handle the brief window where a GenServer.call to a stale Notifier pid would otherwise crash Sonar with an EXIT.

Midwife monitors the Notifier process and re-registers on :DOWN via handle_continue, falling back to a timed retry if the new Notifier isn't available yet.

### DIFF
--- a/lib/oban/stager.ex
+++ b/lib/oban/stager.ex
@@ -104,7 +104,11 @@ defmodule Oban.Stager do
 
     payload = Enum.map(queues, &%{queue: &1})
 
-    Notifier.notify(conf, :insert, payload)
+    try do
+      Notifier.notify(conf, :insert, payload)
+    catch
+      :exit, _ -> :ok
+    end
   end
 
   defp notify_queues(%{conf: conf, mode: :local}) do
--- a/test/oban/stager_test.exs
+++ b/test/oban/stager_test.exs
@@ -1,7 +1,70 @@
+defmodule Oban.StagerTest.ExitOnNotify do
+  @moduledoc false
+  @behaviour Oban.Notifier
+
+  use GenServer
+
+  defstruct [:conf, armed: false, listeners: %{}]
+
+  @impl Oban.Notifier
+  def start_link(opts) do
+    {name, opts} = Keyword.pop(opts, :name)
+
+    GenServer.start_link(__MODULE__, struct!(__MODULE__, opts), name: name)
+  end
+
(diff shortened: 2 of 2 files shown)


QuinnWilton:
I just found a third issue to do with the Notifier crashing! That issue + the fix are in https://github.com/oban-bg/oban/pull/1413/changes/899303ff30e939b9abc86856a0f8ac22ed3e5449

lib/oban/midwife.ex
```diff
@@ -105,5 +118,27 @@ defmodule Oban.Midwife do
     {:noreply, state}
   end
 
+  defp resubscribe(%State{conf: conf} = state) do
+    case Registry.whereis(conf.name, Oban.Notifier) do
+      pid when is_pid(pid) ->
+        Notifier.listen(conf.name, :signal)
+        ref = Process.monitor(pid)
+
+        {:noreply, %{state | notifier_ref: ref}}
```
sorentwo: Purely a style nitpick:

```suggestion

        {:noreply, %{state | notifier_ref: Process.monitor(pid)}}
```

QuinnWilton:
> The changes all look good, but some of the tests are too invasive compared to the way the rest of the tests operate and I'd prefer they were associated with the affected modules.

Thanks for the review!

Pushed up that style change, and tried to simplify the tests. Midwife and Sonar are a lot simpler now, but I didn't see a way to test the Stager behavior without making a mock notifier: happy to adjust if you have ideas though!

sorentwo:
Merged locally. I opted to remove the Stager test entirely, considering the amount of boilerplate for a notifier module to verify that a notifier call had a `try/rescue` around it. Thanks!

PR #1432, opened 2026-03-30
Addresses nested transaction losing error information

tfwright:


sorentwo:
Any changes we make to handle transactions would need to be at the Repo level, so it applies to all of the adapters and `Repo` calls.

tfwright:
Gotcha. Gonna close this for now as it was mostly for info purposes only

PR #1438, opened 2026-04-14
Adds disable_sonar option to skip Sonar for insert-only instances

leepage87:
Insert-only Oban instances in umbrella apps (configured with `peer: false, queues: [], plugins: []`) exist solely to enqueue jobs for a separate processing app. Sonar still starts unconditionally on these instances, opening a persistent notifier connection and sending pg_notify pings every 5 seconds. At scale this wastes database connections and generates unnecessary traffic that no one consumes.

## Solution
Add a disable_sonar boolean option to Oban.Config (defaulting to false). When true, Sonar.start_link/1 returns :ignore and no Sonar process is started — the same mechanism already used when testing is :inline or :manual.

```
Oban.start_link(
  repo: MyApp.Repo,
  peer: false,
  queues: [],
  plugins: [],
  disable_sonar: true
)
```

### DIFF
--- a/lib/oban/config.ex
+++ b/lib/oban/config.ex
@@ -11,6 +11,7 @@ defmodule Oban.Config do
   alias Oban.Validation
 
   @type t :: %__MODULE__{
+          disable_sonar: boolean(),
           dispatch_cooldown: pos_integer(),
           engine: module(),
           get_dynamic_repo: nil | (-> pid() | atom()) | {module(), atom(), list()},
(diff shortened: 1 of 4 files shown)


iacutone (approved):

sorentwo:
There are reasons you'd want to have an insert-only node check the notifier status. Oban Web is a prime example. You want to check that it's receiving notifications despite the fact that it's not processing jobs.

Oban configuration is tricky enough as is, and I'd prefer not to add another control. What I recommend instead is using the `Oban.Notifiers.PG` module instead, which avoids the Postgres query overhead. Internally I'll look at adaptively increasing the sonar ping timing to minimize the amount of background traffic though.

PR #1439, opened 2026-04-14
Add transaction_retry config option

tfwright:
See https://github.com/oban-bg/oban/pull/1432, https://elixirforum.com/t/unexplained-rollback-after-2-21-upgrade/74812/8

Even after fixing the root cause the previous error, we are still seeing generic `:rollback` results from transactions with Oban operations. We'd prefer to allow the underlying error cause to bubble up which doesn't seem to be possible with the transaction retry behavior. This adds an option to disable, defaulting to true in order to preserve existing behavior.

I didn't see any tests for actual retry behavior so I just added one for config.

### DIFF
--- a/lib/oban.ex
+++ b/lib/oban.ex
@@ -88,6 +88,7 @@ defmodule Oban do
           | {:shutdown_grace_period, non_neg_integer()}
           | {:stage_interval, timeout()}
           | {:testing, :disabled | :inline | :manual}
+          | {:transaction_retry, boolean()}
 
   @typedoc """
   Options for draining jobs from a queue.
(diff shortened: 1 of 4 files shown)


sorentwo:
@tfwright Thanks for the PR. Adding a top-level config is too broad, since this is something that can be controlled at the function level. With the linked commit I pushed you can prevent masking errors on insert like this in Pro:

```elixir
Oban.insert_all(jobs, retry: false)
```

After the commit prior to the linked one above, we made `retry_opts` configurable at compile time, so you could disable retrying globally if you want.

tfwright:
OK, I'll look for those options in the next release. Thanks.

PR #1441, opened 2026-04-21
Fix duplicate pid accumulation in Notifier.Postgres put_channels/3

Lgdev07:
Oban.Sonar calls `Notifier.listen/2` on every ping (every 5s) to recover from Notifier crashes, a change introduced in [v2.21](https://github.com/oban-bg/oban/commit/b7042ce7e86cede452670d8a8d95f6c838042442#diff-c8f1d4dc3f3d39a0703a2eaede00a4b68bb416566c9c8c6e7f0162f7bae64b7dR74-R80)

The **Postgres** notifier's `put_channels/3` appended the caller pid unconditionally on every call with no deduplication check:
    `Map.update(acc, channel, [pid], &[pid | &1])`
    
The listeners map ([put_listener/3](https://github.com/oban-bg/oban/blob/cbd74ab67fc388a7091a0f115199e715e176277e/lib/oban/notifiers/postgres.ex#L225)) correctly deduplicates by pid, but the channels map used for delivery did not. This caused the Sonar pid to accumulate in the sonar channel list at 1 entry per 5s, forever.

After 12 hours a pod accumulates ~8,640 duplicate entries. On every sonar ping, `Notifier.relay/4` iterates the full list and floods the Sonar process with duplicate messages, causing per-pod CPU to grow linearly with pod age.

This PR addresses this by using `Enum.uniq/1` to deduplicate on insert

### DIFF
--- a/lib/oban/notifiers/postgres.ex
+++ b/lib/oban/notifiers/postgres.ex
@@ -242,7 +242,7 @@ if Code.ensure_loaded?(Postgrex) do
     defp put_channels(state, pid, channels) do
       listener_channels =
         for channel <- channels, reduce: state.channels do
-          acc -> Map.update(acc, channel, [pid], &[pid | &1])
+          acc -> Map.update(acc, channel, [pid], &Enum.uniq([pid | &1]))
         end
 
       %{state | channels: listener_channels}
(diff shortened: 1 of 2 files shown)


sorentwo:
Fantastic fix, thank you!

PR #1463, opened 2026-06-12
fix: set `successful` state in default unique options

milmazz:
The documentation says that the default `states` value is `:successful` for `unique` options, which includes the recent `suspended` state, but the implementation doesn't reflect that.

This commit should align the implementation with the documentation.

### DIFF
--- a/lib/oban/job.ex
+++ b/lib/oban/job.ex
@@ -234,7 +234,7 @@ defmodule Oban.Job do
     fields: ~w(args queue worker)a,
     keys: [],
     period: 60,
-    states: ~w(scheduled available executing retryable completed)a,
+    states: :successful,
     timestamp: :inserted_at
   }
 
--- a/test/oban/job_test.exs
+++ b/test/oban/job_test.exs
@@ -101,7 +101,7 @@ defmodule Oban.JobTest do
                fields: [:args, :queue, :worker],
                keys: [],
                period: 60,
-               states: ~w(scheduled available executing retryable completed)a,
+               states: ~w(suspended scheduled available executing retryable completed)a,
                timestamp: :inserted_at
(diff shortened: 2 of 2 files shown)


milmazz:
Current failure in CI seems unrelated to this change; is it coming from `igniter`? That's the only thing that uses `req` according to `mix deps.tree`:

```console
Run mix test --exclude lite || mix test --failed
** (Mix) Could not start application req: Req.Application.start(:normal, []) returned an error: shutdown: failed to start child: Req.Finch
    ** (EXIT) an exception was raised:
        ** (ArgumentError) errors were found at the given arguments:

  * 2nd argument: invalid options

            (stdlib 3.17.2.4) :ets.new(Req.Finch.PoolMetrics, [:ordered_set, :public, :named_table, {:read_concurrency, true}, {:write_concurrency, :auto}])
            (finch 0.22.0) lib/finch/pool_metrics.ex:25: Finch.PoolMetrics.new/1
            (finch 0.22.0) lib/finch.ex:460: Finch.init/1
            (stdlib 3.17.2.4) supervisor.erl:330: :supervisor.init/1
            (stdlib 3.17.2.4) gen_server.erl:423: :gen_server.init_it/2
            (stdlib 3.17.2.4) gen_server.erl:390: :gen_server.init_it/6
            (stdlib 3.17.2.4) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
** (Mix) Could not start application req: Req.Application.start(:normal, []) returned an error: shutdown: failed to start child: Req.Finch
    ** (EXIT) an exception was raised:
        ** (ArgumentError) errors were found at the given arguments:
```

sorenone:
Thank you!

PR #1464, opened 2026-06-12
add Docker compose for tests

milmazz:
While working on #1463, I wanted to run the whole test suite locally, but didn't want to install MySQL, so I prefer to keep those dependencies in Docker containers when I need them for testing.

After finishing the previous work, I thought this might be helpful for other contributors.


sorenone:
Fabulous, thanks Milton!

PR #1465, opened 2026-06-13
Lock Finch version to keep CI compatible w/OTP-24

milmazz:
As noted in this [comment](https://github.com/oban-bg/oban/pull/1463#issuecomment-4695713530), there is an error in CI that's coming from Finch 0.22 because it is passing some invalid arguments to `:ets.new` inside the new `Finch.PoolMetrics` module, the issue is that `write_concurrency: :auto` was added in OTP-25, that's why the argument is invalid against this OTP version.

I'm locking the `finch` version because keeps the CI matrix as it's now and this package is only used during tests, so, it should be safe to take this approach.

See: https://www.erlang.org/blog/my-otp-25-highlights/#ets-tables-with-adaptive-support-for-write-concurrency


sorenone:
Very appreciated. We decided to bump to OTP 25 and Elixir 1.16. Especially, now that Elixir 1.15 is no longer maintained.

PR #1468, opened 2026-06-22
Update Oban.Migration docs to use different versions

sheharyarn:
The "Usage" section in the Oban.Migration docs incorrectly used the same version in the sample migration.

It has now been updated to use different versions, which reflects real-world migration scenarios.

### DIFF
--- a/lib/oban/migration.ex
+++ b/lib/oban/migration.ex
@@ -43,17 +43,17 @@ defmodule Oban.Migration do
   additional migrations. To do this, generate a new migration:
 
   ```bash
-  mix ecto.gen.migration upgrade_oban_to_v13
+  mix ecto.gen.migration upgrade_oban_to_v14
   ```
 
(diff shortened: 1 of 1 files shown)


lib/oban/migration.ex
```diff
@@ -43,17 +43,17 @@ defmodule Oban.Migration do
...
 
   ```elixir
-  defmodule MyApp.Repo.Migrations.UpgradeObanToV13 do
+  defmodule MyApp.Repo.Migrations.UpgradeObanToV14 do
     use Ecto.Migration
 
-    def up, do: Oban.Migrations.up(version: 13)
+    def up, do: Oban.Migrations.up(version: 14)
 
     def down, do: Oban.Migrations.down(version: 13)
```
sheharyarn, line 58: Previously both were same versions

sorentwo:
Thanks for the fix!

PR #1479, opened 2026-07-07
Accept period durations in plugins

milmazz:
This PR extends previous work to accept period durations in more Plugin user-facing options.

In essence, now we accept period durations in the `interval` option from `Pruner` and `Lifeline` plugins, and `period` in the `Reindexer` plugin.

Assisted-by: Claude:claude-4-8-opus

### DIFF
--- a/lib/oban/plugins/reindexer.ex
+++ b/lib/oban/plugins/reindexer.ex
@@ -30,7 +30,9 @@ defmodule Oban.Plugins.Reindexer do
 
     * `:schedule` — a cron expression that controls when to reindex. Defaults to `"@midnight"`.
 
-    * `:timeout` - time in milliseconds to wait for each query call to finish. Defaults to 15 seconds.
+    * `:timeout` - the time to wait for each query call to finish, as either an integer number of
+      milliseconds, `:infinity`, or an `Oban.Period` tuple like `{15, :seconds}`. Defaults to 15
+      seconds.
 
     * `:timezone` — which timezone to use when evaluating the schedule. To use a timezone other than
       the default of "Etc/UTC" you *must* have a timezone database like [tz][tz] installed and
@@ -43,7 +45,7 @@ defmodule Oban.Plugins.Reindexer do
 
   use GenServer
 
-  alias Oban.{Cron, Peer, Plugin, Repo, Validation}
+  alias Oban.{Cron, Peer, Period, Plugin, Repo, Validation}
   alias __MODULE__, as: State
 
   require Logger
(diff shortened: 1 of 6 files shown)


lib/oban/plugins/reindexer.ex
```diff
@@ -53,7 +55,7 @@ defmodule Oban.Plugins.Reindexer do
           | {:indexes, [String.t()]}
           | {:schedule, String.t()}
           | {:timezone, Calendar.time_zone()}
-          | {:timeout, timeout()}
+          | {:timeout, timeout() | Period.t()}
```
sorentwo, line 58: Thinking about this one a bit and the overlap between these types. Should `:infinity` be a valid period (probably not)? Do we expect to accept periods everywhere a timeout is accepted, like `Repo` operations (probably not)?

milmazz: > Thinking about this one a bit and the overlap between these types. Should :infinity be a valid period (probably not)?

With your `to_millisecond/1` addition in the `Period` module, I was thinking that a good improvement would be to normalize all these options, at least those that are user-facing, and I wanted to start with the plugins. For the `interval`, the translation was automatic, but for this one (`timeout`), as you already pointed out, the `:infinity` while it refers to a "period", I don't think it should be included in the Period module by default, it would change previous expectations, so that's why I went with the helper function to make the difference explicit.

> Do we expect to accept periods everywhere a timeout is accepted, like Repo operations (probably not)?

Probably not everywhere, but I think it will help normalize user-facing configuration options when the expected values are covered by the minimum unit granularity (seconds).

sorentwo: > With your to_millisecond/1 addition in the Period module, I was thinking that a good improvement would be to normalize all these options, at least those that are user-facing, and I wanted to start with the plugins.

Totally, I like that thinking. This makes the most sense for plugins where you'd read it in config. I think modern Elixir apps should probably use `to_timeout/1`, but we support older Elixir versions and couldn't switch to that.

For the `Reindexer` in particular, there's no reason to pass `:infinity` to disable it. It may be subtly breaking, but since it's the wrong way to disable a plugin, I'd replace the `timeout()` type with `Period.t()` instead.

sorentwo: @milmazz Would you like me to pick this up?

milmazz: @sorentwo oh, yes, please go ahead. Thanks.

sorentwo:
Merged manually as 2220c08, thanks @milmazz!

PR #1483, opened 2026-07-15
Fix migration verification for custom adapters

dannote:
## Summary

Use the repo selected from `opts[:repo]` when resolving a configured custom migrator, and cover the migration-verification path with a regression test.

## Motivation

`migrator/1` selects an explicit `:repo`, but the custom-adapter fallback currently ignores that value and calls `Ecto.Migration.repo/0` again:

```elixir
_ -> Keyword.fetch!(repo().config(), :migrator)
```

`Ecto.Migration.repo/0` is only available while an Ecto migration runner is active. Oban's migration verification passes the configured repo explicitly and runs outside that context, so custom adapters with a configured `:migrator` fail before their migrator can be consulted.

Reading `repo.config()` uses the repo that was already selected, matching the built-in adapter branches and allowing migration verification to work for external Ecto adapters.

## Test

The regression test defines a minimal custom adapter, repo, and migrator, then exercises `Oban.Migration.verify_migrated!/1` with the repo passed explicitly.

Locally verified with formatting, warnings-as-errors compilation, the focused regression test, and strict Credo.

### DIFF
--- a/lib/oban/migration.ex
+++ b/lib/oban/migration.ex
@@ -272,7 +272,7 @@ defmodule Oban.Migration do
       Ecto.Adapters.Postgres -> Oban.Migrations.Postgres
       Ecto.Adapters.SQLite3 -> Oban.Migrations.SQLite
       Ecto.Adapters.MyXQL -> Oban.Migrations.MyXQL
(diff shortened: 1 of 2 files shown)


sorentwo:
@dannote Thanks!

PR #1503, opened 2026-09-17
Fix stray quote in migration comment example

dajiaohuang:
The `COMMENT ON TABLE` example carries a stray trailing quote, so the statement fails with a syntax error when copied as shown.


sorentwo:
Thank you!

PR #1504, opened 2026-09-17
Fix start_link reference in option typedoc

dajiaohuang:
The `t:option/0` typedoc points at `start_link1/`, which isn't a function on this module. `start_link/1` is the actual name, and correcting it also lets ExDoc render a working link to the function.


sorentwo:
Thank you!

